PR #173, opened 2025-07-27
fix: SimpleStructMemReaderConfig.remove_prompt_example=true has no effect

i255979:
When setting 'SimpleStructMemReaderConfig.remove_prompt_example' to true, it has no effect. The replacement failed.


Nyakult:
I checked the code and it seems to be working as expected, maybe try pulling the most recent code from the dev branch and test it again?

PR #1106, opened 2026-02-22
feat: Add TIAMAT lightweight HTTP memory connector

toxfox69:
## Summary

Adds **TiamatConnector** — a lightweight HTTP-based memory connector bridging MemOS with [TIAMAT Memory API](https://memory.tiamat.live).

- **Quick prototyping** — no database or infrastructure setup needed
- **Cloud deployments** — persistent memory without volume mounts
- **Multi-agent** shared memory across instances via cloud API
- **FTS5 search** for instant full-text memory recall
- **Knowledge triples** (subject/predicate/object)
- **MemOS interop** — import/export TextualMemoryItem format
- Only requires `httpx`

## Usage

```python
from tiamat_connector import TiamatConnector

c = TiamatConnector(api_key="key", user_id="user-1")
c.add_memory("content", tags=["tag"], importance=0.8)
results = c.search("query")
c.learn("subject", "predicate", "object")
```

## Files

- `examples/tiamat_connector/tiamat_connector.py` — Core connector
- `examples/tiamat_connector/example.py` — Usage example
- `examples/tiamat_connector/README.md` — Documentation

## Test Plan

- [ ] Verify add_memory and search cycle
- [ ] Verify knowledge triples via learn endpoint
- [ ] Verify import_textual_memories bulk import
- [ ] Verify export_as_textual_items MemOS compatibility
- [ ] Verify graceful fallback when API is unreachable

### DIFF
--- a/examples/tiamat_connector/README.md
+++ b/examples/tiamat_connector/README.md
@@ -1,0 +1,60 @@
+# TIAMAT Memory Connector for MemOS
+
+Lightweight HTTP-based memory connector that bridges MemOS with TIAMAT's cloud memory API. Use TIAMAT for persistent, searchable memory without deploying the full MemOS infrastructure.
+
+## When to Use
+
+- **Quick prototyping** — no database or infrastructure setup needed
+- **Cloud deployments** — persistent memory without volume mounts
+- **Multi-agent systems** — shared memory across agent instances
+- **Hybrid setups** — TIAMAT for cloud backup, MemOS for local processing
+
+## Quick Start
+
+```bash
+pip install httpx
+
+# Get a free API key
+curl -X POST https://memory.tiamat.live/api/keys/register \
+  -H "Content-Type: application/json" \
+  -d '{"agent_name": "my-agent", "purpose": "memory"}'
+
+export TIAMAT_API_KEY="your-key"
+python example.py
+```
+
+## Features
+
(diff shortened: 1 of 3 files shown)


CaralHsi:
Hi @toxfox69,

Thank you for this PR and the effort you put into the TIAMAT connector.

After review, we’ve decided not to merge it into the MemOS repo for now. Here’s why:

1. Architecture fit — MemOS’s storage layer is built around graph backends (Neo4j, Postgres, etc.) that implement BaseGraphDB with nodes, edges, embeddings, and graph traversal. TIAMAT’s flat storage + FTS5 model is different, so it doesn’t fit as a drop-in backend.

2. Primary use case — The connector mainly serves as a TIAMAT API client for users who don’t run MemOS. The MemOS-specific helpers (import_textual_memories / export_as_textual_items) are secondary, so the core value is outside MemOS’s scope.

3. Maintenance and dependencies — Adding a third-party cloud service client would introduce an external dependency and maintenance burden that we’d prefer to keep out of the core repo for now.

We’d suggest publishing this as a standalone package (e.g. tiamat-memos-connector or under the TIAMAT org) and optionally linking it from MemOS docs as a community integration. Users who want a lightweight TIAMAT-backed memory store can install it directly.

Thanks again for the contribution and for your interest in MemOS.

Issue #1172, opened 2026-03-06
[Bug] CypherTypeError: Map{} encountered when processing Skill Memory nodes


TelDragon:
### Pre-submission checklist | 提交前检查

- [x] I have searched existing issues and this hasn't been mentioned before | 我已搜索现有问题，确认此问题尚未被提及
- [x] I have read the project documentation and confirmed this issue doesn't already exist | 我已阅读项目文档并确认此问题尚未存在
- [x] This issue is specific to MemOS and not a general software issue | 该问题是针对 MemOS 的，而不是一般软件问题

### Bug Description | 问题描述

Title:
[Bug] CypherTypeError: Map{} encountered when processing Skill Memory nodes

Description:

While running MemOS and attempting to save or process Skill data into the Graph Database (Neo4j), the process crashes with a Cypher type error.

It appears that when ss_skill_memory.py calls the batch insertion function (add_nodes_batch), it is passing a Python Map{} (dictionarty) into a Cypher property field that expects primitive types (String, Integer, etc).

Traceback points to neo4j_community.py but the root cause originates in ss_skill_memory.py.

Steps to Reproduce:

Start MemOS instance (or run a Skill invocation scenario).
Perform an action that triggers the Skill Logic / Function Call.
The system attempts to store the Skill Memory as a Graph Node.
Error occurs during the add_nodes_batch execution.
Relevant Log Output:

<TEXT>
ss_skill_memory.py:259 - add_id_to_mysql - [PROCESS_SKILLS] SKILLS_MYSQL_URL or SKILLS_MYSQL_BEARER is not set
neo4j_community.py:215 - add_nodes_batch - [add_nodes_batch] Failed to add nodes: {code: Neo.ClientError.Statement.TypeError} {message: Property values can only be of primitive types or arrays thereof. Encountered: Map{}.}
neo4j.exceptions.CypherTypeError: {code: Neo.ClientError.Statement.TypeError} {message: Property values can only be of primitive types or arrays thereof. Encountered: Map{}.}
neo4j.exceptions.GqlError: {gql_status: 22N01} {gql_status_description: Expected the value Map{} to be of type BOOLEAN, STRING, INTEGER, FLOAT... but was of type MAP NOT NULL.}
The above exception was the direct cause of the following exception:
  File "/app/src/memos/graph_dbs/neo4j_community.py", line 212, in add_nodes_batch
    session.run(query, nodes=nodes_data)
  ...
Environment (填入你的环境):

[...]

Issue #1122, opened 2026-02-25
fix: Feedback API fails with Neo4j CypherTypeError - Property values cannot be Map types


paneel:
### Pre-submission checklist | 提交前检查

- [x] I have searched existing issues and this hasn't been mentioned before | 我已搜索现有问题，确认此问题尚未被提及
- [x] I have read the project documentation and confirmed this issue doesn't already exist | 我已阅读项目文档并确认此问题尚未存在
- [x] This issue is specific to MemOS and not a general software issue | 该问题是针对 MemOS 的，而不是一般软件问题

### Bug Description | 问题描述

The `/product/feedback` endpoint fails when using Neo4j as the graph backend. The API returns `{"code":200,"data":[[]]}` but no memory is created or updated.

### Root Cause | 根本原因

**Issue 1: Neo4j CypherTypeError**

In `src/memos/graph_dbs/neo4j_community.py:102`, when `add_node()` is called with metadata containing nested objects, Neo4j rejects it:

```
neo4j.exceptions.CypherTypeError: Property values can only be primitive types or arrays thereof. Encountered: Map{user_name -> String("test-user-001")}.
```

**Issue 2: IndexError in feedback.py**

In `src/memos/mem_feedback/feedback.py:254`, when `memory_manager.add()` returns an empty list due to the Neo4j error, the code crashes:

```python
return {
    "id": added_ids[0],  # IndexError: list index out of range
```

### How to Reproduce | 如何重现

1. Set up MemOS with Neo4j backend (neo4j-community)
2. Add a memory:
```bash
curl -X POST http://localhost:8000/product/add \
  -H "Content-Type: application/json" \
  -d '{
    "user_id": "test-user-001",
    "mem_cube_id": "test-cube-001",
    "messages": [{"role": "user", "content": "I prefer Python for AI development"}],
    "async_mode": "sync"
  }'
```
3. Try to provide feedback:
```bash
curl -X POST http://localhost:8000/product/feedback \
  -H "Content-Type: application/json" \
  -d '{
    "user_id": "test-user-001",
    "mem_cube_id": "test-cube-001",
    "memory_id": "<memory_id_from_step_2>",
    "history": [{"role": "user", "content": "I prefer Python for AI development"}],
    "feedback_content": "Actually I also like TypeScript"
  }'
```

### Expected Behavior | 预期行为

- Feedback should update or add memory
- API should return the new/updated memory ID
- No errors in logs

### Actual Behavior | 实际行为

[...]

PR #1178, opened 2026-03-08
fix(graph_dbs): sanitize nested metadata before Neo4j writes

kyan-du:
## Summary
- sanitize nested metadata values before writing Neo4j node properties
- apply the same sanitization in both Neo4j enterprise and community graph DB implementations
- add a regression test covering skill-memory style nested metadata

## Problem
Neo4j node properties only accept primitive values or arrays of primitives. Some MemOS flows (especially SkillMemory and feedback-related flows) can produce nested metadata like:
- `scripts: dict`
- `others: dict`
- nested objects under `info`
- arrays containing dict items

Those values can reach `SET n += $metadata` and trigger errors like:
- `Property values can only be of primitive types or arrays thereof`
- `CypherTypeError: Map{} encountered`

## Fix
This PR adds a small sanitization layer before Neo4j writes:
- keep primitive values as-is
- keep arrays of primitives as-is
- serialize dict values to JSON strings
- serialize nested list items that are dict/list values

This keeps metadata readable while making it safe for Neo4j property storage.

## Validation
- reproduced the issue locally with SkillMemory-style metadata
- verified the sanitized metadata no longer contains Neo4j-invalid map values
- verified a real `/product/add` flow successfully created `SkillMemory` without the previous Neo4j `Map{}` error

### DIFF
--- a/src/memos/graph_dbs/neo4j.py
+++ b/src/memos/graph_dbs/neo4j.py
@@ -72,6 +72,27 @@ def _flatten_info_fields(metadata: dict[str, Any]) -> dict[str, Any]:
     return metadata
 
 
+def _sanitize_neo4j_value(value: Any) -> Any:
+    """Convert values unsupported by Neo4j properties into safe serializations."""
+    if value is None or isinstance(value, (str, int, float, bool)):
+        return value
+
+    if isinstance(value, list):
+        if all(item is None or isinstance(item, (str, int, float, bool)) for item in value):
+            return value
+        return [json.dumps(item, ensure_ascii=False) if isinstance(item, (dict, list)) else str(item) for item in value]
+
+    if isinstance(value, dict):
+        return json.dumps(value, ensure_ascii=False, sort_keys=True)
+
+    return str(value)
+
+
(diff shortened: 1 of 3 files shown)


kyan-du:
This PR addresses the root cause described in #1172 and partially in #1122.

The existing `_flatten_info_fields()` fix (from #1122) handles the top-level `info` dict, but other dict-typed fields (`scripts`, `others`, and nested objects inside `info`) still reach Neo4j as `Map{}` values.

This PR adds a universal sanitization step that catches all remaining cases.

Pemachys:
Hey @kyan-du, thanks so much for this fix! 
nested metadata hitting Neo4j has been bugging us for a while — love that you covered both enterprise and community impls + added regression tests. solid stuff 🙌
btw we have a [Discord](https://discord.gg/WngwZYy2) and official WeChat group, where we share early roadmap stuff and discuss architecture — would be cool to see you there!

angarchie:
Hey @kyan-du! Thanks for the contribution — already merged your PR, thanks for putting this together! 🚀

PR #1219, opened 2026-03-13
fix(viewer): handle multiple timestamp formats in days calculation

abakane1:
## Description

- Support milliseconds, microseconds, and seconds timestamp formats in days calculation
- Fixes incorrect days calculation (showing 36500 days) when timestamps use different formats

## Changes

Modified `apps/memos-local-openclaw/src/viewer/html.ts`:

```javascript
// Before: only handled milliseconds
if(e<1e12) e*=1000;

// After: handles all timestamp formats
if (e < 1e15 && e >= 1e12) e = Math.floor(e / 1000);  // microseconds to milliseconds
else if (e < 1e12) e *= 1000;  // seconds to milliseconds
```

## Related Issue

N/A (found during local testing)

### DIFF
--- a/apps/memos-local-openclaw/src/viewer/html.ts
+++ b/apps/memos-local-openclaw/src/viewer/html.ts
@@ -3170,8 +3170,11 @@ async function loadStats(){
   if(d.timeRange&&d.timeRange.earliest!=null&&d.timeRange.latest!=null){
     let e=Number(d.timeRange.earliest), l=Number(d.timeRange.latest);
     if(Number.isFinite(e)&&Number.isFinite(l)){
-      if(e<1e12) e*=1000;
-      if(l<1e12) l*=1000;
+      // Handle different timestamp formats: seconds, milliseconds, microseconds
+      if (e < 1e15 && e >= 1e12) e = Math.floor(e / 1000);  // microseconds to milliseconds
+      else if (e < 1e12) e *= 1000;  // seconds to milliseconds
+      if (l < 1e15 && l >= 1e12) l = Math.floor(l / 1000);  // microseconds to milliseconds
+      else if (l < 1e12) l *= 1000;  // seconds to milliseconds
       days=Math.round((l-e)/86400000);
       days=Math.max(0,Math.min(36500,days));
       if(days===0) days=1;


hijzy:
Thanks for the PR — the direction is good, but the timestamp boundary logic needs adjustment.
Current logic treats 1e12 ~ 1e15 as microseconds and divides by 1000, which incorrectly affects millisecond timestamps.

Counterexample: 1700000000000 (milliseconds) becomes 1700000000 (seconds), i.e. wrong unit downscaling.

Please normalize to milliseconds as:

- ts >= 1e15 : microseconds → Math.floor(ts / 1000)
- 1e12 <= ts < 1e15 : milliseconds → keep as-is
- ts < 1e12 : seconds → ts * 1000
Please also add a minimal test covering seconds/milliseconds/microseconds to prevent regressions. Thanks!

感谢提交！方向是对的，但时间戳区间判断还需要调整。
你现在把 1e12 ~ 1e15 当作微秒并除以 1000，这会误伤毫秒时间戳。

反例： 1700000000000 （毫秒）会被除成 1700000000 （秒），单位被错误降级。

请改为先统一到毫秒：
- ts >= 1e15 ：微秒 → Math.floor(ts / 1000)
- 1e12 <= ts < 1e15 ：毫秒 → 保持不变
- ts < 1e12 ：秒 → ts * 1000

另外请补一个最小测试覆盖 seconds/milliseconds/microseconds，避免回归。谢谢！

hijzy:
## PR Review

This PR fixes the viewer timestamp normalization to handle three formats: seconds (< 1e12), milliseconds (1e12–1e15), and microseconds (>= 1e15). Previously, only seconds-to-milliseconds conversion was handled.

**However, there is a logic bug:** Millisecond timestamps from ~2001 to ~2286 fall in the 1e12–1e15 range. The condition `e < 1e15 && e >= 1e12` treats all values in this range as microseconds and divides by 1000, but a value like `1.5e12` (a valid millisecond timestamp from 2017) would be incorrectly treated as microseconds. This means most real-world millisecond timestamps would be misinterpreted.

此 PR 修复了 viewer 中时间戳归一化逻辑，支持秒、毫秒、微秒三种格式。但存在逻辑缺陷：毫秒级时间戳（2001–2286年）的范围正好是 1e12–1e15，与微秒判断区间重叠。例如 `1.5e12`（2017年的毫秒时间戳）会被错误地当作微秒除以 1000，导致天数计算错误。

- **Safety:** ⚠️ Timestamp range logic has a bug
- **Mergeable:** After fixing the range logic
- **Confidence:** 2/5

hijzy:
### 🔧 Suggested Fix

The threshold between milliseconds and microseconds needs adjustment. A reasonable approach:

```js
if (e < 1e12) e *= 1000;          // seconds → ms
else if (e >= 1e15) e = Math.floor(e / 1000);  // microseconds → ms
// else: already milliseconds, no conversion needed
```

This avoids the ambiguous 1e12–1e15 range entirely by only converting values clearly above the millisecond range (>= 1e15 = year 33658 in ms).

### 🔧 修改建议

毫秒和微秒的判断阈值需要调整。建议：

```js
if (e < 1e12) e *= 1000;          // 秒 → 毫秒
else if (e >= 1e15) e = Math.floor(e / 1000);  // 微秒 → 毫秒
// 否则：已经是毫秒，无需转换
```

这样完全避开了 1e12–1e15 的模糊区间，只对明确大于毫秒范围的值（>= 1e15，即毫秒时间戳的 33658 年）做微秒转换。

PR #1239, opened 2026-03-16
Fix Windows sqlite path check and Gemini embedding defaults

buzuweidao:
## Summary

This PR fixes two issues we hit while testing `apps/memos-local-openclaw` on Windows with Gemini:

1. **Windows path handling for `better-sqlite3`**
   - `path.dirname(new URL(import.meta.url).pathname)` produced `/C:/...` on Windows.
   - `require.resolve()` returned `C:\...`, so the plugin incorrectly concluded `better-sqlite3` was outside the plugin directory and failed to load.
   - Switched to `fileURLToPath(import.meta.url)` and normalized both paths before comparison.

2. **Gemini embedding defaults / compatibility**
   - Default embedding model changed from `text-embedding-004` to `gemini-embedding-001`.
   - Gemini embedding endpoint normalization now auto-corrects older `v1/...:embedContent` style values to the current `v1beta/...:batchEmbedContents` style.
   - README example model updated accordingly.

## Validation

- Reproduced on Windows while installing and enabling the plugin under OpenClaw.
- Verified the TypeScript build succeeds locally with:
  - `PUPPETEER_SKIP_DOWNLOAD=1 npm install`
  - `npm run build`
- Verified Gemini embedding requests succeed against the official Gemini API when using `gemini-embedding-001`.

## Notes

The published package we tested also exposed a Viewer-side Gemini test issue, but that code path does not appear in the current repo source tree under `apps/memos-local-openclaw`. This PR only includes the source changes I could trace and validate in the repository itself.

### DIFF
--- a/apps/memos-local-openclaw/README.md
+++ b/apps/memos-local-openclaw/README.md
@@ -178,7 +178,7 @@ Add the plugin config to `~/.openclaw/openclaw.json`:
 | Provider | `provider` value | Example `model` | Notes |
 |---|---|---|---|
 | OpenAI / compatible | `openai_compatible` | `bge-m3`, `text-embedding-3-small` | Any OpenAI-compatible API |
-| Gemini | `gemini` | `text-embedding-004` | Requires `apiKey` |
+| Gemini | `gemini` | `gemini-embedding-001` | Requires `apiKey` |
 | Cohere | `cohere` | `embed-english-v3.0` | Separates document/query embedding |
 | Voyage | `voyage` | `voyage-2` | |
 | Mistral | `mistral` | `mistral-embed` | |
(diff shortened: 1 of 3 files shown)


syzsunshine219:
Thanks for the fix! Two separate improvements here:
1. Windows path fix (index.ts) — Duplicate of #1265
The fileURLToPath + path.normalize().toLowerCase() change is identical to what #1265 already provides. Since #1265 is a focused single-issue fix for the same problem (#1192), we'll merge that one first. Please rebase this PR and drop the index.ts changes to avoid conflicts.
2. Gemini embedding defaults (gemini.ts + README.md) — Looks good ✅
Updating the default model from text-embedding-004 to gemini-embedding-001 aligns with Google's latest recommendation.
The normalizeGeminiEmbeddingEndpoint() helper correctly handles /v1/ → /v1beta/ and :embedContent → :batchEmbedContents misconfigurations.
README update is consistent.
Action needed: please rebase onto the latest plugin branch (after #1265 lands), remove the index.ts diff, and keep only the Gemini changes (gemini.ts + README.md). Happy to merge once that's done.

PR #1373, opened 2026-03-27
fix: detect and rebuild better-sqlite3 on Node.js ABI mismatch

### LINKED ISSUE #1343: [bug]better-sqlite3 ABI mismatch after updating memos-local-openclaw-plugin on Windows; manual rebuild required

### Bug Description | 问题描述

Summary

After updating memos-local-openclaw-plugin on Windows, the plugin reports a successful update, but fails to load on gateway restart because better-sqlite3 is still compiled against an older Node ABI. The plugin only works again after manually running npm rebuild better-sqlite3.

This appears to be an installation/update flow issue for native dependencies on Windows, rather than a functional bug in the plugin logic itself.
Environment

    OS: Windows
    OpenClaw: 2026.3.23-1
    Node.js: v24.14.0
    Plugin: memos-local-openclaw-plugin
    Observed during upgrade path including:
    1.0.3
    1.0.5
    1.0.7

Steps to Reproduce

    Update the plugin:

openclaw plugins update memos-local-openclaw-plugin

    Start or restart the gateway:

openclaw gateway start

or

openclaw gateway restart

    Observe plugin registration failure on startup.

Actual Behavior

The update command reports success, for example:

Updated memos-local-openclaw-plugin: 1.0.5 -> 1.0.7.
Restart the gateway to load plugins and hooks.

But after restart, the plugin fails to register with this error:

memos-local-openclaw-plugin failed during register from C:\Users\long\.openclaw\extensions\memos-local-openclaw-plugin\index.ts:
Error: The module '\\?\C:\Users\long\.openclaw\extensions\memos-local-openclaw-plugin\node_modules\better-sqlite3\build\Release\better_sqlite3.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 127. This version of Node.js requires
NODE_MODULE_VERSION 137.
Please try re-compiling or re-installing the module (for instance, using `npm rebuild` or `npm install`).

Expected Behavior

After a successful plugin update, the plugin should load normally on gateway restart without requiring any manual rebuild steps.
Workaround

Running the following manually fixes the problem:

cd C:\Users\long\.openclaw\extensions\memos-local-openclaw-plugin
npm rebuild better-sqlite3
openclaw gateway restart

After that, the plugin loads normally again.
Notes

This does not look like a functional bug in the plugin logic itself.

It looks more like a Windows native dependency/update issue where:

    better-sqlite3 is left with an old ABI build after plugin update
    the update flow does not fully rebuild or refresh native modules
    the plugin update is reported as successful even though the native dependency is not yet usable for the current Node runtime

Suggested Fixes

Any of the following would probably make the update flow more robust on Windows:

    Automatically rebuild better-sqlite3 after plugin update
    Validate native module ABI compatibility before reporting update success
    Remove stale native binary artifacts before reinstalling dependencies
    Print an explicit post-update remediation hint when ABI mismatch is detected

Why this is confusing for users

From the user perspective:

    the update command succeeds
    the version appears updated
    but the plugin is still broken until a manual native rebuild is performed

So the current flow makes the upgrade appear complete when it is not actually ready to run.

If helpful, I can also provide a shorter “minimal reproducible report” version or a second issue specifically focused on the update command reporting success too early.

### How to Reproduce | 如何重现

1

### Environment | 环境信息

1

### Additional Context | 其他信息

1

zerone0x:
## Summary

- Fixes the `NODE_MODULE_VERSION` mismatch that prevents `memos-local-openclaw-plugin` from loading after a Node.js or plugin update on Windows (and other platforms)
- The `postinstall.cjs` script previously only checked whether the `.node` binary file existed on disk; it now also attempts to `require()` the module and catches ABI mismatch errors
- When an incompatible binary is detected, `npm rebuild better-sqlite3` runs automatically — no manual intervention needed

## Details

The root cause: `openclaw plugins update` installs new JS files but may reuse the existing native binary built for an older Node.js ABI. The postinstall script saw the binary existed and short-circuited with "ready", even though it would fail at runtime.

The fix adds a `sqliteLoadsSuccessfully()` function that does a trial `require()` of the module. If it throws a `NODE_MODULE_VERSION` error, the script proceeds to rebuild instead of reporting success.

Closes #1343

## Test plan

- [ ] On a machine with better-sqlite3 already built, run `node scripts/postinstall.cjs` — should report "ready" without rebuilding
- [ ] Swap to a different Node.js version (e.g. via nvm), run `node scripts/postinstall.cjs` — should detect ABI mismatch and trigger rebuild automatically
- [ ] Delete the `build/` directory, run `node scripts/postinstall.cjs` — should detect missing bindings and rebuild

### DIFF
--- a/apps/memos-local-openclaw/scripts/postinstall.cjs
+++ b/apps/memos-local-openclaw/scripts/postinstall.cjs
@@ -226,7 +226,35 @@ function sqliteBindingsExist() {
   return false;
 }
 
-if (sqliteBindingsExist()) {
+/**
+ * Check whether better-sqlite3 can actually be loaded by the current
+ * Node.js runtime.  A binding file may exist on disk but still fail
+ * with NODE_MODULE_VERSION mismatch after a Node.js or plugin update.
+ */
+function sqliteLoadsSuccessfully() {
+  try {
+    require(path.join(sqliteModulePath, "lib", "index.js"));
+    return true;
+  } catch (e) {
+    if (e && e.message && e.message.includes("NODE_MODULE_VERSION")) {
+      warn("ABI version mismatch — native module was built for a different Node.js version.");
+      return false;
+    }
+    // For any other load error fall through to the rebuild path as well.
+    return false;
+  }
+}
+
+let needsRebuild = false;
+
+if (!sqliteBindingsExist()) {
+  warn("better-sqlite3 native bindings not found in plugin dir.");
+  log(`Searched in: ${DIM}${sqliteModulePath}/build/${RESET}`);
+  needsRebuild = true;
+} else if (!sqliteLoadsSuccessfully()) {
+  log("Native binding file exists but cannot be loaded by current Node.js.");
+  needsRebuild = true;
+} else {
   ok("better-sqlite3 is ready.");
   console.log(`
 ${GREEN}${BOLD}  ┌──────────────────────────────────────────────────┐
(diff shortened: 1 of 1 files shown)


Hun-ger:
This issue has been fixed. You can try the latest version.

Issue #1311, opened 2026-03-20
bug(openclaw-plugin): auto-recall contaminates cron runs with prior meta-discussion


dreamcoin1998:
### Pre-submission checklist

- [x] I have searched existing issues and this hasn't been mentioned before
- [x] I have read the project documentation and confirmed this issue doesn't already exist
- [x] This issue is specific to MemOS and not a general software issue

### Bug Description

When `@memtensor/memos-local-openclaw-plugin` is used as the OpenClaw memory slot, cron runs can be contaminated by prior meta-discussion about the cron itself.

The plugin performs auto-recall in `before_agent_start` for every turn. For cron runs, it uses the cron prompt as the recall query, which causes it to retrieve previous conversations about prompt tuning, policy changes, debugging, and rerun requests for that same cron job.

Those recalled memories are then injected back into the scheduled run as system context, so the cron output drifts away from the actual task and starts replying with unrelated meta-discussion.

Observed effects include:

- cron output says things like “I already changed the cron”
- output summarizes previous human discussions about task policy
- output mentions rerun/debugging status instead of only executing the scheduled task
- session logs contain injected memory blocks beginning with:
  - `## User's conversation history (from memory system)`

This is especially harmful for autonomous scheduled jobs because it changes task semantics.

### How to Reproduce

1. Install and enable `@memtensor/memos-local-openclaw-plugin` as the OpenClaw memory slot.
2. Create an OpenClaw cron job with a stable name and repeated scheduled runs.
3. Run the cron normally at least once.
4. In regular chat, discuss this cron job several times, for example:
   - ask why its output is wrong
   - ask to rewrite its prompt or policy
   - discuss whether it should trade or not trade
   - ask for reruns, debugging, or cron body changes
5. Let the cron run again.
6. Observe that the cron output begins to include prior meta-discussion instead of only the intended scheduled task output.

A representative session key pattern is:

- `agent:main:cron:<jobId>`

Relevant observations during reproduction:

- runtime logs showed auto-recall queries derived from the cron prompt itself
- session logs showed injected memory blocks starting with:
  - `## User's conversation history (from memory system)`
- disabling the plugin and resetting the cron session immediately stopped the contamination

### Environment

- Repository: `MemTensor/MemOS`
- Plugin package: `@memtensor/memos-local-openclaw-plugin`
- Plugin version: `1.0.3`
- OpenClaw version: `v2026.3.11`
- Node.js runtime observed in gateway logs: `v22.20.0`
[...]

PR #1375, opened 2026-03-27
feat(openclaw-plugin): add configurable capture skipSessionPatterns

duoyidavid-eng:
## Summary

Allow users to configure session key substrings to skip memory capture via `capture.skipSessionPatterns` config option.

**Default:** `[":cron:"]` (backward compatible — cron sessions still skipped by default)

## Motivation

Previously, only cron sessions were hard-coded to skip capture. Users who run multiple agents (e.g. a dedicated wechat agent) had to modify source code to exclude specific agents from memory capture.

## Changes

| File | Change |
|------|--------|
| `src/types.ts` | Add `skipSessionPatterns?: string[]` to `MemosLocalConfig.capture` |
| `src/config.ts` | Resolve config with default `[":cron:"]` |
| `src/capture/index.ts` | Add `skipSessionPatterns` parameter to `captureMessages()`, filter before processing |
| `src/index.ts` | Pass config value to `captureMessages()` |

## Example config

```json
{
  "capture": {
    "skipSessionPatterns": [":cron:", "agent:wechat", "agent:bot"]
  }
}
```

This replaces the previous hard-coded cron check and makes session filtering data-driven.

### DIFF
--- a/apps/memos-local-openclaw/src/capture/index.ts
+++ b/apps/memos-local-openclaw/src/capture/index.ts
@@ -62,10 +62,19 @@ export function captureMessages(
   log: Logger,
   owner?: string,
   userSearchTime?: number,
+  skipSessionPatterns?: string[],
 ): ConversationMessage[] {
   const now = Date.now();
   const result: ConversationMessage[] = [];
   let lastTimestamp = 0;
+
+  // Skip capturing sessions matching configured patterns (e.g. [":cron:", "agent:wechat"]).
+  if (sessionKey && skipSessionPatterns && skipSessionPatterns.length > 0) {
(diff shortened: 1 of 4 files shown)


Hun-ger:
Hi @duoyidavid-eng , thanks for this great initiative! While this PR successfully prevents matched sessions (like cron) from being captured (written) into memory, it doesn't stop them from triggering auto-recall (reading). This means these background sessions will still perform unnecessary memory searches and suffer from context pollution (related to Issue #1311). To fully fix this, the filtering must apply to both phases. Our official team will take over to implement a comprehensive solution that blocks both capture and auto-recall based on your idea. We will likely close this PR in favor of the complete fix, but we truly appreciate your contribution and inspiration!

PR #1379, opened 2026-03-29
feat(openclaw-plugin): add topic-judge pre-filter to skip auto-recall on topic continuation

duoyidavid-eng:
## Problem

The `before_prompt_build` auto-recall hook can run a candidate search and relevance-filter LLM call even when a longer user message is clearly continuing the current topic.

## Solution

Add an **opt-in topic-judge pre-filter** using the existing `Summarizer.judgeNewTopic()` implementation. It runs after the existing cheap session and short-query gates, and before the current parallel local + Hub search flow.

**Flow:**
```
before_prompt_build fires
  -> Existing session and short-query gates
  -> If topicJudgeRounds is 0, proceed directly to recall
  -> Build context from the last N completed user/assistant rounds
  -> Strip injected metadata with the shared capture helper
  -> If context is insufficient, proceed with recall (fail open)
  -> SAME: skip auto-recall
  -> NEW: proceed with parallel local + Hub search
  -> Parse/provider error: proceed with recall (fail open)
```

## Config

`recall.topicJudgeRounds` defaults to `0`, so the extra LLM call is disabled unless explicitly enabled:

```json
{
  "plugins": {
    "entries": {
      "memos-local-openclaw-plugin": {
        "config": {
          "recall": {
            "topicJudgeRounds": 4
          }
        }
      }
    }
  }
}
```

| Value | Behavior |
|-------|----------|
| `0` (default) | Disabled; auto-recall follows the existing flow |
| `N > 0` | Uses up to the last N completed rounds for topic judgment |

## Implementation

- Extracts context building and topic-judge control flow into `src/recall/topic-judge.ts`.
- Reuses `stripInboundMetadata()` instead of duplicating injection markers.
- Preserves the current parallel local + Hub candidate search and single `filterRelevant` pass.
- Fails open for insufficient context, unparseable responses, and provider errors.
- Adds unit coverage for disabled/default behavior, SAME, NEW, null, thrown errors, insufficient context, metadata stripping, and config resolution.


shinetata (changes requested):
Thanks @duoyidavid-eng — the underlying problem is real (per-turn auto-recall carries non-trivial cost/latency) and reusing the already-present `judgeNewTopic()` is a clever idea. My feedback is mostly about the *shape* of the solution, not the goal. I'd suggest reshaping this before merge rather than landing as-is.

**1. Please don't enable it by default.** `topicJudgeRounds: 4` puts a synchronous LLM round-trip on the critical path of *every* turn. That works against #1452 (auto-recall blocking gateway startup / first-turn health checks): on a first turn / new topic, `judgeNewTopic` returns NEW and we still run recall — so it's strictly +1 LLM call on the slowest path. Suggest defaulting to `0` (opt-in).

**2. The saving is smaller than it looks.** The dominant cost is the LLM round-trip, not the local embedding + search. On a continuation turn we still pay one round-trip (`judgeNewTopic` instead of `filterRelevant`); on a new-topic turn it's pure overhead. So it's "swap one LLM call for a slightly smaller one + skip local search," not "save the ~2-3s pipeline."

**3. The motivating cases are largely covered already.** "ok" / typos / short follow-ups are already short-circuited by `recall.autoRecallMinQueryLength` (#1347), which runs before this pre-filter. After a rebase, `judgeNewTopic` only fires on longer same-topic messages — a narrower slice than the description implies.

**4. Product tension.** "Same topic" != "no memory needed" — same-topic deepening is often exactly when a memory from another past session helps. A coarse topic gate risks suppressing useful cross-session recall, which is the plugin's headline behavior; `filterRelevant` is more precise because it inspects the actual candidates.

If we keep it (as an opt-in fallback layered *after* the cheap gates), a few implementation must-dos:

- **Rebase** — the hook was rewritten since March (#1311 / #1347 / #1514 / #1532). The patch anchor `engine.search({ ... maxResults: 10 ... })` no longer exists; it's now a parallel local + hub search followed by a single `filterRelevant`. The PR is currently CONFLICTING.
- **Tests** — none yet. Note the green checks here are the Python build matrix; they do not run the `apps/memos-local-openclaw` vitest suite, so this change isn't actually exercised by CI. Please add unit tests (SAME / NEW / null branches, metadata stripping, parse-failure fallback), following e.g. `topic-judge-minimax-1315.test.ts` / `auto-recall-cron-gate.test.ts`.
- **Extract the pure logic** into `src/recall/` (mirroring `session-policy.ts`) instead of a ~100-line nested function in `index.ts`, and avoid the hardcoded injected-prompt string stripping — those markers are defined elsewhere and will drift.
- **Fail open** — too-few-lines / parse error should `proceed` (recall), not `skip`; for a memory system, skipping is the aggressive choice, and this also matches your own "LLM error -> proceed" rule.
- Minor: the description says `before_agent_start`, but the hook is `before_prompt_build`.

Happy to help with the rebase, or with an extracted `src/recall/topic-judge.ts` + tests, if that's useful.

Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #1379
**Task:** `82961631dccc9e41`
**Base:** `main`
**Head:** `feat/topic-judge-pre-filter`

---

### 1. `apps/memos-local-openclaw/index.ts` (L1853)

The `topicJudgeRounds` value is already resolved with its default by `resolveConfig` (see `src/config.ts`), so `ctx.config.recall` will never be `undefined` here and `?? DEFAULTS.topicJudgeRounds` is a dead fallback in practice. This is a minor inconsistency: other config values on the same surrounding lines (e.g. `autoRecallMinQueryLength` on line 1843) follow the same pattern, but it adds no safety value when going through `resolveConfig`. Consider reading `ctx.config.recall.topicJudgeRounds` directly (as the resolved config always provides it), or at minimum leave a comment explaining why the defensive fallback is retained.

---

### 2. `apps/memos-local-openclaw/src/recall/topic-judge.ts` (L69-L71)

Same-role message texts are merged without any size cap. Within the 20-message window, if multiple consecutive same-role messages carry large content, `previous.text` can grow unboundedly before `truncateContextLine` is applied at output time. Consider capping the accumulated text early (e.g., truncate at `MAX_CONTEXT_LINE_CHARS` immediately after each concatenation) to bound intermediate memory allocation.

---

### 3. `apps/memos-local-openclaw/src/recall/topic-judge.ts` (L91-L95)

The validation checks string prefixes on already-formatted output lines (`line.startsWith("USER:")`) rather than inspecting the structured `completed` array before formatting. If the prefix strings ever change (e.g., case or separator), this silently returns `null` for valid contexts. Prefer checking `completed.some(l => l.role === "user")` on the structured data before the `flatMap`.

---

### 4. `apps/memos-local-openclaw/src/recall/topic-judge.ts` (L7)

`judgeNewTopic` is typed as `Promise<boolean | null>` but the caller only branches on strict `=== false` and `=== true`, so `undefined` (not in the declared type but possible if an implementor is sloppy) silently falls through to the same `null` path. Document or enforce the invariant — e.g., rename `null` to a dedicated sentinel type or add a runtime assertion — so future implementors get a clear contract rather than accidentally-correct behavior.

---

### 5. `apps/memos-local-openclaw/src/recall/topic-judge.ts` (L82-L83)

`Array.shift()` is O(n) per call, making these trim loops O(n²) in the size of `completed`. While `completed` is currently bounded by `topicJudgeRounds * 2`, a large configured value amplifies the cost unnecessarily. Replace with index-based slicing: `const start = completed.findIndex(l => l.role === 'user'); const trimmed = start > 0 ? completed.slice(start) : completed;` and mirror the pattern for the trailing pop.

Memtensor-AI:

All tests passed (21/21 executed). memos_local_openclaw/unit: 21/21. Duration: 3s

**Branch:** `feat/topic-judge-pre-filter`

syzsunshine219:
Resolved the conflicts against the current `main` and addressed the requested changes:

- `topicJudgeRounds` is opt-in with a default of `0`.
- The pre-filter now layers onto the current parallel local + Hub recall flow.
- Topic/context logic lives in `src/recall/topic-judge.ts` and reuses `stripInboundMetadata()`.
- Insufficient context, parse/provider errors, invalid config, and context-construction errors all fail open.
- New sessions bypass topic judging.
- Unit coverage includes SAME, NEW, null, thrown errors, metadata stripping, completed-round handling, default/disabled config, and new-session behavior.

Current head: `4ceed826`. Python CI (16/16), Open Code Review, and AutoTest/Test Engine are all successful. Re-review requested.

Hun-ger:
The issue is valid as a cost/latency optimization, but the proposed topic-based gating may be unnecessarily complex and risks missing relevant memories during same-topic follow-ups.

Users who prefer fully agent-driven retrieval can already disable automatic recall with `hooks.allowPromptInjection: false` while keeping tools such as `memory_search` available for the model to call. A clearer solution may be to expose and document an explicit `autoRecall.enabled` option, independent of memory capture and tool registration, rather than adding another LLM call to decide whether recall should run.

PR #1424, opened 2026-04-05
fix(viewer): show session-expired toast when loadConfig gets 401

Danielwangyy:
## Problem

When a user's Viewer session expires and they navigate to the **Settings** tab, `loadConfig()` fetches `/api/config` which returns HTTP 401. The function silently returns on `!r.ok`, leaving all form fields at their empty/default values.

**User-visible symptoms:**
- Embedding model field appears blank (even though `openclaw.json` has a valid config)
- Clicking "Test Connection" shows: *"❌ 连接失败 — Provider and Model are required"*
- No indication that the actual issue is an expired session
- Users may think their configuration was lost after a plugin upgrade

## Root Cause

```js
// html.ts — loadConfig()
const r = await fetch('/api/config');
if (!r.ok) return;          // ← silently swallows 401
```

Other settings functions already handle this correctly:

| Function | 401 handling |
|---|---|
| `doSaveConfig()` | ✅ `toast(t('settings.session.expired'), 'error')` |
| `saveModelsConfig()` | ✅ `toast(t('settings.session.expired'), 'error')` |
| `testModel()` | ✅ Shows expired message in result element |
| **`loadConfig()`** | ❌ Silent return — **this PR** |

## Fix

One-line addition: check for 401 before the generic `!r.ok` guard, show the same session-expired toast that other functions use.

```diff
 async function loadConfig(){
   try{
     const r=await fetch('/api/config');
+    if(r.status===401){toast(t('settings.session.expired'),'error');return;}
     if(!r.ok) return;
```

No new i18n keys — reuses existing `settings.session.expired`.

## Testing

1. Open Viewer → Settings tab (should load normally)
2. Wait for session to expire (or manually delete the session cookie)
3. Navigate away from Settings and back
4. **Before fix:** Blank form, no error
5. **After fix:** Toast "登录已过期，请刷新页面重新登录"

Made with [Cursor](https://cursor.com)


angarchie:
Hey @Danielwangyy! Thanks for the contribution — already merged your PR, thanks for putting this together! 🚀

PR #1453, opened 2026-04-10
fix(viewer): singleton guard prevents port drift on reactivation

### LINKED ISSUE #1430: memos-local-openclaw-plugin@1.0.8 self-start viewer repeats under host lifecycle mismatch, causing port drift (18799→18800+), Memory unavailable

### Bug Description | 问题描述

Environment
OpenClaw: 2026.4.5
Plugin: @memtensor/memos-local-openclaw-plugin@1.0.8
OS: Windows 11
Node: 24.14.0
Config:
plugins.entries.memos-local-openclaw-plugin.config.viewerPort = 18799
Current behavior
Even with viewerPort fixed to 18799, Memory still becomes unavailable in status checks.

Observed logs repeatedly show:

memos-local: service.start() not called by host, self-starting viewer...
then port fallback:
Viewer port 18799 in use, trying 18800
Viewer port 18800 in use, trying 18801
...
openclaw status --deep often reports:

Memory: enabled (plugin memos-local-openclaw-plugin) · unavailable
Reproduction (minimal)
Install/update plugin to 1.0.8
Set fixed port:
Restart gateway
Run openclaw status --deep / openclaw plugins inspect ... multiple times
Observe self-start logs and viewer moving to 18800+, while Memory may stay unavailable
Expected behavior
If viewerPort=18799, plugin should either:
use exactly that port and fail clearly if occupied, or
keep health as available when core memory service is healthy.
Repeated plugin register/init in same process should not trigger repeated self-start attempts.
status --deep should not degrade Memory health due to duplicate self-start side effects.
Actual behavior
service.start() is not always called by host path.
Plugin fallback self-starting viewer... is triggered repeatedly.
Viewer fallback consumes next ports (18800+).
Memory health may show unavailable despite plugin DB/embedding being initialized.
Root-cause hypothesis
Lifecycle mismatch: plugin registerService occurs, but service.start is not guaranteed on all load paths.
Fallback self-start is not globally idempotent across repeated register/init cycles.
Health status likely coupled too tightly to viewer startup result.
Local workaround/patch used
I added a process-level singleton guard in index.ts to make startup idempotent per (pid,stateDir,viewerPort):

guard state on globalThis:
starting
started
skip duplicate starts when already starting/started
reset guard on stop()
This reduced repeated same-process viewer starts and stabilized fixed port behavior.

Patch idea (pseudo)
Copy
const globalState = globalThis as Record<string, any>;
const root = "__memos_local_start_guard__";
if (!globalState[root]) globalState[root] = {};
const key = `${process.pid}:${stateDir}:${viewerPort}`;
const singleton = (globalState[root][key] ||= { starting: false, started: false });

if (singleton.started) return;
if (singleton.starting) return;
singleton.starting = true;
try {
  await viewer.start();
  singleton.started = true;
} finally {
  singleton.starting = false;
}
Suggested upstream fix
Make service startup idempotent at plugin level across repeated register/init.
Revisit host lifecycle integration to avoid “registered but start hook not fired” path.
Decouple Memory health from viewer port fallback side-effects (or treat viewer as optional health dimension).
If viewerPort is explicitly configured, consider strict mode (no auto-drift) + explicit warning.
Extra notes
This is not a ClawHub rate-limit issue (already bypassed with npm install/update).
Not primarily embedding failure either; logs show embedding can initialize while health still unavailable.

### How to Reproduce | 如何重现

1

### Environment | 环境信息

1

### Additional Context | 其他信息

1

kagura-agent:
## Summary

Fixes #1430 — viewer self-starts repeatedly under host lifecycle mismatch, causing port drift (18799→18800+) and "Memory unavailable".

## Root Cause

When the gateway restarts or reloads, `register()` is called again. Each call creates a new `ViewerServer` instance, but the previous one may still hold the port. The new instance hits `EADDRINUSE` and falls back to port+1, eventually drifting away from the expected 18799.

## Fix

- **Module-level singleton**: Track the active viewer, hub server, and store instances outside `register()` scope
- **Cleanup on re-registration**: `stop()` the previous viewer/hub and `close()` the previous store before creating new ones
- **Reliable stop**: `ViewerServer.stop()` now returns a `Promise<void>` that resolves only after `server.close()` callback fires, ensuring the port is fully released

## Scope

This PR addresses the **plugin side** of the lifecycle mismatch. The host-side gap (deferred `service.start()` call path) would need a separate OpenClaw upstream fix.

## Testing

- `npm run build` passes

### DIFF
--- a/apps/memos-local-openclaw/index.ts
+++ b/apps/memos-local-openclaw/index.ts
@@ -33,6 +33,11 @@ import { MEMORY_GUIDE_SKILL_MD } from "./src/skill/bundled-memory-guide";
 import { Telemetry } from "./src/telemetry";
 
 
+// Module-level singleton tracking for viewer/hub to prevent port drift on re-registration
+let _previousViewer: ViewerServer | null = null;
+let _previousHubServer: HubServer | null = null;
+let _previousStore: import('./src/storage/sqlite').SqliteStore | null = null;
+
 /** Remove near-duplicate hits based on summary word overlap (>70%). Keeps first (highest-scored) hit. */
 function deduplicateHits<T extends { summary: string }>(hits: T[]): T[] {
   const kept: T[] = [];
@@ -2365,6 +2370,20 @@ Groups: ${groupNames.length > 0 ? groupNames.join(", ") : "(none)"}`,
 
     const derivedHubPort = gatewayPort + 11;
 
+    // Cleanup previous instances to prevent port drift (issue #1430)
+    if (_previousViewer) {
+      try { _previousViewer.stop(); } catch (e) { api.logger.warn(`memos-local: previous viewer cleanup: ${e}`); }
+      _previousViewer = null;
+    }
+    if (_previousHubServer) {
+      try { _previousHubServer.stop(); } catch (e) { api.logger.warn(`memos-local: previous hub cleanup: ${e}`); }
+      _previousHubServer = null;
+    }
+    if (_previousStore) {
+      try { _previousStore.close(); } catch (e) { api.logger.warn(`memos-local: previous store cleanup: ${e}`); }
+      _previousStore = null;
+    }
+
     const viewer = new ViewerServer({
       store,
       embedder,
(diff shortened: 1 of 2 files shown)


kagura-agent:
Closing — happy to reopen if there's interest in this fix. Thanks!

PR #1466, opened 2026-04-10
fix(memos-local-openclaw-plugin): integrate stfix(memos-local): report memory status correctly and avoid CLI-spawned viewersatus runtime and gate viewer self-start

LEVO-Hephaestus:
## Summary

This PR fixes two integration issues in the MemOS local OpenClaw plugin:

1. Register an OpenClaw memory runtime adapter so host status probes can read plugin health correctly
2. Gate viewer service registration and self-start logic to real gateway processes only

## Problem
The plugin could still read and write memory correctly, but `openclaw status` reported memory as `unavailable`.

At the same time, CLI processes such as `openclaw status` could accidentally trigger the plugin's viewer self-start fallback, which caused extra viewer instances to be spawned and fallback ports to be consumed.

## Changes

- add `registerMemoryRuntime(...)` so OpenClaw can query memory status through the standard runtime interface
- expose plugin-backed memory status details such as files, chunks, sources, and vector readiness
- restrict viewer service registration to actual gateway processes
- prevent non-gateway CLI runs from triggering viewer self-start
- keep the fix limited to plugin code only

## Result

After this change:
- `openclaw status` reports normal memory status instead of `unavailable`
- memory status can show values such as `files`, `chunks`, `sources sessions`, and `vector ready`
- CLI commands no longer spawn extra viewer instances
- the memory viewer lifecycle stays bound to the gateway process as expected

## Notes

This PR does not include any local machine config workarounds or user-specific environment changes.
It only addresses the plugin-side integration and lifecycle issues.

## Validation

Validated locally by confirming that:
- memory read/write was still working before and after the fix
- `openclaw status` no longer reported `unavailable`
- status output showed plugin-backed memory health information
- non-gateway CLI commands no longer created additional viewer listeners

### DIFF
--- a/apps/memos-local-openclaw/index.ts
+++ b/apps/memos-local-openclaw/index.ts
@@ -333,6 +333,86 @@ const memosLocalPlugin = {
     } catch {}
     const telemetry = new Telemetry(ctx.config.telemetry ?? {}, stateDir, pluginVersion, ctx.log, pluginDir);
 
+    api.registerMemoryRuntime({
+      async getMemorySearchManager(params: { agentId: string }) {
+        const ownerFilter = [`agent:${params.agentId}`, "public"];
+        const readCount = (sql: string) => {
+          try {
+            const db = (store as { db?: { prepare: (query: string) => { get: (...args: string[]) => { count?: number } | undefined } } }).db;
+            const row = db?.prepare(sql).get(...ownerFilter);
+            return Number(row?.count ?? 0);
+          } catch {
+            return 0;
+          }
+        };
+
+        return {
+          manager: {
+            async search(query: string, opts?: { maxResults?: number; minScore?: number }) {
+              const result = await engine.search({
+                query,
+                maxResults: opts?.maxResults,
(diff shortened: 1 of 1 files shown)


Hun-ger:
Thanks for the PR! 🙏

A couple of things to flag:

Multiple window instances — already fixed in our latest release, so this part isn't needed anymore.
openclaw status still shows unavailable — tested on the latest OpenClaw and this PR doesn't resolve it. The issue is actually upstream: the status scan gates on either ~/.openclaw/memory/main.sqlite existing or an explicit agents.defaults.memorySearch key being present in config — before it ever reaches the plugin runtime. Since MemOS Local uses its own DB path (~/.openclaw/memos-local/memos.db), both checks fail and the scan exits early. The registerMemoryRuntime call added here never gets invoked. We'll handle the fix on the plugin side.

Really appreciate you digging into this! 🙌 Love seeing the initiative and the ideas you're bringing—please keep them coming and feel free to continue contributing!

PR #1481, opened 2026-04-14
feat(memos-local-openclaw): add sqlite-vec acceleration and embedding cache

abakane1:
## Summary

Add optional `sqlite-vec` indexed vector search, a signature-aware query embedding cache, and an Ollama embedding provider to `memos-local-openclaw-plugin`.

This branch has been merged with the latest `main`, and the previous review blockers have been addressed.

### sqlite-vec indexed search

- Load `sqlite-vec` through the existing native dependency checks, with a verified fallback when the extension is unavailable.
- Create the `vec0` table using the configured `Embedder.dimensions` and `distance_metric=cosine`.
- Store `owner` and `session_key` metadata in the index so access and session filters are applied during KNN search.
- Keep the existing `embeddings` table as the source of truth and fallback path.
- Backfill or repair the index when it is created, dimensions change, or indexed row counts drift.
- Preserve exact `vectorSearchMaxChunks` behavior by using the existing brute-force path when a bounded candidate scan is requested.
- Fall back through the injected logger on extension, synchronization, or query-dimension errors.
- Keep index rows synchronized across upsert, deduplication, session deletion, and full deletion flows.

Set `MEMOS_USE_VEC_INDEX=false` to force the original brute-force search path.

### Embedding cache

- Add a bounded LRU cache with TTL.
- Key entries by `provider:model:dimensions` plus the exact query text.
- Share the cache across `Embedder` instances and return cloned vectors.
- Do not cache provider fallback vectors whose dimensions do not match the configured embedder.

### Ollama provider

- Add batched `/api/embed` support with `nomic-embed-text` as the default model.
- Validate response shape and batch size, report HTTP failures, and normalize endpoints with trailing slashes.

### Regression fixes

- Preserve descending cosine-score ordering for skill vector candidates.
- Avoid creating or mutating the sqlite-vec schema during `re-embed --dry-run`.

## Validation

- Focused regression suite: 73 tests passed across cache, Ollama, sqlite-vec, recall, native packaging, and re-embedding coverage.
- Integration suite: 26 tests passed, including owner isolation and hub flows.
- Added indexed-vs-brute score/ranking parity, dimension mismatch fallback, metadata filtering, backfill/repair, cache TTL/LRU/signature, Ollama error handling, and native loading tests.

## Backward compatibility

- Existing data remains in the regular `embeddings` table.
- Missing or incompatible native binaries disable the index without disabling recall.
- Existing databases are backfilled automatically after the index is enabled.


shinetata (changes requested):
Thanks for the contribution! The direction (indexed vector search + query-embedding cache + Ollama) is welcome, but this needs changes before it can land. Key blockers:

1. **Sort regression (`src/recall/engine.ts`)**: skill vector candidates are now sorted with `(a, b) => b.score - b.score`, which is always 0 — this disables ranking. Must be `b.score - a.score`.
2. **Hardcoded `FLOAT[2048]` (`src/storage/sqlite.ts`)**: default embeddings are 384-D (local) or 1536-D (remote), never 2048. So `vec_chunks` stays empty and every query throws + falls back to brute force. Derive the dimension from `Embedder.dimensions` instead of hardcoding.
3. **Distance→score (`src/storage/vector.ts`)**: vec0 defaults to L2, so `1 - distance` is not cosine similarity and is inconsistent with the brute-force path — it breaks `minScore`/RRF. Please declare `distance_metric=cosine` and convert accordingly.
4. **Cache key (`src/embedding/cache.ts`)**: keyed on text only. It must incorporate `embeddingSignature` (`provider:model:dimensions`), otherwise cached vectors are reused across models.
5. **Native dependency**: `sqlite-vec` ships native binaries. Please wire it through the existing postinstall/prebuild + ESM packaging flow (see the #1733 module-format tests and the better-sqlite3 native-binding setup).
6. **Tests**: no tests were added. Please cover cache hit/expiry/LRU/model-keying, indexed-vs-brute-force score parity, dimension-mismatch fallback, and the Ollama provider.

Also minor: use the injected `Logger` instead of `console.warn`; the indexed path ignores `vectorSearchMaxChunks`; Ollama default model `"qwen"` isn't an embedding model; please avoid non-English inline comments.

Suggestion: split into 3 smaller PRs (Ollama provider / embedding cache / sqlite-vec index) and rebase onto `main` (currently conflicting). Happy to re-review once updated.

syzsunshine219:
@shinetata The branch is now updated on top of the latest `main` and ready for re-review.

The previous blockers are addressed in `2afd7b2b`:

- fixed skill candidate sorting (`b.score - a.score`)
- derived vec dimensions from `Embedder.dimensions`
- configured cosine distance and verified score/ranking parity with brute force
- keyed the cache by `provider:model:dimensions` plus query text
- integrated and tested the `sqlite-vec` native dependency loading path
- added cache, sqlite-vec, Ollama, packaging, and recall regression coverage
- preserved `vectorSearchMaxChunks`, injected logger behavior, and a valid Ollama embedding default

Local validation completed with 73 focused tests and 26 integration tests passing. The PR description has also been updated to match the current implementation.

syzsunshine219:
### Automated Test Results: **PASSED**

Manual cloud Test Engine rerun against the current PR head completed successfully after the automated OCR gate failed to start AutoTest.

- Run: `tr-d29c3739-a99` on cloud Test Engine `10010`
- Head: `2afd7b2b88b253cba9466f5f069eeb4505747761`
- Base: latest `main`
- `memos_local_openclaw/unit`: 51 passed, 0 failed, 0 skipped
- Verdict: `pass`
- Duration: 10s

The `open-code-review` service itself failed twice at approximately 17 minutes with `Open Code Review execution failed; retry required`; no OCR finding was produced. That remains an OCR infrastructure blocker separate from the passing code tests above.

Memtensor-AI:

All tests passed (51/51 executed). memos_local_openclaw/unit: 51/51. Duration: 11s

**Branch:** `feat/sqlite-vec-fast-search`

Hun-ger:
Thanks for the contribution and for addressing the previous review feedback. We are closing this PR because `apps/memos-local-openclaw` is no longer the active plugin implementation. Future plugin development is now based on `apps/memos-local-plugin`, which already includes embedding caching and bounded streaming vector search, while Ollama can be used through the OpenAI-compatible embedding provider.

Any future work on indexed or ANN-based vector retrieval should target `apps/memos-local-plugin`. Closing this PR as superseded.

PR #1498, opened 2026-04-19
feat(add): chunk long content in fast mode (~500 token chunks, 50 overlap)

sergiocoding96:
## Summary
- Fast mode previously stored a long message as a SINGLE `TextualMemoryItem` with one giant embedding, washing out late content during semantic search.
- `_process_chat_data` fast branch now sub-chunks each window when its token count exceeds `2 * MOS_FAST_CHUNK_TOKENS` (default threshold 1000). Each sub-chunk becomes its own item with `chunk_index` / `chunk_total` in `metadata.info`.
- New env knobs: `MOS_FAST_CHUNK_TOKENS` (default 500), `MOS_FAST_CHUNK_OVERLAP_TOKENS` (default 50).
- Adds `chunk_text_by_tokens()` in `mem_reader/utils.py` — paragraph→sentence→token fallback, reuses existing tiktoken-based counter (no LLM in the loop).
- Fine mode untouched (already per-fact extraction via DeepSeek).

## Design choices
- **Tokenizer**: reused `count_tokens_text` (tiktoken, char-proxy fallback) — already imported by the reader, accurate for OpenAI-style tokens, free of new deps.
- **Threshold**: chunk only when `tokens > 2 * MOS_FAST_CHUNK_TOKENS` so docs that fit in two chunks aren't chunked unnecessarily; small docs keep unchanged single-memory behavior.
- **Timestamp prefix**: `_iter_chat_windows` already prepends `user: [chat_time]: ` to the window text. When a window is sub-chunked, that prefix naturally lands on the FIRST sub-chunk only; later sub-chunks are mid-message slices. Documented inline in `simple_struct.py`.
- **Dedup**: chunking is deterministic (same input + same env vars → same chunks → same embeddings), so the existing write-time dedup in `single_cube._process_text_mem` catches every chunk on a re-write.

## Test plan
- [x] `tests/mem_reader/test_fast_mode_chunking.py` — 10 unit tests, all green:
  - chunker: short→single, empty→none, long→multi, overlap present, deterministic, giant single sentence falls through to token slice
  - fast-mode expansion: short→1 memory (no chunk metadata), long→multiple memories with consistent `chunk_total` and contiguous `chunk_index`, env override changes chunk count, repeat write produces identical chunk text (dedup-safe)
- [x] No regression in existing `tests/mem_reader/test_simple_structure.py`, `test_memory.py`, `test_coarse_memory_type.py` (15 total tests pass when run together)
- [ ] End-to-end smoke per TASK.md test plan (curl against running gateway): defer to reviewer / staging — fast mode already exercised in the changed unit tests; live run requires the full server stack.

### DIFF
--- a/PATCHES.md
+++ b/PATCHES.md
@@ -1,0 +1,79 @@
+# Security & Tuning Patches
+
+This fork maintains a set of patches on top of upstream `MemTensor/MemOS` for
+production deployment inside the Hermes multi-agent system.
+
+## Why a fork
+
+Upstream is a reference implementation. We need:
+- Per-agent API key auth with spoof detection and cube-level ACL
+- Hardened defaults (localhost binding, required passwords, secret encryption)
+- Search recall tuning (was returning 1/12 relevant results on vague queries)
+- Memory extraction quality fixes (English enforcement, granularity rule)
+
+See the full audit history at
+`memos-setup/learnings/` in the [hermes-multi-agent](https://github.com/sergiocoding96/hermes-multi-agent)
+repo.
+
+## Patched Files
+
+| File | Category | What changed |
+|------|----------|-------------|
+| `src/memos/api/middleware/agent_auth.py` | **NEW** | Per-agent Bearer-key auth middleware, bcrypt v2 format, auth-failure rate limit, auto-reload on config mtime change |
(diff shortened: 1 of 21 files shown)


sergiocoding96:
Opened against wrong repo by mistake — reopening on the fork.

PR #1500, opened 2026-04-19
fix(delete): accept singular+plural params and return distinct status codes

sergiocoding96:
## Summary
- `DELETE /product/delete_memory` now accepts both singular (`mem_cube_id`, `memory_id`) and plural (`writable_cube_ids`, `memory_ids`) forms; plural wins when both are provided. Singular is coerced into the plural internally by the `DeleteMemoryRequest` validator.
- Endpoint responds on both `POST` and `DELETE` so existing callers keep working.
- Status codes are now distinct:
  - **400** — memory-id mode intended but cube or memory id(s) missing
  - **403** — caller cannot access a targeted cube (unchanged)
  - **404** — every requested memory id is missing from the targeted cube(s) (previously returned 403 when the cube was wrong)
  - **200** with `data.deleted` + `data.not_found` arrays on partial success (don't 404 the whole request when some ids exist)
- Handler's mode-precedence was relaxed: explicit `memory_ids`/`file_ids` now take priority over filter/user_id/session_id mode, so callers can pass `user_id` alongside `memory_ids` as identity context without hitting the old "multiple modes" rejection.

Fixes the confusing delete-endpoint behavior flagged in blind-audit-report.md § 12 + Bug 2.

## Test plan
Unit + FastAPI TestClient coverage:
- `tests/api/test_memory_handler_delete.py` (10 tests, was 7) — singular/plural coercion, `memory_ids + user_id` uses memory-id mode, `memory_ids + file_ids` still rejected.
- `tests/api/test_server_router.py::TestServerRouterDeleteMemory` (new, 9 tests) — legacy plural form over POST still works, singular form over DELETE works, 404 for missing memory, 404 for memory in a different cube (the Bug 2 repro), 403 for forbidden cube, 200 with split for partial delete, 400 for empty body / cube-only / memory-only, filter mode untouched.

Command: `PYTHONPATH=src python3.12 -m pytest tests/api/ -q`
Result: **41 passed** (was 32; new tests cover the behavior changes).

The full flow described in TASK.md's test plan (write → delete by plural → delete by singular → 404 on unknown id → 200 partial-delete) is covered by the router tests using a mocked `graph_db` and `user_manager`; live-curl verification requires a running MemOS + Neo4j stack which wasn't exercised here.

### DIFF
--- a/PATCHES.md
+++ b/PATCHES.md
@@ -1,0 +1,79 @@
+# Security & Tuning Patches
+
+This fork maintains a set of patches on top of upstream `MemTensor/MemOS` for
+production deployment inside the Hermes multi-agent system.
+
+## Why a fork
+
+Upstream is a reference implementation. We need:
+- Per-agent API key auth with spoof detection and cube-level ACL
+- Hardened defaults (localhost binding, required passwords, secret encryption)
+- Search recall tuning (was returning 1/12 relevant results on vague queries)
+- Memory extraction quality fixes (English enforcement, granularity rule)
+
+See the full audit history at
+`memos-setup/learnings/` in the [hermes-multi-agent](https://github.com/sergiocoding96/hermes-multi-agent)
+repo.
+
+## Patched Files
+
+| File | Category | What changed |
+|------|----------|-------------|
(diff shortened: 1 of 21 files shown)


sergiocoding96:
Closing — opened against the wrong repo. Reopening on sergiocoding96/MemOS.

PR #1524, opened 2026-04-22
fix(utils): re-raise in timed_with_status when no fallback configured

### LINKED ISSUE #1523: [Bug] timed_with_status decorator silently swallows exceptions and returns None
## Summary

`timed_with_status` in `src/memos/utils.py` catches every exception, but when no `fallback` callable is configured the wrapper falls through to an implicit `return None`. Decorated functions therefore return `None` on any failure instead of raising, which masks the real error.

## Where

`src/memos/utils.py`, lines 43-56 (current `main` @ cddc252).

```python
try:
    result = fn(*args, **kwargs)
    success_flag = True
    return result
except Exception as e:
    exc_type = type(e)
    stack_info = "".join(traceback.format_stack()[:-1])
    exc_message = f"{stack_info}{traceback.format_exc()}"
    success_flag = False

    if fallback is not None and callable(fallback):
        result = fallback(e, *args, **kwargs)
        return result
    # ← no `raise` here; wrapper falls through to implicit return None
finally:
    ...
```

## Impact

`OpenAILLM.generate` is decorated with `@timed_with_status(...)`. When the upstream LLM returns 4xx/5xx (e.g. MiniMax 400 `chat content is empty (2013)` for a system-only message), the BadRequestError is caught, logged as `status: FAILED`, and then **swallowed**. `generate()` returns `None` to its caller.

Downstream that `None` flows into `clean_json_response(response)` (`src/memos/mem_os/utils/format_utils.py:1403`) and crashes with:

```
```

Two consequences:

1. **The user sees a confusing AttributeError** instead of the real 400 from the LLM. Diagnosis is hard because nothing in the traceback names the LLM call.
2. **It is a silent fail by API contract.** Callers cannot tell whether `generate()` succeeded with empty output or failed with an exception, because the same `None` represents both.

## Reproduction

1. Set `MOS_CHAT_MODEL=MiniMax-M2.7`, `OPENAI_API_BASE=https://api.minimax.io/v1`, valid `OPENAI_API_KEY`.
2. Start memos server, ensure default cube exists.
3. `POST /product/suggestions` with a `mem_cube_id` whose recent memories are empty (so the suggestion prompt has only a system message).
4. Observe HTTP 500 `'NoneType' object has no attribute 'replace'` in the response, and `[TIMER_WITH_STATUS] OpenAI LLM took 5051 ms, status: FAILED, error_type: BadRequestError, error_message: ...` immediately above it in the log.

## Proposed fix

Add an explicit `raise` after the fallback branch:

```python
                if fallback is not None and callable(fallback):
                    result = fallback(e, *args, **kwargs)
                    return result
                raise
            finally:
                ...
```

This preserves existing fallback semantics and makes the no-fallback path fail-fast.

I will open a PR with this change against `main`.

## Related

The same swallow likely contributes to other reports where downstream code receives unexpected `None`/empty values from LLM helpers (e.g. #1324 `memory_search always returns no results with reasoning-enabled models` — different root cause, but the same pattern of LLM-call failure being invisible to the caller).

Ptah-CT:
Fixes #1523.

## Problem

`timed_with_status` caught every exception, ran the optional `fallback`, but when no fallback was configured the wrapper fell through to an implicit `return None`. Decorated functions returned `None` on any failure instead of raising.

Concrete consequence: `OpenAILLM.generate` (decorated with this) returned `None` on a MiniMax 400 (`chat content is empty (2013)` for system-only messages), and the suggestion handler then died in `clean_json_response` with `AttributeError: 'NoneType' object has no attribute 'replace'`. The real 400 was hidden.

## Change

Add a single `raise` after the fallback branch:

```python
                if fallback is not None and callable(fallback):
                    result = fallback(e, *args, **kwargs)
                    return result
                # No fallback configured -> re-raise so callers see the failure
                # instead of receiving an implicit None. ...
                raise
            finally:
                ...
```

`finally` still runs, so the timing/status log line is unchanged.

## Type

## Tested

- Reproduced locally with `MOS_CHAT_MODEL=MiniMax-M2.7`, `OPENAI_API_BASE=https://api.minimax.io/v1`, system-only suggestion prompt.
  - **Before:** HTTP 500 `'NoneType' object has no attribute 'replace'`
  - **After:** HTTP 500 surfaces the real `openai.BadRequestError: 400 ... invalid params, chat content is empty (2013)` to the caller, exactly as `[TIMER_WITH_STATUS]` already logged.
- Existing fallback path (`fallback=` arg supplied) is exercised by other handlers; behaviour unchanged because `raise` is reached only when no fallback returns.

## Checklist

- [x] Self-review done
- [x] Comment added explaining the fix
- [ ] Unit test added (existing test suite did not cover this branch; happy to add a small one if maintainers want it)
- [x] Linked to issue

### DIFF
--- a/src/memos/utils.py
+++ b/src/memos/utils.py
@@ -53,6 +53,11 @@ def timed_with_status(
                 if fallback is not None and callable(fallback):
                     result = fallback(e, *args, **kwargs)
                     return result
+                # No fallback configured -> re-raise so callers see the failure
+                # instead of receiving an implicit None. Otherwise downstream
+                # parsers (e.g. clean_json_response) crash with confusing
+                # AttributeErrors that hide the real LLM/HTTP error.
+                raise
             finally:
                 elapsed_ms = (time.perf_counter() - start) * 1000.0
 


Memtensor-AI:
Closure review result: after merging dev-v2.0.22 locally, this PR has no remaining diff against dev-v2.0.22. The timed_with_status no-fallback re-raise behavior is already present on dev, so this PR is superseded and does not need another cloud test run.\n\nNote: pushing the merge commit back to the contributor branch was blocked by GitHub because the current OAuth token lacks workflow scope and the dev merge includes workflow-file history; closing based on the empty-diff audit.

PR #1526, opened 2026-04-22
fix(format_utils): raise ValueError on None in clean_json_response

### LINKED ISSUE #1525: [Bug] clean_json_response crashes with cryptic AttributeError when given None
## Summary

`clean_json_response(response)` in `src/memos/mem_os/utils/format_utils.py:1393` calls `response.replace(...)` unconditionally. When `response` is `None` it dies with:

```
```

The traceback points to `format_utils.py:1403` and gives no hint about the real problem (an upstream LLM call returning `None`).

## How `None` reaches here

In the codebase today this is reachable via the suggestion endpoint:

```
POST /product/suggestions
→ suggestion_handler.handle_get_suggestion_queries
→ llm.generate(message_list)            # OpenAILLM.generate, decorated with @timed_with_status
→ (LLM call raises BadRequestError)
→ timed_with_status catches, no fallback configured, falls through to implicit `return None`
→ clean_json_response(None)
→ AttributeError ← user sees this
```

The deeper bug is in `timed_with_status` (filed separately as #1523). Even after that is fixed, however, defending against `None` here is cheap and turns the worst possible diagnostic experience (a wrong-line AttributeError) into a clear message that names the actual root cause.

## Proposed fix

```python
if response is None:
    raise ValueError(
        "clean_json_response received None — upstream LLM call likely "
        "failed silently (check timed_with_status / generate() error handling)."
    )
return response.replace("```json", "").replace("```", "").strip()
```

PR follows.

Ptah-CT:
Fixes #1525.

## Problem

`clean_json_response(None)` raises `AttributeError: 'NoneType' object has no attribute 'replace'` — the traceback points to `format_utils.py` and tells the operator nothing about the real cause (upstream LLM call returned `None`).

## Change

Add an explicit `None` guard that raises a descriptive `ValueError`. Existing call-sites already wrap their dispatch in `try/except` (e.g. `suggestion_handler`'s outer try), so callers continue to bubble a 500 — but with a message that names the root cause.

```python
if response is None:
    raise ValueError(
        "clean_json_response received None — upstream LLM call likely "
        "failed silently (check timed_with_status / generate() error handling)."
    )
return response.replace("```json", "").replace("```", "").strip()
```

## Why also patch here, not just the upstream silent-fail

PR #1524 (against #1523) fixes the silent-fail in `timed_with_status` so future failures will raise instead of returning `None`. Even with that landed, defensive coding here is cheap and avoids re-introducing the same misleading symptom should another path ever hand `None` into `clean_json_response`.

## Type

## Tested

- Reproduced with `MOS_CHAT_MODEL=MiniMax-M2.7`, system-only suggestion prompt → `clean_json_response(None)` → fails with clear ValueError mentioning `timed_with_status / generate()`.

## Checklist

- [x] Self-review
- [x] Comment / docstring updated
- [ ] Unit test (trivial; happy to add `pytest.raises(ValueError)` if maintainers want it)
- [x] Linked to issue


angarchie:
Hey @Ptah-CT! Thanks for the contribution — already merged your PR, thanks for putting this together! 🚀

PR #1541, opened 2026-04-24
fix(auto-recall): respect recall.maxResultsDefault config instead of hardcoding 10

### LINKED ISSUE #1514: Auto-recall hardcodes maxResults: 10, ignoring recall.maxResultsDefault config
## Problem

The auto-recall feature in the `before_prompt_build` hook currently hardcodes `maxResults: 10` for both local and Hub memory searches:

https://github.com/MemTensor/MemOS/blob/main/apps/memos-local-openclaw/index.ts#L1893-L1896

This means the `recall.maxResultsDefault` configuration option is effectively ignored during automatic recall, even though users may set it to a lower value (e.g., 3 or 5) to reduce token consumption.

## Current Behavior

- `memory_search` tool correctly respects `maxResultsDefault` (default 6)
- Auto-recall always fetches **10 results** from local + **10 from Hub**, regardless of config
- After LLM filtering and deduplication, this still injects ~5-10 memory entries into every system prompt
- At ~300-500 tokens per entry, this adds ~2,000-5,000 tokens per turn

## Expected Behavior

Auto-recall should either:
1. Read from `recall.maxResultsDefault` config instead of hardcoding 10, OR
2. Support a separate `recall.autoRecallMaxResults` config key for independent control

## Suggested Fix

```typescript
// index.ts ~1893
const recallCfg = ctx.config.recall ?? {};
const autoRecallMax = recallCfg.maxResultsDefault ?? 10;

const arLocalP = engine.search({
  query,
  maxResults: autoRecallMax,
  minScore: 0.45,
  ownerFilter: recallOwnerFilter
});
```

## Environment

- Plugin version: 1.0.8
- OpenClaw version: 2026.4.x

Thank you for the excellent plugin!

octo-patch:
Fixes #1514

## Problem

The auto-recall `before_prompt_build` hook hardcodes `maxResults: 10` for both local and Hub memory searches, ignoring the user-configured `recall.maxResultsDefault` value. This means users who set a lower limit (e.g. 3 or 5) to reduce token consumption still get up to 10+10 candidates fetched on every turn.

## Solution

Read `ctx.config?.recall?.maxResultsDefault ?? DEFAULTS.maxResultsDefault` (default: 6) and use it for both the local engine search and Hub search calls.

```ts
const arMaxResults = ctx.config?.recall?.maxResultsDefault ?? DEFAULTS.maxResultsDefault;
const arLocalP = engine.search({ query, maxResults: arMaxResults, ... });
const arHubP = ctx.config?.sharing?.enabled
  ? hubSearchMemories(store, ctx, { query, maxResults: arMaxResults, ... })
  ...
```

## Testing

- Configured `recall.maxResultsDefault: 3` in plugin config; verified auto-recall now fetches at most 3 results from local search instead of 10.
- Default behavior (no config set) uses `DEFAULTS.maxResultsDefault` (6), unchanged from existing behavior intent.


Memtensor-AI:
AutoDev closure update: after rebasing/merging dev-v2.0.22, this PR has 0 changed files. The auto-recall max-results behavior is already covered by the current base implementation, so there is no remaining diff to test or merge. Closing as superseded/no-op.

PR #1547, opened 2026-04-27
fix(auth): startup gate + rate-limit MEMOS_REDIS_URL + key-prefix BCrypt lookup

sergiocoding96:
## Summary

Three audit findings on the auth surface, one PR. Each stands alone but they share the same files — easier to review together.

| | Bug | File | Before | After |
|---|---|---|---|---|
| 1 | Server starts with empty/missing agent registry | `server_api.py` | every authed request 401s silently | `sys.exit(2)` + clear FATAL on stderr |
| F-04 | Hard-coded `redis://redis:6379` floods logs | `rate_limit.py` | WARN per request | one-shot startup WARN + SQLite fallback |
| F-01 | BCrypt-walks all N agents per request | `agent_auth.py` | 3.6s/bad-key | 0.00ms/bad-key (bucket miss → no BCrypt) |

## (a) Deployment starts successfully with non-empty registry

Smoke against the live `agents-auth.json` (15 agents, schema v2):

| case | env | expected | actual |
|---|---|---|---|
| 1 | `MEMOS_AUTH_REQUIRED=false` | exit 0 | ✅ exit 0 |
| 2 | required + valid file | exit 0 | ✅ exit 0 |
| 3 | required + path unset | exit 2 + FATAL | ✅ exit 2 + `FATAL: MEMOS_AUTH_REQUIRED=true but MEMOS_AGENT_AUTH_CONFIG is unset…` |
| 4 | required + missing file | exit 2 + FATAL | ✅ exit 2 + `FATAL: …agent-auth file is missing at '/tmp/nope-12345.json'…` |
| 5 | required + `{"agents":[]}` | exit 2 + FATAL | ✅ exit 2 + `FATAL: …contains zero agent keys…` |
| 6 | required + malformed JSON | exit 2 + FATAL | ✅ exit 2 + `FATAL: …unreadable or not valid JSON (JSONDecodeError…)…` |

## (b) Bad-key DoS timing — before/after

In-process bench against the live 15-agent registry (`MEMOS_AGENT_AUTH_CONFIG=/home/openclaw/Coding/Hermes/agents-auth.json`):

```
loaded agents=15  buckets=14  unprefixed=0

=== AFTER (bucket lookup) ===
  bad-key x20         p50=    0.00ms  max=    0.02ms  n=20
  valid-key cold      p50=  238.35ms  (1 BCrypt: bucket has 1 candidate)
  valid-key warm x20  p50=    0.00ms  (cache hit, fix/auth-perf)

=== BEFORE (forced O(N) walk via empty prefix index) ===
  bad-key x10         p50= 3601.11ms  max= 3613.68ms  n=10
```

Bad-key drop: **3601ms → 0.00ms**, comfortably beating the <300ms target. Zero BCrypts on a bucket miss because no candidate shares the bad key's prefix.

## (c) Rate-limit trigger behaviour — before/after

**Before:** Redis DNS-failure WARN fired *per request* and state was per-process. Bad-key DoS made each attempt take 3.6s, so the 60s window expired before 11 attempts could accumulate → 429 never fired.

**After:**
- One startup WARN with no Redis: `[RateLimit] MEMOS_REDIS_URL is unset. Running in SQLite-fallback mode at /var/tmp/memos-ratelimit.db…`
- Sentinel `_REDIS_DISABLED` short-circuits subsequent `_get_redis()` calls — confirmed exactly **1 emission across 201 calls** (1 init + 200 hits) in the bench.
- SQLite fallback: 200 ops in 18ms (0.09ms/op), 100 allowed + 100 denied at `RATE_LIMIT=100`. State survives close→reopen (100 rows persisted).
- Combined with F-01: a bad-key flood now completes attempts fast enough that 11 attempts in 60s reliably triggers 429.

## (d) Confirmation: `agents-auth.json.archived` removed from repo root

Handled in the Hermes companion PR — see cross-link below.

## Relationship to `fix/auth-perf` (PR #4, already merged)

That PR added an `sha256(raw_key) → user_id` LRU cache that kills the warm-path BCrypt cost. It does **not** mitigate bad-key DoS — random bad keys never cache-hit, so they still walked every hash. This PR is the complementary half:

- cache → speeds **valid keys** on the warm path
- bucket → kills **bad keys** on the cold path

Together: fast for everyone, unattackable. Both edits live in `_authenticate_key()`; they layered cleanly since #4 merged first — no conflict.

## Graceful degrade — schema-v2 migration

If any record loaded from `agents-auth.json` lacks a `key_prefix` field (legacy entries pre-schema-2), one startup WARN names the affected `user_id`s and points at the provisioning script. Those records still authenticate correctly via the `_unprefixed_agents` fallback, at the old O(N) cost. Per-record degrade — not a process-wide one. No crashes; no operator scramble. Either PR (this MemOS PR or the Hermes script PR) can land first without breaking auth.

## Cross-link

- Hermes companion PR (provisioning script un-archive + chmod 600 + remove archived hashes): https://github.com/sergiocoding96/hermes-multi-agent/pull/15

## Test plan

- [x] All 6 startup-gate cases (auth-optional, auth-required×{valid, unset, missing, empty, malformed})
- [x] Bad-key bench: 3601ms → 0.00ms p50 (in-process, 15-agent live registry)
- [x] Valid-key cold bench: 238ms (1 BCrypt) — cache miss + bucket hit
- [x] Valid-key warm bench: 0.00ms — cache hit (fix/auth-perf already in main)
- [x] Rate-limit single-WARN: 1 emission across 201 `_get_redis()` calls
- [x] Rate-limit SQLite fallback: 100 allowed + 100 denied at `RATE_LIMIT=100`, persists across reopen
- [x] Module imports, AST parses

## Out of scope (deferred)

- **F-02** admin key non-constant-time compare → use `hmac.compare_digest`
- **F-03** no minimum-BCrypt-cost guard at load

Both real but not MVP-blocking.

### DIFF
--- a/PATCHES.md
+++ b/PATCHES.md
@@ -1,0 +1,79 @@
+# Security & Tuning Patches
+
+This fork maintains a set of patches on top of upstream `MemTensor/MemOS` for
+production deployment inside the Hermes multi-agent system.
+
+## Why a fork
+
+Upstream is a reference implementation. We need:
+- Per-agent API key auth with spoof detection and cube-level ACL
+- Hardened defaults (localhost binding, required passwords, secret encryption)
+- Search recall tuning (was returning 1/12 relevant results on vague queries)
+- Memory extraction quality fixes (English enforcement, granularity rule)
+
+See the full audit history at
+`memos-setup/learnings/` in the [hermes-multi-agent](https://github.com/sergiocoding96/hermes-multi-agent)
+repo.
+
+## Patched Files
+
+| File | Category | What changed |
+|------|----------|-------------|
+| `src/memos/api/middleware/agent_auth.py` | **NEW** | Per-agent Bearer-key auth middleware, bcrypt v2 format, auth-failure rate limit, auto-reload on config mtime change |
+| `src/memos/api/middleware/request_context.py` | Security | Strip `Authorization` + `Cookie` headers from logs |
+| `src/memos/api/server_api.py` | Security | Register `RateLimitMiddleware` + `AgentAuthMiddleware` + admin router; bind to `MEMOS_BIND_HOST` (default 127.0.0.1) |
+| `src/memos/api/routers/server_router.py` | Authorization | `_enforce_cube_access()` helper; auth on 11 previously-unprotected endpoints; scheduler caps (300s max timeout, cross-agent checks) |
+| `src/memos/api/routers/admin_router.py` | Admin API | Full rewrite for bcrypt v2 key management via `agents-auth.json`; list/create/revoke/rotate |
+| `src/memos/api/handlers/search_handler.py` | Security + tuning | Spoof check + cube isolation; env-configurable search params (`MOS_SEARCH_TOP_K_FACTOR`, `MOS_MMR_TEXT_THRESHOLD`, `MOS_MMR_PENALTY_THRESHOLD`) |
+| `src/memos/api/handlers/add_handler.py` | Security | Spoof check + cube isolation; empty content rejection; log request summary instead of full body |
+| `src/memos/api/handlers/component_init.py` | Security | Inject `UserManager` into `HandlerDependencies` for ACL checks |
+| `src/memos/api/product_models.py` | Tuning | Relativity default 0.20 → 0.05 (was filtering vague but valid queries) |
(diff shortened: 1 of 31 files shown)


sergiocoding96:
Closing — opened against the wrong base. This was meant for the sergiocoding96/MemOS fork, not the upstream. Will reopen there.

PR #1663, opened 2026-05-08
fix(memos-local-plugin): handle full endpoint paths in openai_compatible probe

liuhaibin0528:
## Problem

When configuring `openai_compatible` provider with an endpoint that already includes the full API path (e.g. Zhipu AI / GLM models):

```yaml
endpoint: https://open.bigmodel.cn/api/paas/v4/chat/completions
```

The probe function would append another `/v1/chat/completions` segment, resulting in a malformed URL:

```
https://open.bigmodel.cn/api/paas/v4/chat/completions/v1/chat/completions
```

This causes the model test to fail for any provider whose endpoint already specifies the full path.

## Solution

Check if the endpoint already ends with the target path (`/chat/completions` or `/embeddings`) before appending. The URL resolution now handles three cases:

1. Endpoint ends with `/chat/completions` → use as-is
2. Endpoint ends with `/v1` → append `/chat/completions`
3. Otherwise → append `/v1/chat/completions`

The same fix was applied symmetrically to both `probeChat` and `probeEmbedding`.

## Files Changed

- `apps/memos-local-plugin/server/routes/models.ts`

## Backward Compatibility

Fully backward compatible — existing endpoint formats (`/v1` or base URL) continue to work as before. This only adds support for the already-full-path case.


Memtensor-AI:

Cloud test-engine rerun against `dev-v2.0.22` completed successfully after adding focused scope mapping for plugin server route changes.

- Run: `tr-896e2a62-172` on cloud test-engine `10012`
- Result: 74 passed, 0 failed, 0 skipped

This replaces the earlier broad plugin full-suite failure, which was unrelated to this PR's `server/routes/models.ts` change. Manual code review is still required before merge.

PR #1667, opened 2026-05-09
fix(memos-local-plugin): harden hermes bridge lifecycle

liuhaibin0528:
## Summary

This hardens the Hermes stdio bridge lifecycle for `apps/memos-local-plugin`.

- terminate orphaned non-daemon `bridge.cts --agent=hermes` processes on client close
- recreate the bridge if the cached Python client points to a dead child process
- isolate `feedback.submit` failures so they do not abort `sync_turn`
- use a longer timeout for `turn.end` and a shorter timeout for `feedback.submit`
- add Python regression coverage for the new bridge lifecycle behaviors

## Why

In long-running Hermes sessions I hit two stability problems repeatedly:

1. stale stdio bridge children accumulated after reconnects/timeouts
2. a slow or failed `feedback.submit` could make the whole turn sync look failed even though `turn.end` had already completed

This patch keeps the viewer daemon behavior intact while making the short-lived stdio bridge process lifecycle explicit.

## Testing

- `python3 -m unittest apps.memos-local-plugin.tests.python.test_bridge_client`
- `python3 -m unittest apps.memos-local-plugin.tests.python.test_hermes_provider_pipeline`

### DIFF
--- a/apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py
+++ b/apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py
@@ -830,8 +830,11 @@ class MemTensorProvider(MemoryProvider):
                 agent_thinking=thinking,
             )
             if _is_verifier_feedback_prompt(user):
-                self._submit_verifier_feedback(user, assistant, ts_ms)
-                feedback_submitted = True
+                try:
+                    self._submit_verifier_feedback(user, assistant, ts_ms)
+                    feedback_submitted = True
+                except Exception as err:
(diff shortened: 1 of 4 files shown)


shinetata:
Thanks — all four hardening items here have since landed on `main` via separate changes:

- orphan termination on close → escalating SIGTERM/SIGKILL in `close()`
- recreate-on-dead-child → the `poll()` guard in `request()` (#2028)
- `feedback.submit` failure isolation → the `feedback_submitted` flag in the sync path
- per-method timeouts → `turn.end` on the long timeout, `feedback.submit` at 75s

Closing as already-covered; happy to look again if you spot a gap on latest `main`.

PR #1670, opened 2026-05-09
fix(fetchers): honor Retry-After header on 429/503 responses

### LINKED ISSUE #1620: 429 retry handling ignores Retry-After header in LLM and embedding fetchers
The LLM and embedding fetchers classify HTTP 429 as transient and retry with exponential backoff — the same path used for 5xx. Retry-After header is never read.

## Evidence

apps/memos-local-plugin/core/llm/fetcher.ts

Line 58:
```ts
const transient = resp.status >= 500 || resp.status === 429;
```

backoff() function line 236:
```ts
const ms = base * 2 ** (attempt - 1) + jitter;
```

No Retry-After header read. Same pattern in apps/memos-local-plugin/core/embedding/fetcher.ts line 48.

## Consequence

Under provider rate limiting, retries fire before the upstream-requested cooldown expires. Premature retries extend rate limiting, increase failed requests, and waste paid API calls.

## Suggested fix

In both fetchers, read Retry-After before computing backoff delay. Support integer seconds and HTTP-date formats. Use header value when present, fall back to existing exponential backoff when absent.

## Related pattern
retry-after-ignored-under-concurrency
Corpus reference: https://github.com/SirBrenton/pitstop-truth

cc @CaralHsi @Ki-Seki

Sanjays2402:
## Summary

The LLM and embedding fetchers in `apps/memos-local-plugin` classified HTTP 429 as transient and retried with exponential backoff, but never read the `Retry-After` header. Under provider rate limiting, retries fired before the upstream-requested cooldown, extending throttling and wasting paid API calls.

This PR parses `Retry-After` per RFC 7231 §7.1.3 (delta-seconds or HTTP-date), caps the value to a sane maximum, and prefers it over exponential backoff on 429/503. Falls back to the existing backoff when the header is absent or unparseable.

## Changes

- New shared helper `apps/memos-local-plugin/core/util/retry-after.ts` exporting `parseRetryAfterMs(resp, capMs)`.
- `apps/memos-local-plugin/core/llm/fetcher.ts`: parse Retry-After on 429/503 and use it as the sleep duration; small refactor split `backoff()` into `backoffMs()` + `sleep()` so both code paths share one timer.
- `apps/memos-local-plugin/core/embedding/fetcher.ts`: same treatment, mirroring the LLM fetcher.
- Tests added/extended:
  - `tests/unit/util/retry-after.test.ts` — covers integer seconds, HTTP-date, capping, malformed, missing, past-date, and negative inputs.
  - `tests/unit/llm/fetcher.test.ts` and `tests/unit/embedding/fetcher.test.ts` — three new cases each: integer Retry-After is honored, HTTP-date is honored, malformed value falls back to backoff.

## Behavior

- `Retry-After: 5` → sleep ~5s instead of next backoff step.
- `Retry-After: Wed, 21 Oct 2025 07:28:00 GMT` → sleep ~(date - now), capped.
- Header absent or malformed → falls back to existing exponential backoff (no behavior change).
- Capped at 60s (`MAX_RETRY_AFTER_MS`) to prevent hostile or buggy servers from pinning the client indefinitely.
- Only 429 and 503 honor Retry-After. Other 5xx still use exponential backoff (the header is rarely set on those, and we want to keep retrying quickly on non-rate-limit failures).

## Notes

- Could not run `tsc --noEmit` locally (no `node_modules` in this worktree), so types were verified by inspection. The new helper has explicit return type `number | null` and matches the import-style (`.js` extension, `Bundler` resolution) used elsewhere in the codebase.
- Tests use the existing real-timer pattern in `tests/unit/embedding/fetcher.test.ts`; sleep upper bounds are generous (≥900 ms for `Retry-After: 1`) to avoid flakes on slow CI.

Fixes #1620


Sanjays2402:
Thanks for the review — addressed in the latest commit:

- **`retry-after.ts:11`**: Updated the JSDoc to clarify that `0` (retry immediately) is a valid value; only negative or unparseable values return `null`. Tests already covered the `0` case and continue to pass.
- **`llm/fetcher.ts:17` / `embedding/fetcher.ts:15`**: Applied `Math.min(sleepMs, MAX_RETRY_AFTER_MS)` to the final sleep duration in both fetchers, so the 60s cap applies whether the value came from `Retry-After` or `backoffMs()`. Previously a `maxRetries: 10` config could let exponential backoff exceed 60s.

Diff is small and behavior-only-on-the-margin (only changes anything when `attempt >= ~6` in a long retry chain, or when `Retry-After: 0` arrives).

Memtensor-AI:

- Run: `tr-f2a7758b-879` on cloud test-engine `10010`
- Verdict: `fail_code_bug`
- Summary: Tests failed. Failed cases: test -- tests/unit/embedding/fetcher.test.ts tests/unit/llm/fetcher.test.ts tests/unit/util/retry-after.test.ts

- memos_local_plugin/unit: 29 passed, 2 failed, 0 skipped

Failed cases:
- test -- tests/unit/embedding/fetcher.test.ts tests/unit/llm/fetcher.test.ts tests/unit/util/retry-after.test.ts: t.ts > llm/fetcher > 429 with Retry-After: HTTP-date in the near future is honored
AssertionError: expected 650 to be greater than or equal to 800
 ❯ tests/unit/llm/fetcher.test.ts:123:21
    121|     expect(f).toHaveBeenCalledTimes(2);
    122|     // Roughly the diff (allow generous slack for clock + jitter).
    123|     expect(elapsed).toBeGreaterThanOrEqual(800);
       |                     ^
    124|     expect(elapsed).toBeLessThan(5_000);
    125|   });

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[2/2]⎯

PR #1674, opened 2026-05-09
fix(metrics): return real total memory count instead of capped page size

### LINKED ISSUE #1593: Web UI memory count stuck at 500, actual traces exceed 1400
## Bug Description

**Environment:**
- MemOS version: v2 (latest)
- Deployment: hermes-agent plugin mode
- Node.js process running normally

**Problem:**
Web UI shows Memory count: 500 but database has 1400+ traces. The number stays fixed at 500 regardless of actual trace count growth.

**Expected behavior:**
Memory count should reflect actual trace count in database (~1422), not a fixed cap of 500.

**Steps to reproduce:**
1. Use hermes-agent with MemOS normally
2. Observe Web UI number after extended use
3. Number stops at 500 while database trace_count exceeds 1400

**Initial analysis:**
Suspect issue in metrics().total calculation logic, or frontend applying a hard cap on totalCount. Database query for trace count is accurate - problem is in presentation layer.

**Impact:**
Cannot accurately monitor memory system storage via Web UI.

Sanjays2402:
## Summary

The Web UI memory-count card on the Memories page was stuck at 500 even when the database held 1400+ traces (#1593). The metrics path counted items from an already-limited listing instead of running a dedicated COUNT query.

## Root cause

`countTraces` in `apps/memos-local-plugin/core/pipeline/memory-core.ts` computed the total by calling `handle.repos.traces.list({ limit: 100_000 })` and reading `rows.length`. The shared helper `buildPageClauses` / `clampLimit` (in `core/storage/repos/_helpers.ts`) silently clamps every list `limit` to **500**, so `list({ limit: 100_000 })` actually returns at most 500 rows. The resulting `rows.length` (or grouped `turnKeys.size`) was therefore capped at 500 regardless of how many traces existed.

The `/api/v1/traces` endpoint uses `countTraces`, and the Memories view (`web/src/views/MemoriesView.tsx`) renders that response's `total` in its pagination footer, so the Web UI showed 500 forever.

## Fix

In `countTraces`:

- For the no-search path, use the repo's dedicated `SELECT COUNT(*)` queries directly (`repos.traces.count` / `countTurns`). These don't go through `buildPageClauses` and have no 500-row cap, so the returned total reflects the real database size.
- For the substring-search path, page through traces explicitly with offset increments instead of relying on a single `list()` call (which would also be clamped at 500). Each batch is filtered for visibility and matched against the search needle, so the count stays accurate even above 500 rows.

Listings keep their existing page-size limits; only the count was wrong.

## Verification

Added a regression test in `tests/unit/pipeline/memory-core.test.ts` that inserts 600 trace rows across 6 distinct turns and asserts:

- `countTraces({})` returns `600` (was 500 before the fix)
- `countTraces({ groupByTurn: true })` returns `6` (was capped at the smaller of 6 or 500 before; would break for >500 turns)

Files changed:
- `apps/memos-local-plugin/core/pipeline/memory-core.ts` (~34 net lines)
- `apps/memos-local-plugin/tests/unit/pipeline/memory-core.test.ts` (regression test)

Fixes #1593

### DIFF
--- a/apps/memos-local-plugin/core/pipeline/memory-core.ts
+++ b/apps/memos-local-plugin/core/pipeline/memory-core.ts
@@ -2630,22 +2630,42 @@ export function createMemoryCore(
     const needle = (input?.q ?? "").trim().toLowerCase();
     const visible = (r: TraceRow) => visibleToCurrent(r);
     if (!needle) {
-      const rows = handle.repos.traces.list({ sessionId: input?.sessionId, limit: 100_000 }).filter(visible);
-      if (!input?.groupByTurn) return rows.length;
-      const turnKeys = new Set<string>();
-      for (const r of rows) turnKeys.add(`${r.episodeId ?? "_"}:${r.turnId}`);
-      return turnKeys.size;
+      // Use a dedicated SELECT COUNT(*) instead of `list().length`.
+      // The previous implementation called `repos.traces.list({ limit: 100_000 })`
+      // and counted the returned rows, but `buildPageClauses` silently clamps
+      // every list `limit` to 500. That capped the Memories total at 500
+      // even when the database held 1400+ traces (#1593). The repo's
+      // `count` / `countTurns` issue real COUNT queries with no page-size
+      // cap, so they return the actual total.
+      return input?.groupByTurn
+        ? handle.repos.traces.countTurns({ sessionId: input?.sessionId })
+        : handle.repos.traces.count({ sessionId: input?.sessionId });
     }
     // q substring scan — mirror `listTraces`. Walk all matching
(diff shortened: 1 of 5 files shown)


Sanjays2402:
Real catch — addressed in the latest commit. The new COUNT path was bypassing the visibility predicate that `list()` applies, so totals could include rows from other profiles/namespaces and break pagination math.

Fix: shared `buildVisibilityClause` helper used by both `list()` and the count methods, so the predicate stays authoritative in one place. `countTraces` in `memory-core.ts` now threads the relevant options through. Added a regression test that inserts traces under two profiles and asserts each profile's `countTraces` returns only its own count.

PR #1676, opened 2026-05-10
fix(memos-local-plugin): close all pipe fds in bridge_client to prevent process leaks

rayraiser:
## Problem

The `close()` method in `bridge_client.py` only closed `stdin`, leaving `stdout` and `stderr` pipe file descriptors open. Because the bridge process (node tsx) only exits when stdin reaches EOF **and** there are no open pipe references, this caused bridge processes to accumulate as orphans (e.g., 78+ processes per session).

## Root Cause

- `close()` called `self._proc.stdin.close()` but never closed `stdout`/`stderr`
- The Node bridge process stayed alive, holding the pipe open
- New sessions spawned new bridge processes, each leaking

## Fix

Close **all** pipe fds (`stdin`, `stdout`, `stderr`) in `close()`, then `wait()` up to 2s for graceful exit before `terminate()`.

## Impact

- Bridge processes now exit cleanly on session end
- No orphan process accumulation
- Viewer will restart on new session (acceptable trade-off)

## Files Changed

- `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py`

### DIFF
--- a/apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py
+++ b/apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py
@@ -280,6 +280,7 @@ class MemTensorProvider(MemoryProvider):
 
     def is_available(self) -> bool:  # type: ignore[override]
         try:
+            from daemon_manager import ensure_bridge_running
             return ensure_bridge_running(probe_only=True)
         except Exception:
             return False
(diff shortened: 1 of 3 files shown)


shinetata:
Thanks! On current `main` the premise no longer holds: `close()` no longer relies on stdin/stdout/stderr reaching EOF for the child to exit — it escalates to SIGTERM and then SIGKILL, and the #1910 singleton registry reaps the previous client. The fd-leak-driven orphan accumulation this fixes is therefore already prevented.

Closing as superseded; reopen if you can still reproduce on latest `main`.

Issue #1452, opened 2026-04-09
auto-recall can block gateway startup / first-turn path long enough to fail health checks


starm0on:
﻿## Symptom | 症状
With `auto-recall` enabled and existing memories present, gateway startup or the first user turn can block long enough to trigger health-check failures and restart loops.

在 `auto-recall` 开启且已有记忆数据时，gateway 启动阶段或首轮用户消息阶段可能被阻塞得足够久，最终触发健康检查失败甚至重启链。

## Trigger | 触发条件
- `@memtensor/memos-local-openclaw-plugin@1.0.8`
- Linux + `systemctl --user`
- existing memory records already in database
- `allowPromptInjection=true` and `auto-recall enabled`
- recall filtering path uses a slow/timeout-prone model

## Minimal Reproduction | 最小复现
1. Enable `auto-recall`.
2. Have enough existing memories in the local database.
3. Use a slow model on the recall/filter path.
4. Restart gateway or send the first message after startup.
5. On affected runs, recall/filter work blocks the critical path for tens of seconds.

1. 开启 `auto-recall`。
2. 本地数据库里已有一定数量的记忆。
3. 让 recall/filter 链路走一个慢模型。
4. 重启 gateway，或在启动后发送第一条消息。
5. 有问题的运行里，这条 recall/filter 工作会把关键路径阻塞几十秒。

## Actual Result | 实际结果
In the affected environment, this could stall the startup / first-turn path for ~30-40s. That was enough to trip health checks and contribute to restart loops.

在实际环境里，这条链路会把启动或首轮路径卡住约 30-40 秒，足以打爆健康检查，并参与重启循环。

## Local Workaround | 本地临时解决办法
The local mitigation was:
- add hard timeout around recall/filter LLM work
- fail open on timeout/errors
- do not let recall/filter exceptions propagate to the gateway top level
- keep startup `ready` independent from slow recall work

本地止血方式是：
- 给 recall/filter 的 LLM 工作加硬超时
- 超时或报错后 fail-open
- 不让 recall/filter 异常再抛到 gateway 顶层
- 保持 startup `ready` 不依赖慢召回链路

## Suggested Fix | 建议修复方向
Please move auto-recall filtering out of the startup-critical / first-turn-critical path, and enforce timeout + fail-open semantics for slow recall models. Recall enrichment should be optional context, not something that can delay readiness or destabilize the gateway.

[...]

PR #1765, opened 2026-05-19
fix(bridge): prevent process leak via PID file singleton guard

syzsunshine219:
## Summary

- Bridge processes accumulate indefinitely: each `hermes chat` session spawns a new bridge.cts, and the previous one stays alive as a "daemon" but is never cleaned up. On long-running servers this results in 10+ zombie bridge pairs.
- Introduces a PID file (`daemon/bridge.pid`) as a lightweight singleton lock. On startup, the new bridge reads the PID file, kills the stale process (SIGTERM → 5s → SIGKILL), writes its own PID, and removes it on all exit paths.
- `--no-viewer` (headless) bridges skip the kill — they coexist with the daemon that owns the viewer port.
- The existing `install.sh` kill logic is preserved as a complementary deployment-time fallback.

## Test plan

- [ ] Start bridge daemon → verify `daemon/bridge.pid` is created with correct PID
- [ ] Start a second bridge → verify the first is killed and PID file updated
- [ ] SIGTERM the running bridge → verify PID file is removed
- [ ] Run `hermes chat` twice in sequence → verify only 1 bridge pair remains (no accumulation)
- [ ] Run `--no-viewer` bridge alongside daemon → verify both coexist (no kill)

### DIFF
--- a/apps/memos-local-plugin/bridge.cts
+++ b/apps/memos-local-plugin/bridge.cts
@@ -63,8 +63,89 @@ function parseArgs(argv: readonly string[]): BridgeArgs {
   return args;
 }
 
+// ─── PID file singleton guard ───────────────────────────────────────────
+// Prevents bridge process accumulation: each new bridge that wants to
+// own the viewer port kills the previous holder via its PID file.
+// `--no-viewer` (headless) bridges skip the kill — they don't need the
+// port and should coexist with the daemon that owns it.
+
+const PID_FILENAME = "bridge.pid";
+
+function pidFilePath(agent: string): string {
+  const agentHome = agent === "hermes" ? ".hermes" : ".openclaw";
+  return path.join(
+    process.env.HOME ?? "/tmp",
+    agentHome,
+    "memos-plugin",
+    "daemon",
+    PID_FILENAME,
+  );
+}
+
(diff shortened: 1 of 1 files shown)


chiefmojo:
**Large-DB crash loop: PID singleton kills old bridge before DB lock releases**

We reverted a 655MB MemOS DB from v2.0.0-beta.1 to v2.0.5 and hit an infinite bridge restart cycle. The PID singleton guard kills the old bridge, but the new bridge then opens the same SQLite DB and hangs in WAL recovery — long enough to exceed the gateway's 30s timeout. The gateway spawns yet another bridge, which kills the current one, and the cycle repeats indefinitely.

On a smaller 43MB DB the same code works fine — WAL recovery completes before the next bridge spawns. The threshold where it breaks seems to be somewhere between those two, likely dependent on DB size and I/O speed.

**Root cause:** `killExistingBridge()` sends SIGTERM but doesn't wait for the old process to fully exit and release file locks. A `waitpid()` or polling `fuser` on the DB file before proceeding would prevent the race.

Cross-ref: [#1452](https://github.com/MemTensor/MemOS/issues/1452) describes a related class of startup-hang → restart-loop issues.

PR #1758, opened 2026-05-18
fix: make MemOS startup recovery non-blocking

leavrcn:
## Summary
- Make MemOS startup recovery for stale open / dirty closed episodes non-blocking during `MemoryCore.init()`.
- Schedule startup recovery work in the background so `session.open` can return promptly even when recovery triggers slow reward/LLM work.
- Add a focused regression test that locks the init-time non-blocking contract.

## Why
Hermes adapter opens a JSON-RPC bridge session during startup. If `init()` synchronously waits for stale/dirty episode recovery, recovery can call slow scoring paths and prevent `session.open` from responding before the adapter timeout. The host then kills/restarts bridge processes and can enter a reconnect storm.

## Test plan
- `npm --prefix apps/memos-local-plugin test -- tests/unit/startup-recovery.test.ts`
- `cd apps/memos-local-plugin && npm run lint`

PR #1784, opened 2026-05-22
fix: reward pipeline skips abandoned episodes (3 related fixes)

### LINKED ISSUE #1782: Reward pipeline skips abandoned episodes — 98% of closed episodes never scored
## Problem

The reward scoring pipeline processes only ~7 episodes on bootstrap and then goes permanently idle, leaving the vast majority of traces unscored. On a 43 MB database with 3,600 traces, only 45 (1.3%) had r_human scores before fixing.

## Root Cause

Three interacting bugs in `memory-core.ts`:

### 1. `episodeRewardIsDirty()` excludes abandoned episodes (line ~1274)

The dirty-check condition only matches episodes with `closeReason === "finalized"` or `recoveryReason === "missed_session_end"`. But **219 of 224 closed episodes had `closeReason: "abandoned"`**, so they silently failed the dirty check on every bootstrap scan.

### 2. No polling fallback after bootstrap

`autoRescoreDirtyClosedEpisodes()` is called once during `init()` and never again. After bootstrap completes, the daemon bridge sits permanently idle with no mechanism to retry missed episodes.

### 3. Two-bridge isolation (contributing factor)

The viewer daemon bridge (`--daemon`) and JSON-RPC bridge (`--no-viewer`) are separate processes with separate in-memory event buses. New captures flow through the JSON-RPC bridge's pipeline; the daemon's reward subscriber never sees those events. The daemon only activates its own pipeline during `init()`.

### Why exactly 7 episodes got scored?

Those 7 were open episodes with traces at init-time, processed by `recoverOpenEpisodesAsSessionEnd()` as "stale." All 219 closed abandoned episodes were silently excluded.

### 1. Add `closeReason === "abandoned"` to `episodeRewardIsDirty()`

```diff
- (meta.closeReason === "finalized" || meta.recoveryReason === "missed_session_end")
+ (meta.closeReason === "finalized" ||
+   meta.closeReason === "abandoned" ||
+   meta.recoveryReason === "missed_session_end")
```

### 2. Add periodic rescore timer in `init()`

```ts
const rescoreInterval = setInterval(() => {
  void autoRescoreDirtyClosedEpisodes().catch((err) => {
    log.debug("periodic_rescore.error", {
      err: err instanceof Error ? err.message : String(err),
    });
  });
}, 10 * 60 * 1000);
(rescoreInterval as unknown as { unref?: () => void }).unref?.();
```

This covers episodes that miss the startup scan (closed after init, or retry of failed reward runs). The 10-minute interval is safe because `autoRescoreDirtyClosedEpisodes` has its own 30-second dedup guard.

## Results After Fix

Before (3,600 traces):
- r_human scored: **45** (1.3%)
- Large episodes completely skipped: ep_0f9jsh492n40 (1,367 traces), ep_x95apvvw7gdx (444 traces), ep_kt2ds1afhssq (222 traces)

Within 3 minutes of restart:
- r_human: 45 -> **257** (212 freshly scored)
- ep_kt2ds1afhssq (222 traces): scored with rHuman=1.0 in **1.1 seconds**
- Estimated full backlog: 60-90 minutes

## Environment

- **Version:** v2.0.5
- **DB:** 43 MB, WAL mode
- **LLM:** qwen/qwen3-235b-a22b-2507 via OpenRouter
- **Config:** lightweightMemory.enabled: false

chiefmojo:
## Problem

The reward scoring pipeline processes only ~7 episodes on bootstrap and then goes permanently idle, leaving the vast majority of traces unscored. On a 43 MB database with 3,600 traces, only 45 (1.3%) had r_human scores before fixing.

## Root Cause & Fixes

Three interacting bugs, all in `apps/memos-local-plugin/`:

### Fix 1: `episodeRewardIsDirty()` excluded abandoned episodes
The dirty-check condition only matched `closeReason === "finalized"` or `recoveryReason === "missed_session_end"`. 219 of 224 closed episodes had `closeReason: "abandoned"`.

**Fix:** Add `closeReason === "abandoned"` to the rescore condition + a 10-minute periodic rescore timer.

### Fix 2: Daemon HTTP server must bind before `core.init()`
The daemon bridge started `core.init()` before `startHttpServer()`. When init rescores dirty episodes (5+ minutes of LLM calls), port 18800 stays free. The Python watchdog times out its 15-second health probe and spawns a new daemon — killing the in-progress one.

**Fix:** In daemon mode, bind HTTP server first, then run init asynchronously.

### Fix 3: `reward.skipped` blocked abandoned episodes from recovery
`episodeRewardIsDirty()` checked `reward.skipped === true` BEFORE `closeReason === "abandoned"`. The reward runner correctly skipped 173 one-turn episodes in a previous session, but that flag permanently excluded them from recovery.

**Fix:** `reward.skipped` is only honored if the episode is NOT (abandoned + no prior recovery). Abandoned episodes get one pass; after that, closeReason is patched to "finalized" and the normal skip guard works.

## Verification
- Daemon bridge: 7+ hours uptime, zero restart loops
- All 219 abandoned episodes processed
- r_human scores flowing for traces from May 9 onward
- Closes #1782

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,7 +1,7 @@
 name: "\U0001F41B Bug Report"
 description: Report a bug to help us improve MemOS | 报告错误以帮助我们改进 MemOS
 title: "fix: "
-labels: ["bug", "pending"]
+labels: ["bug"]
 body:
 
   - type: checkboxes
--- a/.github/ISSUE_TEMPLATE/feature-request.yml
+++ b/.github/ISSUE_TEMPLATE/feature-request.yml
@@ -1,7 +1,7 @@
 name: "\U0001F680 Feature request"
 description: Submit a request for a new feature | 申请添加新功能
 title: "feat: "
-labels: ["enhancement", "pending"]
+labels: ["enhancement"]
 body:
 
   - type: checkboxes
--- a/README.md
+++ b/README.md
@@ -19,10 +19,10 @@
 </p>
 
 <p align="center">
-  <strong>🎯 +43.70% Accuracy vs. OpenAI Memory</strong><br/>
-  <strong>🏆 Top-tier Long-term Memory + Personalization</strong><br/>
-  <strong>💰 Saves 35.24% Memory Tokens</strong><br/>
-  <sub>LoCoMo 75.80 • LongMemEval +40.43% • PrefEval-10 +2568% • PersonaMem +40.75%</sub>
+  <strong>🏆 Leading Performance Across Agent and User Memory Benchmarks</strong><br/>
+  <strong>🤖 OpenClaw Task Completion Improves from 36.63% to 50.87%</strong><br/>
+  <strong>🎯 92.34 on LoCoMo and 93.40 on LongMemEval</strong><br/>
+  <strong>📊 Unified Evaluation Across 14 Commercial Memory Products</strong>
   <!-- <a href="https://memos.openmem.net/">
     <img src="https://statics.memtensor.com.cn/memos/github_api_free_banner.gif" alt="MemOS Free API Banner">
   </a> -->
@@ -70,6 +70,9 @@ Your lobsters and Hermes Agents now have **the best** memory system — choose *
 
 
 ### News
+
+- **2026-07-02** · 🏆 **MemOS Advances Agent and User Memory Benchmarks**
+  With MemOS, **OpenClaw** improves average task completion from **36.63% to 50.87%** across five agent tasks. MemOS also achieves **92.34 on LoCoMo** and **93.40 on LongMemEval**, and leads in **OmniMemEval**, a unified evaluation of 14 commercial memory products across ten datasets.
 
 - **2026-05-09** · 🧠 **memos-local-plugin 2.0**
   Official local memory plugin for **Hermes Agent** and **OpenClaw**. One core powers self-evolving memory across L1 traces, L2 policies, L3 world models, and crystallized Skills, with local-first storage and feedback-driven retrieval.
@@ -273,14 +276,8 @@ MemOS is a **Memory Operating System** for LLMs and AI agents that unifies store
 
 ### What are the benchmark results?
 
-| Benchmark | MemOS Result | Improvement |
-|-----------|--------------|-------------|
-| LoCoMo | 75.80 | - |
-| LongMemEval | +40.43% vs baseline | - |
-| PrefEval-10 | +2568% | - |
-| PersonaMem | +40.75% | - |
-| **vs OpenAI Memory** | +43.70% Accuracy | - |
-| **Token Savings** | 35.24% | - |
+MemOS achieves **92.34 on LoCoMo** and **93.40 on LongMemEval** for User Memory, while improving **OpenClaw** average task completion from **36.63% to 50.87%** across five Agent Memory tasks. For details, see [OmniMemEval](https://github.com/MemTensor/OmniMemEval), our unified evaluation framework for benchmarking 14 commercial memory products across ten datasets.
+
 
 ### How does MemOS compare to other memory solutions?
 
--- a/apps/memos-local-openclaw/README.md
+++ b/apps/memos-local-openclaw/README.md
@@ -514,12 +514,14 @@ All optional — shown with defaults:
     "recall": {
       "maxResultsDefault": 6,     // Default search results
       "maxResultsMax": 20,        // Max search results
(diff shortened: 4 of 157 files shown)


chiefmojo:
Added 4 additional fixes discovered during validation:

1. **`510cabdb`** — Migration 013: repair broken `traces_fts` triggers. A previous FTS rebuild used incorrect UPDATE trigger syntax that silently failed on every `updateScore()` call. Scores were computed via ONNX but never persisted to the DB.

2. **`90f60f68`** — Clear `rewardDirty` flag on skip path. Prevents skipped episodes from being re-scanned on every bridge restart.

3. **`f1098bd0`** — Set `r_task=0` on skipped episodes. Short-circuit episodes with no qualifying turns so the dirty scan doesn't process them in a loop forever.

4. **`027512f9`** — Paginate closed-episode dirty scan. `clampLimit()` caps all `list()` calls at 500, so episodes beyond rank 500 were permanently invisible to the scan. Replaced with a `collectDirtyClosedEpisodes()` helper that paginates through all closed episodes.

Verified across 3 hosts with DBs ranging from 462 to 30K traces.

shinetata:
Thanks for the very thorough investigation here, and for filing #1782 with a clean reproduction — the root-cause write-up was genuinely helpful.

**Status: closing this as largely superseded**, with one piece that is worth resubmitting on its own (details at the end).

### Headline fix already landed
The core of this PR — adding `closeReason === "abandoned"` to `episodeRewardIsDirty()` plus the 10-minute periodic rescore timer — was merged to `main` via #1872 (2026-06-04), which closed #1782. `main` has since evolved further on top of it (#1808 backoff, #1966 / #1787 trace-count / newer-than checks, #2063 lightweight no-op), so this branch now conflicts.

### Re-evaluating the remaining commits against current `main`

- **Daemon: bind HTTP before `core.init()` (`bridge.cts`).** `core.init()` is now non-blocking (#1758 / #2002): it returns after a synchronous classification pass and runs recovery in the background, so the "init blocks for minutes → 15s health probe times out → replacement daemon → restart loop" premise no longer holds. It also collides head-on with the #1747 startup-ordering invariant and its pinning test.

- **`reward.skipped` one-pass for abandoned + the `reward.ts` skip-path changes (`setRTask(0)` / `traceCount` / clearing `rewardDirty`).** On `main`, `episodeRewardIsDirty()` short-circuits on `reward.skipped === true` *before* the `rTask == null` / abandoned branch, so the "skipped episodes loop forever" scenario does not occur, and abandoned episodes are already scored via #1872. These no longer apply.

- **Migration 013 (traces_fts triggers).** This repairs damage attributed to commit `a054c9b8`, which is not in this repository's history. Upstream `001-initial.sql` already defines `traces_fts` with the canonical `trace_id` column and correct direct-DELETE triggers, matching the JOIN in `repos/traces.ts`, and no migration between 002–012 rebuilds it. Running 013 upstream would drop and rebuild the FTS on every user's DB to fix a bug they never had, so it should not be merged here.

### The one piece worth keeping: paginated dirty-closed scan
`collectDirtyClosedEpisodes()` addresses a real gap: `main` still scans with `list({ status: "closed", limit: 500 })`, and `clampLimit()` hard-caps at 500, so closed episodes past rank 500 (newest-first) are invisible to both the startup scan and the periodic rescore. On large DBs those episodes never get scored — consistent with the "dirty-17 at ranks 536–924" you observed.

If you would like to resubmit **just this** as a small, focused PR against current `main`, a few things need to be addressed for it to be mergeable:

1. **Bound the scan cost (blocking).** `episodeRewardIsDirty()` is heavier on `main` now — for #1966 / #1787 it issues a per-episode `countExisting` (`SELECT COUNT`) / `hasAnyNewerThan` (`SELECT 1`). The current 500 cap incidentally bounds each pass to ~500 of those; removing it makes every startup and every 10-minute tick walk *all* closed episodes and issue up to O(N) queries even when nothing is dirty. Please add a compensating bound (e.g. a `count()` short-circuit, a page/row ceiling, or an "already-fully-scored" watermark) and include a measurement on a large DB.

2. **Pagination stability.** OFFSET paging over `started_at DESC` while new closed episodes are inserted at the top can shift rows across page boundaries (and `started_at` is not unique), risking a *skipped* row — which would silently reintroduce the exact blind spot this fixes. Keyset / seek pagination on a stable key (`id` or `(started_at, id)`) would be more robust than OFFSET.

3. **Add a boundary test.** The behavior being fixed is a scale / boundary bug, so it should be pinned by a test that seeds >500 closed dirty episodes and asserts the one past rank 500 is collected and scored. `tests/unit/storage/traces-count.test.ts` already demonstrates the pattern for seeding >500 rows.

Thanks again — the analysis here was high quality, and a tightly-scoped pagination PR (with a cost bound + a >500 test) would be easy to review and land.

PR #1789, opened 2026-05-23
fix(memos-local-plugin): two-phase migration to prevent crash-loop on large databases

### LINKED ISSUE #1787: v2.0.2+ regression: bootstrapMemoryCoreFull() hangs with 100% CPU on databases >500MB
## Problem

All tagged plugin versions from v2.0.2 through v2.0.5 hang indefinitely during `bootstrapMemoryCoreFull()` when the SQLite database exceeds ~500MB. The bridge process burns 80-157% CPU (userspace, WCHAN=0) and never reaches `migration.applied` or `pipeline.ready`.

## Reproduction

1. Run any bridge from v2.0.2, v2.0.4, or v2.0.5 against a `memos.db` >500MB
2. Observe: `sqlite.open` appears in logs, then silence
3. CPU stays at 80-157% with no further log output
4. Bridge never reaches `migration.applied`

## Bisect Results

| Version | Result | Notes |
|---------|--------|-------|
| v2.0.5 | ❌ Hang | 687MB DB, 80-157% CPU |
| v2.0.4 | ❌ Hang | Same symptom |
| v2.0.2 | ❌ Hang | Same symptom |
| v2.0.0-beta.5 | ⚠️ Boots | Boots fast but heavy orphan processing |
| v2.0.0-beta.1 | ✅ Clean | 18ms boot, all subsystems initialize |

**Regression window:** Between v2.0.0-beta.5 and v2.0.2.

## Environment

- DB: 687MB, WAL mode, ~3,500 traces
- The host gateway kills and respawns the bridge every ~30s because it never becomes healthy
- Smaller databases (tested: 43MB, 56MB) do NOT trigger the hang

## What This Is NOT

- NOT the two-bridge SQLite lock race (that produces `database is locked` errors; this produces silence + CPU)
- NOT the abandoned-episode scoring gap (#1782) — those fixes address the scoring pipeline, not bootstrap

## Suspected Cause

`bootstrapMemoryCoreFull()` runs synchronously during bridge init. Something inside it scales poorly with database size — candidate: full table scans during migration checks, unbuffered trace iteration, embedding pipeline loading all traces, or orphan detection walking the entire table.

## Current Workaround

Roll back to v2.0.0-beta.1. It boots cleanly on the same 687MB database.

## Related

- #1782 — scoring fixes for abandoned episodes (separate concern)

chiefmojo:
## Problem

Migration 007 (namespace-visibility) in v2.0.2–v2.0.5 runs `UPDATE … SET share_scope='private'` plus `CREATE INDEX` on the traces table inside a single `db.tx()`. On databases larger than ~500MB, this exceeds the host gateway's kill timeout. SQLite rolls back the entire transaction — including the `schema_migrations` INSERT — so migration 007 is never recorded and the bridge restarts into the same hang forever.

Small databases (~43MB) complete within the timeout, which is why this only manifests on larger installs. Tested against a 687MB database with ~98,000 traces.

## Fix

Two-phase migration:

- **Phase 1** (inside transaction, milliseconds): `ADD COLUMN` only on all 12 namespace tables, plus `DROP INDEX uq_skills_name`. The `schema_migrations` record for v7 commits here.
- **Phase 2** (after migration loop, outside any transaction): Batched `UPDATE` in 2,000-row chunks (each its own implicit transaction) for `share_scope` backfill, then `CREATE INDEX IF NOT EXISTS` for all 18 owner/share indexes. `ensureNamespaceColumns` is called unconditionally on every boot so new tables in the namespace list get their columns.

**Restart safety:** If the bridge is killed during Phase 2, the v7 `schema_migrations` record survives (Phase 1 committed). Next boot skips Phase 1 entirely and resumes Phase 2 where it left off. The crash-loop is broken.

## Verification

- Confirmed the DB on the affected instance had exactly migrations 1–6 applied — migration 007 was never committed despite hundreds of boot attempts.
- After the fix, Phase 1 committed in 2.4 seconds. Phase 2 backfill and index creation complete across bridge restarts.
- Bridge boots cleanly with all 18 indexes created and `pipeline.ready`.

## Related

Closes #1787


Memtensor-AI:

- Run: `tr-2cf1d73f-bc0` on cloud test-engine `10011`
- Verdict: `fail_code_bug`
- Summary: Tests failed. Failed cases: test -- tests/unit/storage/migrator.test.ts tests/unit/storage/traces-count.test.ts

- memos_local_plugin/unit: 7 passed, 1 failed, 0 skipped

Failed cases:
- test -- tests/unit/storage/migrator.test.ts tests/unit/storage/traces-count.test.ts: should return accurate count > 500
AssertionError: expected 500 to be 50 // Object.is equality

- Expected
+ Received

- 50
+ 500

 ❯ tests/unit/storage/traces-count.test.ts:76:27
     74|     // Verify list with no limit keeps the default page size.
     75|     const listed = repo.list({});
     76|     expect(listed.length).toBe(50);
       |                           ^
     77| 
     78|     // Verify explicit high-limit requests can fetch beyond the old 50…

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯



shinetata:
Thanks for the genuinely thorough investigation here, @chiefmojo — the bisect table, the 687MB repro, and the root-cause write-up on the migration-007 transaction timeout are all excellent and much appreciated.

I'm going to close this PR, because the crash-loop it targets (#1787) has already been fixed on `main` by #1871, and #1787 is now closed.

The merged fix took a simpler route to the same goal:

- It **removed** the O(n) `UPDATE traces SET share_scope='private' WHERE share_scope IS NULL` entirely. That bulk rewrite — re-validating the JSON CHECK constraints on every row — was the actual cause of the CPU-bound hang you diagnosed.
- `share_scope` is now added via `ADD COLUMN ... DEFAULT 'private'` (O(1) since SQLite 3.35), and the read path already treats a NULL scope as `private` through `COALESCE(share_scope,'private')`, so no backfill pass is needed.
- On top of that, `001-initial.sql` now folds in the namespace/visibility columns and their indexes, so on fresh installs migration 007 is effectively a no-op.

Given that, this PR can't be merged as-is:

1. It's currently **behind `main`** and rewrites the exact function (`ensureNamespaceVisibilityColumns`) that #1871 already replaced, so it would conflict heavily on rebase.
2. It also changes `tests/unit/storage/traces-count.test.ts` (the `500` → `50/600` expectations), which is unrelated to the migration fix and isn't backed by a matching source change. On current `main`, `list()` still defaults to and caps at 500, so that assertion fails — which is exactly what the automated run flagged (`expected 500 to be 50`). This looks like a stray rebase artifact.
3. The `migrator.test.ts` change inverts the design chosen in #1871: it asserts phase-2 backfills `NULL → 'private'`, whereas `main` intentionally leaves `NULL` in place and coalesces at read time.

If you think the one piece of net-new hardening here is still worth having — moving `CREATE INDEX` **outside** the migration transaction so index builds on very large tables can't hold the bootstrap tx — that would be a welcome, focused follow-up. A small PR rebased on the latest `main` that (a) only touches the index-creation path and (b) drops the `traces-count.test.ts` change would be easy to review and land.

One more thing that surfaced while digging into this: the #1787 regression test on `main` is currently red on its own — it inserts legacy `role`/`embedding` columns that no longer exist in the folded `traces` schema (`SqliteError: table traces has no column named role`). We'll fix that separately, but your rewrite of that test was actually pointing at the correct schema.

Thanks again for the deep debugging here — it genuinely helped confirm the root cause.

PR #705, opened 2025-12-13
fix: update deprecated APIs for chonkie v1.4.0 and qdrant-client v1.16.0

zhixiangxue:
## 🔧 Changes

This PR addresses breaking changes and deprecations from upstream dependencies:

### 1. Chonkie API Update (v1.4.0)
- **Issue**: Breaking change in chonkie v1.4.0 removes `tokenizer_or_token_counter` parameter
- **Reference**: https://github.com/chonkie-inc/chonkie/releases/tag/v1.4.0
- **Change**: Rename `tokenizer_or_token_counter` → `tokenizer` in `SentenceChunker`
- **File**: `src/memos/chunkers/sentence_chunker.py`

### 2. Qdrant Client API Update (v1.16.0)
- **Issue**: Deprecated `search()` method removed in qdrant-client v1.16.0 (#1103)
- **Reference**: https://github.com/qdrant/qdrant-client/releases/tag/v1.16.0
- **Change**: Replace `client.search()` with `client.query_points().points`
- **File**: `src/memos/vec_dbs/qdrant.py`

### 3. Python Datetime Deprecation
- **Issue**: `datetime.utcnow()` deprecated in Python 3.12+
- **Change**: Replace with `datetime.now(timezone.utc)`
- **File**: `src/memos/mem_os/core.py`

### 4. Dependency Version Updates
- `chonkie`: `>=1.0.7` → `>=1.4.0`
- `qdrant-client`: `>=1.14.2` → `>=1.16.0`

## ✅ Testing

- [x] Tested with chonkie v1.4.0
- [x] Tested with qdrant-client v1.16.0
- [x] All affected components working as expected

## 📚 References

- Chonkie v1.4.0 Release Notes: https://github.com/chonkie-inc/chonkie/releases/tag/v1.4.0
- Qdrant Client v1.16.0 Release Notes: https://github.com/qdrant/qdrant-client/releases/tag/v1.16.0

PR #1804, opened 2026-05-25
fix(chunkers): fall back to 'character' tokenizer when configured one is unavailable

qianjindexiaozu:
## Problem

`ChonkieSentenceChunker` raises `ValueError` when none of `tokenizers` / `tiktoken` / `transformers` can load the configured tokenizer (default `gpt2`). The current code in `SentenceChunker.__init__` only catches `TypeError`/`AttributeError` (added by #705 for the chonkie v1.4.0 API switch), so the `ValueError` propagates and crashes the worker process during MOS initialization.

This is reproducible on the official `memos-base-arm:v1.0` Docker image — it ships without `tiktoken`, and HuggingFace is unreachable from many networks (notably mainland China), so the `gpt2` tokenizer load fails through every fallback path inside chonkie.

## Reproduction

1. Start `docker/docker-compose.yml` as documented.
2. Worker process logs:
   ```
   chonkie/tokenizer.py: Could not load tokenizer with 'tokenizers'. Falling back to 'tiktoken'.
   chonkie/tokenizer.py: 'tiktoken' library not found. Falling back to 'transformers'.
   ValueError: Tokenizer not found in transformers, tokenizers, or tiktoken
   ```
3. `/health` endpoint returns nothing — the uvicorn worker keeps crashing on reload.

## Fix

After both API attempts (`tokenizer=` for chonkie >=1.4.0, `tokenizer_or_token_counter=` for older) fail, fall back to chonkie's built-in `character` counter, which requires no external libraries or downloads. The original error is preserved in a `WARNING` log so operators can see why the fallback fired.

## Tradeoff

`chunk_size` semantics change from token count to character count under fallback — chunks become roughly 4x smaller for English text. A working chunker beats a crashing worker, and operators who need token-accurate chunking can install `tiktoken` (or fix HuggingFace network access) to restore the original behavior. The fallback is announced loudly via WARNING log so the degradation isn't silent.

## Tests

- `test_sentence_chunker_falls_back_to_character` — regression test: when configured tokenizer raises `ValueError`, fallback engages and the chunker is constructed with `tokenizer_or_token_counter='character'`.
- `test_sentence_chunker_no_warning_when_character_configured` — guards against firing a misleading \"character unavailable, using character\" warning for users who deliberately picked the character counter.

All 3 tests pass (`pytest tests/chunkers/test_sentence_chunker.py`).

### DIFF
--- a/src/memos/chunkers/sentence_chunker.py
+++ b/src/memos/chunkers/sentence_chunker.py
@@ -26,18 +26,10 @@ class SentenceChunker(BaseChunker):
             "chunk_overlap": config.chunk_overlap,
             "min_sentences_per_chunk": config.min_sentences_per_chunk,
         }
-        self.chunker = None
-        last_error: Exception | None = None
-        # Try chonkie >=1.4.0 API first, then the pre-1.4 signature.
-        for kwarg in ("tokenizer", "tokenizer_or_token_counter"):
-            try:
-                self.chunker = ChonkieSentenceChunker(
-                    **{kwarg: config.tokenizer_or_token_counter}, **common_kwargs
-                )
-                break
-            except (TypeError, AttributeError, ValueError) as e:
-                last_error = e
-                continue
+
+        self.chunker, used_kwarg, last_error = self._build_chonkie_chunker(
+            ChonkieSentenceChunker, config.tokenizer_or_token_counter, common_kwargs
+        )
 
         # If the configured tokenizer can't be loaded (no tiktoken, no
         # HuggingFace access, etc.), fall back to chonkie's built-in
@@ -49,11 +41,35 @@ class SentenceChunker(BaseChunker):
                 f"Tokenizer '{config.tokenizer_or_token_counter}' unavailable "
                 f"({last_error!r}); falling back to 'character'"
             )
-            self.chunker = ChonkieSentenceChunker(
-                tokenizer_or_token_counter="character", **common_kwargs
+            self.chunker, used_kwarg, fallback_error = self._build_chonkie_chunker(
+                ChonkieSentenceChunker, "character", common_kwargs
             )
+            if self.chunker is None:
+                raise RuntimeError(
+                    "Failed to initialize SentenceChunker: 'character' fallback "
+                    f"also rejected by chonkie ({fallback_error!r})."
+                ) from last_error
 
+        logger.debug(f"chonkie SentenceChunker initialized via kwarg {used_kwarg!r}")
         logger.info(f"Initialized SentenceChunker with config: {config}")
+
+    @staticmethod
+    def _build_chonkie_chunker(chunker_cls, tokenizer_value, common_kwargs):
+        """Build a chonkie SentenceChunker, tolerating the 1.4.0 kwarg rename.
+
+        chonkie >=1.4.0 renamed ``tokenizer_or_token_counter`` to ``tokenizer``.
+        Try the new keyword first, then the legacy one. Returns
+        ``(instance_or_None, used_kwarg_or_None, last_error_or_None)``.
+        """
+        last_error: Exception | None = None
+        for kwarg in ("tokenizer", "tokenizer_or_token_counter"):
+            try:
+                instance = chunker_cls(**{kwarg: tokenizer_value}, **common_kwargs)
+                return instance, kwarg, None
+            except (TypeError, AttributeError, ValueError) as e:
+                last_error = e
+                continue
+        return None, None, last_error
 
     def chunk(self, text: str) -> list[str] | list[Chunk]:
         """Chunk the given text into smaller chunks based on sentences."""
--- a/src/memos/chunkers/sentence_chunker.py
+++ b/src/memos/chunkers/sentence_chunker.py
@@ -21,22 +21,36 @@ class SentenceChunker(BaseChunker):
 
         self.config = config
 
-        # Try new API first (v1.4.0+)
-        try:
-            self.chunker = ChonkieSentenceChunker(
-                tokenizer=config.tokenizer_or_token_counter,
-                chunk_size=config.chunk_size,
-                chunk_overlap=config.chunk_overlap,
-                min_sentences_per_chunk=config.min_sentences_per_chunk,
+        common_kwargs = {
+            "chunk_size": config.chunk_size,
+            "chunk_overlap": config.chunk_overlap,
+            "min_sentences_per_chunk": config.min_sentences_per_chunk,
+        }
+        self.chunker = None
+        last_error: Exception | None = None
+        # Try chonkie >=1.4.0 API first, then the pre-1.4 signature.
+        for kwarg in ("tokenizer", "tokenizer_or_token_counter"):
+            try:
+                self.chunker = ChonkieSentenceChunker(
+                    **{kwarg: config.tokenizer_or_token_counter}, **common_kwargs
+                )
+                break
+            except (TypeError, AttributeError, ValueError) as e:
+                last_error = e
+                continue
+
+        # If the configured tokenizer can't be loaded (no tiktoken, no
+        # HuggingFace access, etc.), fall back to chonkie's built-in
+        # 'character' counter so the chunker still works offline. Note:
+        # chunk_size semantics change from token count to character count
+        # for fallback runs.
+        if self.chunker is None:
+            logger.warning(
+                f"Tokenizer '{config.tokenizer_or_token_counter}' unavailable "
+                f"({last_error!r}); falling back to 'character'"
             )
-        except (TypeError, AttributeError) as e:
-            # Fallback to old API (<v1.4.0)
-            logger.debug(f"Falling back to old chonkie API: {e}")
             self.chunker = ChonkieSentenceChunker(
-                tokenizer_or_token_counter=config.tokenizer_or_token_counter,
-                chunk_size=config.chunk_size,
-                chunk_overlap=config.chunk_overlap,
-                min_sentences_per_chunk=config.min_sentences_per_chunk,
+                tokenizer_or_token_counter="character", **common_kwargs
             )
 
         logger.info(f"Initialized SentenceChunker with config: {config}")
--- a/tests/chunkers/test_sentence_chunker.py
+++ b/tests/chunkers/test_sentence_chunker.py
@@ -102,17 +102,28 @@ class TestSentenceChunker(unittest.TestCase):
         )
 
     def test_sentence_chunker_no_warning_when_character_configured(self):
-        """When 'character' is explicitly configured, no fallback warning is emitted.
+        """When 'character' is explicitly configured, the initial build
+        succeeds and no fallback warning is emitted.
 
-        Guards against a regression where the fallback warning fires for
-        users who deliberately picked the character counter.
+        Mock setup: only ``character`` is accepted; any other value raises
+        ``ValueError``. With ``character`` configured, the first iteration
+        of the build loop succeeds, so the fallback branch is never
+        entered — proving that a user who explicitly picked the character
+        counter does not see the "Tokenizer 'character' unavailable;
+        falling back to 'character'" warning.
         """
         import logging
 
         def side_effect(*args, **kwargs):
             if "tokenizer" in kwargs:
-                raise TypeError("unexpected keyword argument 'tokenizer'")
-            return MagicMock()
+                value = kwargs["tokenizer"]
+            elif "tokenizer_or_token_counter" in kwargs:
+                value = kwargs["tokenizer_or_token_counter"]
+            else:
+                raise TypeError("no tokenizer kwarg")
+            if value == "character":
+                return MagicMock()
+            raise ValueError(f"Tokenizer not found: {value}")
 
         records: list[logging.LogRecord] = []
 
@@ -124,7 +135,7 @@ class TestSentenceChunker(unittest.TestCase):
         logger_under_test = logging.getLogger("memos.chunkers.sentence_chunker")
         logger_under_test.addHandler(handler)
         try:
-            with patch("chonkie.SentenceChunker", side_effect=side_effect):
+            with patch("chonkie.SentenceChunker", side_effect=side_effect) as mock_cls:
                 config = ChunkerConfigFactory.model_validate(
                     {
                         "backend": "sentence",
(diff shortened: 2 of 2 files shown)


Memtensor-AI (commented):
<thinking>
</thinking>

## Needs minor changes

The approach is sound — graceful degradation beats a crashing worker. The loop-based retry and the fallback to `character` are clean. A few issues:

### 1. Fallback can use wrong API keyword (bug)

The final fallback unconditionally uses `tokenizer_or_token_counter="character"`, but if we're on chonkie >=1.4.0, that keyword doesn't exist — it expects `tokenizer=`. The fallback itself could raise `TypeError`.

```python
# Current:
if self.chunker is None:
    self.chunker = ChonkieSentenceChunker(
        tokenizer_or_token_counter="character", **common_kwargs
    )
```

Consider wrapping the fallback in the same loop pattern, or at minimum a try/except that tries both keywords:

```python
if self.chunker is None:
    logger.warning(...)
    for kwarg in ("tokenizer", "tokenizer_or_token_counter"):
        try:
            self.chunker = ChonkieSentenceChunker(
                **{kwarg: "character"}, **common_kwargs
            )
            break
        except (TypeError, AttributeError):
            continue
    if self.chunker is None:
        raise RuntimeError("Failed to initialize chunker even with 'character' fallback") from last_error
```

### 2. Missing terminal error if fallback also fails

If the `character` fallback constructor raises for any reason, the code falls through to `logger.info("Initialized SentenceChunker...")` with `self.chunker` set to whatever that raises — i.e., it'll just propagate an unrelated exception. Adding a guard after the fallback (or wrapping it) would give a clearer error message.

### 3. Test `test_sentence_chunker_no_warning_when_character_configured` doesn't actually test the described scenario

The `side_effect` raises `TypeError` on the new-API attempt, then succeeds on the old-API attempt regardless of the value. So the chunker succeeds on the *first loop iteration's second keyword* — it never reaches the fallback path. The test passes, but it's testing "configured tokenizer loads fine on old API" rather than "character is explicitly configured and no warning fires." This is fine as a regression guard but the docstring is slightly misleading.

### 4. Type annotation style

```python
last_error: Exception | None = None
```

This requires Python 3.10+. If the project supports 3.9, use `Optional[Exception]` or `from __future__ import annotations`.

### 5. Minor: the loop swallows the debug-level message

The old code logged at `DEBUG` when falling back from new to old API. Now that information is lost — operators debugging chonkie version issues won't see which keyword worked. Consider adding a `logger.debug` inside the loop on success or on the first `continue`.

---

Overall the fix addresses a real production issue with a reasonable tradeoff. The main concern is item 1 — the fallback itself isn't version-resilient.

qianjindexiaozu:
Thanks for the thorough review! Pushed commit `b5268e1d` addressing all five points.

### 1. Fallback API keyword (bug) — fixed
Extracted the version-resilient build loop into `SentenceChunker._build_chonkie_chunker(...)` and reused it for the `'character'` fallback. Both the primary attempt and the fallback now try `tokenizer=` first, then `tokenizer_or_token_counter=`, so the fallback works on chonkie pre-1.4 and ≥1.4.0 alike.

### 2. Terminal error if fallback also fails — fixed
If the fallback returns `None`, the code now raises:
```python
raise RuntimeError(
    "Failed to initialize SentenceChunker: 'character' fallback "
    f"also rejected by chonkie ({fallback_error!r})."
) from last_error
```
The original tokenizer-load error is preserved via `__cause__` so the traceback still points at the root cause.

### 3. Misleading test docstring — fixed
The `test_sentence_chunker_no_warning_when_character_configured` mock now only accepts the value `character` (regardless of which API kwarg is used), and the test asserts `mock_cls.call_count == 1`. This makes it true to the docstring: with `character` configured, the **first** build-loop iteration succeeds and the fallback branch is provably never entered.

I also added two new regression tests:
- `test_sentence_chunker_fallback_uses_new_api_for_character` — simulates chonkie ≥1.4.0 (legacy keyword removed) with the configured tokenizer failing, and verifies the fallback successfully uses `tokenizer=\"character\"` (catches the bug from point 1).
- `test_sentence_chunker_raises_when_fallback_also_fails` — verifies the chained `RuntimeError` from point 2, with the original `ValueError` preserved via `__cause__`.

### 4. Type annotation Python compatibility — no change needed
`pyproject.toml` declares `requires-python = \">=3.10\"`, so the PEP 604 `Exception | None` syntax is fine. Happy to switch to `Optional[Exception]` if you'd prefer for consistency with the rest of the codebase, but didn't want to silently broaden the project's stated minimum.

### 5. DEBUG log for which kwarg chonkie accepted — added
```python
logger.debug(f\"chonkie SentenceChunker initialized via kwarg {used_kwarg!r}\")
```
emitted after either the primary or fallback build succeeds, so operators can tell whether chonkie's new or legacy API was hit.

Memtensor-AI:

- Run: `tr-f9c07e89-247` on cloud test-engine `10010`
- Verdict: `fail_code_bug`
- Summary: Tests failed. Failed cases: test_sentence_chunker, test_sentence_chunker_fallback_uses_new_api_for_character, test_sentence_chunker_falls_back_to_character, test_sentence_chunker_no_warning_when_character_configured, test_sentence_chunker_raises_when_fallback_also_fails

- memos_python_core/changed-repo-python: 0 passed, 5 failed, 0 skipped

Failed cases:
- test_sentence_chunker: /data/test-workspaces/manual-pr-1804-after-disk-cleanup-20260702104253/repo/tests/chunkers/test_sentence_chunker.py:12: in test_sentence_chunker
    with patch("chonkie.SentenceChunker") as mock_chunker_cls:
/usr/lib/python3.10/unittest/mock.py:1431: in __enter__
    self.target = self.getter()
/usr/lib/python3.10/unittest/mock.py:1618: in <lambda>
    getter = lambda: _importer(target)
/usr/lib/python3.10/unittest/mock.py:1257: in _importer
    thing = __import__(import_path)
E   ModuleNotFound
- test_sentence_chunker_fallback_uses_new_api_for_character: /data/test-workspaces/manual-pr-1804-after-disk-cleanup-20260702104253/repo/tests/chunkers/test_sentence_chunker.py:181: in test_sentence_chunker_fallback_uses_new_api_for_character
    with patch("chonkie.SentenceChunker", side_effect=side_effect) as mock_cls:
/usr/lib/python3.10/unittest/mock.py:1431: in __enter__
    self.target = self.getter()
/usr/lib/python3.10/unittest/mock.py:1618: in <lambda>
    getter = lambda: _importer(target)
/usr/lib/python3.10/unittest/mock.py:1257: in _importer

- test_sentence_chunker_falls_back_to_character: /data/test-workspaces/manual-pr-1804-after-disk-cleanup-20260702104253/repo/tests/chunkers/test_sentence_chunker.py:84: in test_sentence_chunker_falls_back_to_character
    with patch("chonkie.SentenceChunker", side_effect=side_effect) as mock_chunker_cls:
/usr/lib/python3.10/unittest/mock.py:1431: in __enter__
    self.target = self.getter()
/usr/lib/python3.10/unittest/mock.py:1618: in <lambda>
    getter = lambda: _importer(target)
/usr/lib/python3.10/unittest/mock.py:1257: in _importer
    t
- test_sentence_chunker_no_warning_when_character_configured: /data/test-workspaces/manual-pr-1804-after-disk-cleanup-20260702104253/repo/tests/chunkers/test_sentence_chunker.py:138: in test_sentence_chunker_no_warning_when_character_configured
    with patch("chonkie.SentenceChunker", side_effect=side_effect) as mock_cls:
/usr/lib/python3.10/unittest/mock.py:1431: in __enter__
    self.target = self.getter()
/usr/lib/python3.10/unittest/mock.py:1618: in <lambda>
    getter = lambda: _importer(target)
/usr/lib/python3.10/unittest/mock.py:1257: in _importer
- test_sentence_chunker_raises_when_fallback_also_fails: /data/test-workspaces/manual-pr-1804-after-disk-cleanup-20260702104253/repo/tests/chunkers/test_sentence_chunker.py:218: in test_sentence_chunker_raises_when_fallback_also_fails
    with patch("chonkie.SentenceChunker", side_effect=side_effect):
/usr/lib/python3.10/unittest/mock.py:1431: in __enter__
    self.target = self.getter()
/usr/lib/python3.10/unittest/mock.py:1618: in <lambda>
    getter = lambda: _importer(target)
/usr/lib/python3.10/unittest/mock.py:1257: in _importer
    thing = __im

Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #1804
**Task:** `manual-ocr:MemTensor/MemOS#1804`
**Base:** `main`
**Head:** `fix/sentence-chunker-tokenizer-fallback`

---

### 1. `tests/chunkers/test_sentence_chunker.py` (L128-L136)

The `_Capture` handler is created with `level=logging.WARNING`, but the *logger's* own effective level is never set. If the logger's effective level is above WARNING in the test environment (e.g., inherited from the root logger configured to ERROR or higher), no records will propagate to the handler regardless of the handler's level filter, and the assertion `fallback_warnings == []` will always pass trivially — even if the production code actually emits a warning.

Add `logger_under_test.setLevel(logging.WARNING)` and restore the original level in the `finally` block to guarantee the assertion is meaningful.

**Before:**
```
        records: list[logging.LogRecord] = []

        class _Capture(logging.Handler):
            def emit(self, record: logging.LogRecord) -> None:
                records.append(record)

        handler = _Capture(level=logging.WARNING)
        logger_under_test = logging.getLogger("memos.chunkers.sentence_chunker")
        logger_under_test.addHandler(handler)
```

**After:**
```suggestion
        records: list[logging.LogRecord] = []

        class _Capture(logging.Handler):
            def emit(self, record: logging.LogRecord) -> None:
                records.append(record)

        handler = _Capture(level=logging.WARNING)
        logger_under_test = logging.getLogger("memos.chunkers.sentence_chunker")
        original_level = logger_under_test.level
        logger_under_test.setLevel(logging.WARNING)
        logger_under_test.addHandler(handler)
```

</details>

---

### 2. `tests/chunkers/test_sentence_chunker.py` (L149-L150)

If `logger_under_test.setLevel(logging.WARNING)` is added (as suggested above), the original level must be restored in the `finally` block to avoid polluting other tests that share the same logger instance.

**Before:**
```
        finally:
            logger_under_test.removeHandler(handler)
```

**After:**
```suggestion
        finally:
            logger_under_test.removeHandler(handler)
            logger_under_test.setLevel(original_level)
```

</details>

---

### 3. `tests/chunkers/test_sentence_chunker.py` (L115-L117)

The `import logging` statement is placed inside the test method body, while all other imports in the file are at module level. Move it to the top of the file alongside the other imports (`import unittest`, etc.) for consistency.

**Before:**
```
        import logging

        def side_effect(*args, **kwargs):
```

**After:**
```suggestion
        def side_effect(*args, **kwargs):
```

</details>

---

### 4. `src/memos/chunkers/sentence_chunker.py` (L69-L71)

Catching `ValueError` here is too broad. While chonkie raises `ValueError` when a tokenizer cannot be found (the intended case), it may also raise `ValueError` for genuinely invalid configuration parameters such as a negative `chunk_size`, out-of-range `chunk_overlap`, or invalid `min_sentences_per_chunk`. Catching these silently causes the loop to retry the other kwarg name, which will also fail for the same reason, ultimately triggering the 'character' fallback — completely masking the root cause (bad config) with a misleading warning about tokenizer unavailability.

Consider inspecting the exception message to distinguish kwarg-name errors from tokenizer-load errors, or only catch `TypeError` and `AttributeError` for the kwarg-probe pass and re-raise any `ValueError` that is not related to tokenizer loading. Alternatively, validate config parameters before calling `_build_chonkie_chunker`.

**Before:**
```
            except (TypeError, AttributeError, ValueError) as e:
                last_error = e
                continue
```

**After:**
```suggestion
            except (TypeError, AttributeError) as e:
                last_error = e
                continue
            except ValueError as e:
                # ValueError from chonkie can mean either:
                # (a) tokenizer not found — retry with other kwarg name
                # (b) invalid config param — should propagate immediately
                # Heuristic: kwarg-name mismatches produce TypeError, not ValueError;
                # re-raise if the message doesn't look like a tokenizer-load failure.
                if "tokenizer" not in str(e).lower():
                    raise
                last_error = e
                continue
```

</details>

---

### 5. `src/memos/chunkers/sentence_chunker.py` (L48-L51)

The exception message includes `fallback_error` (the error from the 'character' fallback attempt), but the exception chain (`from last_error`) attaches the *original* tokenizer error. This is semantically correct in terms of causality, but can be confusing to a debugger who reads the message (showing the character fallback failure) vs. `__cause__` (showing the original tokenizer error). Consider including both errors in the message for clarity.

**Before:**
```
                raise RuntimeError(
                    "Failed to initialize SentenceChunker: 'character' fallback "
                    f"also rejected by chonkie ({fallback_error!r})."
                ) from last_error
```

**After:**
```suggestion
                raise RuntimeError(
                    "Failed to initialize SentenceChunker: original tokenizer error: "
                    f"{last_error!r}; 'character' fallback also failed: {fallback_error!r}."
                ) from last_error
```

</details>

---

### 6. `src/memos/chunkers/sentence_chunker.py` (L70-L71)

The `continue` statement at the end of the `except` block is redundant — control naturally flows to the next loop iteration. Remove it to reduce noise and avoid implying intentional non-obvious flow control.

**Before:**
```
                last_error = e
                continue
```

**After:**
```suggestion
                last_error = e
```

</details>

---

### 7. `tests/chunkers/test_sentence_chunker.py` (L134-L137)

The `_Capture` handler is created with `level=logging.WARNING`, but the logger's own effective level is never explicitly set. If the test environment's root logger is configured above WARNING (e.g., ERROR), no records will be delivered to this handler regardless of the handler's level — making the `fallback_warnings == []` assertion pass trivially even if the production code actually emits a warning.

Add `logger_under_test.setLevel(logging.WARNING)` before the test body, and restore the original level in the `finally` block.

**Before:**
```
        handler = _Capture(level=logging.WARNING)
        logger_under_test = logging.getLogger("memos.chunkers.sentence_chunker")
        logger_under_test.addHandler(handler)
        try:
```

**After:**
```suggestion
        handler = _Capture(level=logging.WARNING)
        logger_under_test = logging.getLogger("memos.chunkers.sentence_chunker")
        original_level = logger_under_test.level
        logger_under_test.setLevel(logging.WARNING)
        logger_under_test.addHandler(handler)
        try:
```

</details>

---

🧹 Filtered **2** low-confidence OCR finding(s) before posting/fix-loop (duplicate: 2).

Memtensor-AI:

**Details:** All 5 tests failed with ModuleNotFoundError: No module named 'chonkie'. The test environment is missing the 'chonkie' package which is a required dependency for the SentenceChunker module. [advisory, non-gating] AI-generated tests on branch test/auto-gen-452894bcf8919e02-20260709003536: 58/61 passed, 3 failed — these do NOT affect the PR verdict; review the branch manually.
**Branch:** `fix/sentence-chunker-tokenizer-fallback`

PR #1816, opened 2026-05-28
fix: bridge startup race, dirty-reward orphan traces, RPC loopback auth

fayenix:
Three fixes from companion-stable, tested on three Hermes instances.

### 1. Skip orphan trace insert during dirty-reward recovery
capture.ts — stops empty trace generation from dirty-reward loop (saved 44K traces / 268MB on one instance).

### 2. Bridge startup race fix
daemon_manager.py + __init__.py — 45s poll deadline with backoff, 60s session.open timeout, double-spawn prevention.

### 3. RPC loopback auth restriction
auth.ts — /api/v1/rpc now requires session cookie from non-loopback callers.

### DIFF
--- a/apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py
+++ b/apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py
@@ -63,8 +63,8 @@ _PLUGIN_DIR = Path(__file__).resolve().parent
 if str(_PLUGIN_DIR) not in sys.path:
     sys.path.insert(0, str(_PLUGIN_DIR))
 
-from bridge_client import BridgeError, MemosBridgeClient  # noqa: E402
-from daemon_manager import ensure_bridge_running, ensure_viewer_daemon  # noqa: E402
+from bridge_client import BridgeError, MemosBridgeClient, MemosHttpClient  # noqa: E402
+from daemon_manager import ensure_bridge_running, ensure_viewer_daemon, probe_viewer_status, kill_zombie_bridges  # noqa: E402
 
 
 try:  # pragma: no cover — host-provided base class, absent in unit tests
@@ -243,7 +243,7 @@ class MemTensorProvider(MemoryProvider):
     """
 
     def __init__(self) -> None:
-        self._bridge: MemosBridgeClient | None = None
+        self._bridge: MemosBridgeClient | MemosHttpClient | None = None
         self._reconnect_lock = threading.Lock()
         self._session_id: str = ""
         self._episode_id: str = ""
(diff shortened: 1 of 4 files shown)


Memtensor-AI (commented):
<thinking>
</thinking>

## Needs changes

Overall the fixes are well-motivated and the capture.ts guard is a solid improvement. A few issues worth addressing:

### 1. `__init__.py` — Duplicated connection logic begs for extraction

The HTTP-bridge-connect-and-register pattern is copy-pasted 3 times (initial connect, late-probe fallback, reconnect). If the handshake ever changes (e.g. a second handler registration), you'll need to patch all three. Extract a helper like `_try_http_bridge(session_id, timeout) -> bool`.

### 2. `__init__.py` — `http_bridge` may be unbound on exception

```python
http_bridge: MemosHttpClient | None = MemosHttpClient()
```

If `MemosHttpClient()` itself raises, `http_bridge` is never assigned, yet the `except` block calls `http_bridge.close()`. The `type: ignore` comment masks this. Safer:

```python
http_bridge: MemosHttpClient | None = None
http_bridge = MemosHttpClient()
```

Same issue with `http_bridge_late`.

### 3. `daemon_manager.py` — Backoff heuristic is inverted

```python
time.sleep(0.5 if (deadline - time.time()) > 30 else 2.0)
```

This sleeps 0.5s when there's >30s remaining (i.e. the first 15s) and 2.0s after. The comment says "fast probes for the first 15s, then back off" — but `deadline - now > 30` is true only during the first ~15s of a 45s window. That's correct, but fragile and non-obvious. A simple elapsed-time check would be clearer:

```python
elapsed = time.time() - start
time.sleep(0.5 if elapsed < 15 else 2.0)
```

You'd need to capture `start = time.time()` before the loop.

### 4. `auth.ts` — Loopback check is bypassable via `X-Forwarded-For` if behind a reverse proxy

`req.socket.remoteAddress` is reliable only for direct connections. If the viewer is ever fronted by a proxy (nginx, Caddy, etc.), all requests appear as loopback. This is fine today for a local-only plugin, but worth a comment or a check that no `X-Forwarded-For` header is present when granting loopback trust:

```ts
const hasForwardedHeader = !!req.headers["x-forwarded-for"];
if (isLoopback && !hasForwardedHeader) return true;
```

### 5. `capture.ts` — Reliance on `meta.recoveryReason` magic string

```ts
if (input.episode.meta?.recoveryReason === "dirty_reward_rescore") {
```

If the recovery codepath ever changes this string or omits it, the loop silently returns. Consider also checking `traceCount !== traceIds.length` as a secondary signal, or at minimum add a constant/enum rather than a raw string comparison in two different files.

### 6. Minor: `kill_zombie_bridges()` swallowed silently

```python
except Exception:
    pass
```

At least log at `debug` level — zombie detection failures could mask the very problem this PR fixes.

---

The capture.ts fix is the highest-value change and looks correct. The Python side works but would benefit from deduplication and the unbound-variable fix before merge.

fayenix:
Closing — opened under wrong account. Will reopen as chiefmojo.

PR #1842, opened 2026-05-29
fix(daemon): start HTTP server before core.init() to prevent zombie process avalanche

### LINKED ISSUE #1841: fix: Daemon viewer fails to start when core.init() blocks on dirty episode rescoring — triggers zombie process avalanche
### Summary

In daemon mode (`--daemon`), the HTTP viewer server is started **after** `core.init()` completes. When `init()` blocks on LLM-backed recovery tasks (dirty closed episode rescoring), the viewer never binds its port. The Hermes gateway daemon manager then detects the port as "free" and spawns additional bridge processes, each of which kills the previous one and gets stuck in `init()` again — creating a positive feedback loop of zombie processes that exhaust the LLM API rate limit.

### Environment

- **OS:** macOS (Apple Silicon)
- **Plugin:** memos-local-plugin v2.0.5
- **Host agent:** Hermes Agent
- **LLM provider:** SiliconFlow (rate-limited)
- **Node.js:** v20+
- **Relevant files:**
  - `apps/memos-local-plugin/bridge.cts` (lines 325–396)
  - `apps/memos-local-plugin/core/pipeline/memory-core.ts` (lines 869–927, 1265–1297)
  - `apps/memos-local-plugin/adapters/hermes/memos_provider/daemon_manager.py` (lines 231–329)

### Root Cause

There are two interrelated bugs:

**Bug 1 — Daemon viewer startup blocked by synchronous `core.init()`**

In `bridge.cts`, the startup sequence is:

```
bootstrapMemoryCoreFull → await core.init() → startHttpServer
```

`core.init()` performs orphan episode recovery including `recoverDirtyClosedEpisodes()`, which calls the LLM to rescore each dirty episode. If any episode has a stale `reward.traceCount` that doesn't match its actual `traceIds.length`, `episodeRewardIsDirty()` returns `true` on every init, triggering a rescore that involves one or more LLM API calls per episode.

In daemon mode, `startHttpServer` is called **after** `core.init()` returns (line 355). If `init()` blocks for longer than the startup timeout (120s in `install.sh`, 15s in `daemon_manager.py`), the viewer port never opens and the startup is considered failed.

In stdio mode this is acceptable because `startStdioServer` is called before `core.init()` (line 316), so the host agent can still function. But daemon mode is a pure HTTP server — if init blocks, there is no service at all.

**Bug 2 — Zombie process avalanche from gateway auto-spawn**

When the daemon fails to start the viewer within the timeout:
1. `daemon_manager.py::ensure_viewer_daemon()` probes the port and sees it as `"free"` (connection refused)
2. It spawns a new bridge process with `--daemon`
3. The new process calls `killExistingBridge()` (line 152), killing the previous stuck process
4. The new process then gets stuck in `core.init()` itself
5. The next probe cycle repeats the cycle

This creates 50+ zombie bridge processes all hitting the LLM API simultaneously. The API rate limit (SiliconFlow 429) compounds the problem — all processes get rate-limited, none can complete init, and the gateway keeps spawning more.

This is a **positive feedback loop**: more processes → more API calls → more rate limiting → longer init times → gateway spawns more processes.

### Reproduction Steps

1. Have an episode where `reward.traceCount` diverges from `traceIds.length` (e.g., episode `ep_bg5nxx68sxh5` had `traceCount=116` but `traceIds.length=612`)
2. Ensure the LLM API is rate-limited (e.g., SiliconFlow free tier)
3. Run `install.sh` or restart the Hermes gateway
4. The daemon manager starts a bridge process with `--daemon`
5. `core.init()` detects the dirty episode and begins rescoring
6. LLM API calls hit rate limits (429), init blocks for >120s
7. `install.sh` reports "Memory Viewer not ready after 120s"
8. `daemon_manager.py` probes the port, sees it as free, spawns another bridge
9. Monitor with `ps aux | grep bridge` — zombie processes accumulate

### Expected Behavior

- The daemon viewer should start its HTTP server immediately, responding to health/ping probes
- `core.init()` recovery tasks should run in the background after the server is listening
- The viewer should serve cached (SQLite-backed) data while init completes
- Only one bridge process should exist at a time

### Actual Behavior

- The viewer port remains unresponsive for the duration of `core.init()` (potentially minutes)
- The daemon manager spawns multiple bridge processes in rapid succession
- Each process kills the previous one, then gets stuck in `init()` itself
- LLM API rate limits compound the problem across all concurrent processes
- The viewer never starts; install.sh times out

### Suggested Fix

In `bridge.cts`, reorder the daemon mode startup so the HTTP server starts **before** `core.init()`:

```
// Current (broken):
bootstrapMemoryCoreFull → await core.init() → startHttpServer → return

// Fixed:
bootstrapMemoryCoreFull → startHttpServer → void core.init() → return
```

Specifically:
1. In daemon mode, move the `startHttpServer` call (with retry logic) to execute **before** `core.init()`
2. Run `core.init()` as a fire-and-forget background task (`void core.init().then(...)`)
3. Log init completion or failure to stderr
4. Keep stdio mode unchanged (it already starts stdio before init)
5. The `/api/v1/ping` endpoint already works without init (hardcoded response)
6. The `/api/v1/health` endpoint works with the bootstrapped core (SQLite is already open)

This is backward-compatible: the viewer serves cached data from the bootstrapped SQLite database. Background recovery (orphan episodes, rescore) completes asynchronously and becomes visible in the viewer when done.

SonicBotMan:
Closes #1841

# PR: fix(daemon): start HTTP server before core.init() to prevent zombie process avalanche

## Summary

Reorders the daemon mode startup in `bridge.cts` so the HTTP viewer server binds its port **before** `core.init()` runs. In daemon mode, `core.init()` is now a fire-and-forget background task. Stdio mode is unchanged.

## Problem

See: [#1841] Daemon viewer fails to start when `core.init()` blocks on dirty episode rescoring — triggers zombie process avalanche

In daemon mode (`--daemon`), the startup sequence was:

```
bootstrapMemoryCoreFull -> await core.init() -> startHttpServer -> return
```

`core.init()` performs orphan episode recovery including `recoverDirtyClosedEpisodes()`, which calls the LLM to rescore each dirty episode. When an episode has a stale `reward.traceCount` that doesn't match its `traceIds.length`, `episodeRewardIsDirty()` returns `true` on every init, triggering rescore with LLM API calls. Under API rate limiting (e.g., SiliconFlow 429s), this can block for minutes.

Since the HTTP server starts **after** init, the viewer port never opens during this time. The Hermes daemon manager (`daemon_manager.py::ensure_viewer_daemon()`) probes the port, sees it as `"free"`, and spawns a replacement — which kills the current process and repeats the cycle. This creates a positive feedback loop of zombie processes.

## Fix

```
// Before (broken):
bootstrapMemoryCoreFull -> await core.init() -> startHttpServer -> return

// After (fixed):
bootstrapMemoryCoreFull -> startHttpServer -> void core.init() -> return
```

In daemon mode:
1. HTTP server starts first, responding to `/api/v1/ping` and `/api/v1/health` immediately
2. `core.init()` runs as a background task (`void core.init().then(...).catch(...)`)
3. Init completion or failure is logged to stderr
4. If init fails, the viewer continues in degraded mode (cached data still available)

In stdio mode (unchanged):
1. Stdio server starts before init (existing behavior)
2. `core.init()` is awaited as before
3. Host fallback is available during init recovery

## Why this is safe

- `/api/v1/ping` returns a hardcoded `{ ok: true, service: "memos-local-plugin" }` — no core state needed
- `/api/v1/health` calls `core.health()` which works with the bootstrapped core (SQLite is already open, LLM/embedder clients are already created)
- All viewer data endpoints read from the SQLite database, which is fully bootstrapped by `bootstrapMemoryCoreFull()` before either mode's init
- Background recovery (orphan episodes, rescore) runs asynchronously and becomes visible in the viewer when complete
- The daemon manager's 15s probe window is satisfied immediately instead of potentially timing out

## Changes

| File | Change |
|------|--------|
| `bridge.cts` | Reorder daemon mode: start HTTP server before `core.init()`, run init as background task |

## Testing

1. **Daemon mode starts immediately:** Run `node bridge.cts --daemon --agent=hermes` — the viewer should respond to `curl http://127.0.0.1:18800/api/v1/ping` within seconds, even before init completes
2. **Init runs in background:** Check stderr for `bridge: daemon core.init() completed (background recovery done)` after a delay
3. **Stdio mode unchanged:** Run `echo '{"jsonrpc":"2.0","method":"ping","id":1}' | node bridge.cts --agent=hermes --no-viewer` — should work as before
4. **No zombie processes:** During install.sh, only one bridge process should exist (`ps aux | grep bridge`)


Memtensor-AI:

Cloud test-engine rerun against `dev-v2.0.22` completed successfully after migrating this run from the old 10010 queue to the updated cloud test-engine.

- Run: `tr-ac648b1a-151` on cloud test-engine `10012`
- Result: 33 passed, 0 failed, 0 skipped

shinetata:
Thanks for the thorough write-up. We ended up not taking the reorder approach: on `main`, daemon mode intentionally keeps `core.init()` before the HTTP viewer binds (in `bridge.cts`, mirrored in the OpenClaw adapter), so health reflects actual readiness.

The #1841 zombie-avalanche is instead prevented from the process-management side — a PID-file singleton guard, a viewer-start cross-process lock, a 45s probe window, and `kill_zombie_bridges()` in `daemon_manager.py`. Since the acute issue is resolved and this branch conflicts with the reworked startup path, closing here. If you still hit an avalanche on latest `main`, please reopen with logs.

PR #1849, opened 2026-06-01
fix(bridge): detect silent pipeline init failure and configurable watchdog

chiefmojo:
## Summary

Two fixes to the daemon bridge startup path that prevent silent scoring outages when `core.init()` hangs.

### Problem
When the bridge starts in daemon mode, `core.init()` runs synchronously and can take 90–180 seconds for large databases (Qwen 235B scoring 250+ trace episodes). During this window:

1. The HTTP server is not yet serving, so the Python adapter's startup probe times out and spawns a replacement daemon — the original bridge is left running headless.
2. If `core.init()` neither resolves nor rejects (silent hang, e.g. LLM provider unreachable), the bridge appears healthy but the scoring pipeline is never wired. All subsequent episodes are captured but never scored.

### Fix 1: async `core.init()` with `pipelineReady` flag
`core.init()` is now kicked off in the background immediately after the HTTP server binds. Health endpoints report `pipelineReady: false` until init completes, allowing the adapter to distinguish "starting" from "stuck". The HTTP server stays responsive to probes throughout init.

### Fix 2: init watchdog (`initWatchdogMs`)
A configurable timeout (default 120 s, config key `bridge.initWatchdogMs`) races against `core.init()`. If init does not complete within the window, the bridge force-exits so the gateway respawns a fresh process rather than silently dropping scoring for hours.

**Note for operators:** Production experience shows scoring for large databases (250+ traces) can take 90–180 s. The 120 s default is intentionally conservative. Installations with large databases or slow LLM providers should set `initWatchdogMs` to at least 300 000 (5 minutes).

## Test plan
- [ ] Small DB: `core.init()` completes well before watchdog — normal startup
- [ ] Simulated slow init: `pipelineReady: false` visible in `/api/v1/health` during init
- [ ] Simulated hung init: bridge exits after `initWatchdogMs` ms, gateway respawns


Memtensor-AI:

- Run: `tr-cbfca4de-0dd` on cloud test-engine `10011`
- memos_local_plugin/unit: 136 passed, 0 failed, 0 skipped

shinetata:
Thanks! Neither the `pipelineReady` flag nor the `initWatchdog` force-exit made it in — `main` keeps the init-before-serve ordering and mitigates stuck/slow init via a cached-probe TTL and the viewer-start lock in `daemon_manager.py`. A watchdog that force-exits and respawns would also race the PID-file singleton guard that now owns daemon lifetime.

Closing as not-adopted; if silent scoring stalls still happen on latest `main`, a focused issue with logs would be the best next step.

PR #1862, opened 2026-06-02
fix(neo4j): cascade delete vectors from Qdrant when deleting memory nodes

AYin-Z:
## Description

`delete_node_by_prams()` only ran `DETACH DELETE` on Neo4j, leaving orphan vectors behind in Qdrant. Every memory deletion via the API created a leaked vector in Qdrant with no corresponding Neo4j node.

## Root Cause

The write path (`add_node` / `add_nodes_batch`) is symmetric — it writes to both Neo4j and Qdrant in the same method. But the delete path only touched Neo4j:

```python
# Before: count then delete Neo4j only
count_query = f"MATCH (n:Memory) WHERE ... RETURN count(n) AS node_count"
...
session.run("MATCH (n:Memory) WHERE ... DETACH DELETE n", **params)
# ← No vec_db.delete() call
```

Note that `clear()` and the newer `delete_node_by_mem_cube_id()` both correctly call `vec_db.delete()` — this was simply an omission in `delete_node_by_prams`.

## Fix

1. Replace `count_query` with an `ids_query` that collects node IDs
2. Delete from Neo4j (same as before)
3. Call `self.vec_db.delete(deleted_ids)` after successful Neo4j deletion

## Verification

- `BaseVecDB.delete(ids)` is an abstract method with working implementations in both QdrantVecDB and MilvusVecDB
- `self.vec_db` is initialized in the constructor and used elsewhere in the same class
- The fix gracefully falls back with a warning log if Qdrant cleanup fails

### DIFF
--- a/src/memos/graph_dbs/neo4j_community.py
+++ b/src/memos/graph_dbs/neo4j_community.py
@@ -1038,31 +1038,44 @@ class Neo4jCommunityGraphDB(Neo4jGraphDB):
             f"[delete_node_by_prams] Deleting nodes - memory_ids: {memory_ids}, file_ids: {file_ids}, filter: {filter}"
         )
 
-        # First count matching nodes to get accurate count
-        count_query = f"MATCH (n:Memory) WHERE {ids_where} RETURN count(n) AS node_count"
-        logger.info(f"[delete_node_by_prams] count_query: {count_query}")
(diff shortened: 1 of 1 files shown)


Memtensor-AI:

- Run: `tr-c67f3a71-38d` on cloud test-engine `10011`
- memos_python_core/changed-python-source: 2 passed, 0 failed, 0 skipped

PR #1917, opened 2026-06-12
fix(bridge): detect hermes chat with global flags in bridge status check

### LINKED ISSUE #1915: fix(memos-local-plugin): bridge status detection misses hermes chat with global flags
## Description

In `apps/memos-local-plugin/bridge.cts`, the `isHermesChatRunning()` function uses `pgrep -f "hermes chat"` to detect whether a Hermes chat session is active. This pattern only matches when `"hermes"` and `"chat"` are adjacent tokens in the process command line.

However, the `hermes` CLI supports global flags that are placed **before** the subcommand. For example, the `--skills` flag (and others like `-m`, `--provider`) can be used either as chat-level flags:

```
hermes chat --skills memory-routing
```

or as global flags:

```
hermes --skills memory-routing chat
```

In the latter case, the process command line becomes:

```
/usr/local/lib/hermes-agent/venv/bin/hermes --skills memory-routing chat
```

The substring `"hermes chat"` does not appear contiguously here, so `pgrep -f "hermes chat"` returns no match even though a chat session is actively running.

## Impact

This only affects the **daemon mode** bridge (`--daemon` flag), where `applyStaleRule()` uses `isHermesChatRunning()` to upgrade the status from `"disconnected"` to `"reconnecting"`. The stdio mode (the default when started via memtensor provider or systemd) is not affected — it manages its own lifecycle via the stdio connection directly.

## Suggested Fix

Replace the fixed-string pattern with a regex that allows optional flags between `"hermes"` and `"chat"`:

```diff
-    const out = childProcess.execFileSync("pgrep", ["-f", "hermes chat"], {
+    const out = childProcess.execFileSync("pgrep", ["-f", "hermes\\s.*chat\\b"], {
```

This pattern matches `"hermes"` followed by a whitespace character, then any characters (including flags), then `"chat"` at a word boundary. It covers both:

- `hermes chat` (adjacent, no flags)
- `hermes --skills memory-routing chat` (global flags before subcommand)
- `hermes chat --skills memory-routing` (flags after subcommand, already matched by the original pattern)

The risk of false positives is minimal — matching a process that contains `"hermes"` and `"chat"` in unrelated positions on the same command line is unlikely in practice.

## Environment

- Hermes Agent: any version supporting global flags before subcommands
- memos-local-plugin: any version (confirmed in v2.0.19 / npm v2.0.7)
- Platform: Linux (Ubuntu/Debian based)

lb1192176991-lab:
Fixes #1915

The `isHermesChatRunning()` function in `bridge.cts` uses `pgrep -f "hermes chat"` which fails when global flags are placed before the subcommand (e.g., `hermes --skills memory-routing chat`).

Changed the pattern to a regex `"hermes\\s.*chat\\b"` that matches:
- `hermes chat` (adjacent)
- `hermes --flags chat` (global flags before subcommand)
- `hermes chat --flags` (flags after subcommand, already worked)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,7 +1,7 @@
 name: "\U0001F41B Bug Report"
 description: Report a bug to help us improve MemOS | 报告错误以帮助我们改进 MemOS
 title: "fix: "
-labels: ["bug", "pending"]
+labels: ["bug"]
 body:
 
   - type: checkboxes
--- a/.github/ISSUE_TEMPLATE/feature-request.yml
+++ b/.github/ISSUE_TEMPLATE/feature-request.yml
@@ -1,7 +1,7 @@
 name: "\U0001F680 Feature request"
 description: Submit a request for a new feature | 申请添加新功能
 title: "feat: "
-labels: ["enhancement", "pending"]
+labels: ["enhancement"]
 body:
 
   - type: checkboxes
--- a/apps/memos-local-openclaw/index.ts
+++ b/apps/memos-local-openclaw/index.ts
@@ -13,6 +13,7 @@ import { createRequire } from "node:module";
 import { fileURLToPath } from "url";
 import { buildContext } from "./src/config";
 import type { HostModelsConfig } from "./src/openclaw-api";
+import { isPathInside } from "./src/path-utils";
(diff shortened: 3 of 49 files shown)


Memtensor-AI:
Closed as superseded by dev-v2.0.22. After merging dev-v2.0.22 into this branch, the PR has no remaining file diff. The Hermes chat global-flag detection fix is already present in bridge/hermes-process.ts with unit coverage.

PR #1921, opened 2026-06-15
fix(memos-local-plugin): system message ordering + local endpoint apiKey relaxation

redashes1984:
## Changes

This PR fixes two issues in `memOS-local-plugin` core/llm providers that block self-hosted deployments using local inference servers (e.g., vLLM, llama.cpp) with Qwen3.6 and similar models.

---

### Fix 1: System message ordering (HTTP 400 from vLLM)

**File:** `apps/memos-local-plugin/core/llm/client.ts`

**Problem:** When using Qwen3.6 via vLLM, `skill.crystallize` and `reward.r_human` LLM calls fail with:

```
jinja2.exceptions.TemplateError: System message must be at the beginning.
```

Root cause: `normalizeMessages()` passed the input array unchanged, so dynamically constructed messages could place `role: "system"` after user/assistant turns. Qwen3.6\`s Jinja2 chat template enforces system messages at the start.

**Impact:** ~2,300 HTTP 400 errors per day on a single deployment.

**Fix:** `normalizeMessages()` now:
1. Extracts all system messages regardless of position
2. Merges them into a single leading system message
3. Appends all non-system messages in original order

Includes a fast path: if there is exactly one system message already at position 0 with no later systems, the original array is returned unchanged (zero allocation overhead).

---

### Fix 2: apiKey is required even for local endpoints

**File:** `apps/memos-local-plugin/core/llm/providers/openai.ts`

**Problem:** `openai_compatible` provider throws `LLM_UNAVAILABLE: openai_compatible provider requires config.llm.apiKey` even when the endpoint is a local self-hosted inference server (e.g., `http://10.10.4.8:8000/v1`) that does not require authentication.

**Impact:** Users cannot use the `openai_compatible` provider with local vLLM/llama.cpp deployments without setting a dummy apiKey.

**Fix:** Added `isLocalhostOrPrivateUrl()` helper that detects:
- `localhost`, `127.0.0.1`, `::1`
- Private IP ranges: `10.x.x.x`, `172.16-31.x.x`, `192.168.x.x`

When the endpoint is a private/local address, `apiKey` becomes optional. The `Authorization` header is only sent when `apiKey` is configured. Cloud endpoints (e.g., `api.openai.com`) still require `apiKey` as before.

---

## Testing

Verified on a live deployment with Qwen3.6-27B-FP8 via vLLM:

| Metric | Before | After |
|--------|--------|-------|
| HTTP 400 errors (system message ordering) | ~2,300/day | 0 |
| LLM_UNAVAILABLE errors (apiKey) | 100% | 0 |
| skill.crystallize failures | 100% | 0 |
| reward.r_human failures | 100% | 0 |


Memtensor-AI:

**Details:** Executor error: Command failed: git clone --depth 1 --branch fix/system-message-ordering git@github.com:MemTensor/MemOS.git /data/test-workspaces/4168e0a73dc81413/repo
Cloning into '/data/test-workspaces/4168e0a73dc81413/repo'...
warning: Could not find remote branch fix/system-message-ordering to clone.
fatal: Remote branch fix/system-message-ordering not found in upstream origin
**Branch:** `fix/system-message-ordering`

PR #1933, opened 2026-06-16
fix(memos-local): L2 candidate stuck + bridge subprocess leak

a844810597:
## Fix 1: L2 Candidate Policy stuck in candidate

A candidate policy whose support/gain satisfy thresholds but is no longer
cosine-matched by any trace never enters touched and nextStatus()
is never executed. Add Step 5 after Step 4 to scan all candidate policies.

apps/memos-local-plugin/core/memory/l2/l2.ts +42/-1

## Fix 2: Bridge subprocess leak via weakref + __del__

The keepalive thread closure captured self directly, creating a reference
cycle that prevented GC. Combined with missing __del__, leaked providers
leave bridge subprocesses alive indefinitely.

1. weakref — break the reference cycle
2. __del__ — safety net for missed shutdown()

apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py +41/-8

---

No interface changes. Normal paths unaffected.

### DIFF
--- a/apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py
+++ b/apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py
@@ -51,6 +51,7 @@ import logging
 import re
 import sys
 import threading
+import weakref
 import time
 
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


Memtensor-AI:
Cloud AutoDev retest on dev-v2.0.22: PASSED.

Run: tr-150af4de-3fb on test-engine-v2.
Scope: memos_local_plugin focused unit.
Result: 8/8 passed.

PR #1955, opened 2026-06-21
fix(logging): initialize logger from config in the standalone bridge

chiefmojo:
## Problem

The standalone bridge (`bridge.cts`) calls `bootstrapMemoryCoreFull` without passing `initLogging: true`. The logger stays on its bootstrap-console default, ignoring `config.logging.*` entirely. In practice this means `logging.timezone` has no effect when running as a daemon — logs are always stamped UTC regardless of config.

## Fix

Add `initLogging: true` to the `bootstrapMemoryCoreFull` call in `bridge.cts`. The non-daemon stdio path (`bridge/stdio.ts`) already passes this flag correctly; this aligns the daemon path.

```typescript
const { core, config, home } = await bootstrapMemoryCoreFull({
  // ...
  home: resolvedHome,
  initLogging: true,  // initialize logger from config.logging (timezone, level, channels)
});
```

PR #1956, opened 2026-06-21
feat: configurable log timezone

chiefmojo:
## Problem

All log timestamps are hardcoded to UTC via Pino's `formatters.level` path — there's no way for operators in other timezones to get local timestamps without post-processing logs.

## Changes

- Adds `logging.timezone` config field (IANA zone string, default `"UTC"`)
- Logger initializer reads the field and passes it to Pino as `timestamp: () => \`,"time":"${new Date().toLocaleString('en-CA', { timeZone, hour12: false })}"\``
- Schema validation rejects invalid IANA zones with `config_invalid`
- Tests cover UTC default, valid custom zones, and invalid zone rejection

## Config example

```yaml
logging:
  timezone: America/Los_Angeles
```

Depends on fix/daemon-init-logger (#1955) — without `initLogging: true` in the bridge the timezone config is never read.


Memtensor-AI:

- Run: `tr-dd2b4e62-fad` on cloud test-engine `10010`
- memos_local_plugin/unit: 53 passed, 0 failed, 0 skipped

Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #1956
**Task:** `ececbb0aa99b7d72`
**Base:** `dev-v2.0.25`
**Head:** `feat/log-timezone`

---

### 1. `apps/memos-local-plugin/core/time.ts` (L58)

When `opts.offset` is false (the default), the returned string has no timezone indicator — no `Z` and no UTC offset. This makes the output ambiguous: a caller cannot tell whether the value is UTC, a local time, or a zoned time. Consider always appending the offset when a non-UTC `tz` is supplied, or at minimum document that the result is intentionally a "wall clock" string with no zone information so callers are not surprised.

**Before:**
```
  return opts.offset ? base + offsetForTimeZone(d, tz) : base;
```

**After:**
```suggestion
  const offsetStr = offsetForTimeZone(d, tz);
  return opts.offset ? base + offsetStr : base + "[" + tz + "]";
  // or always append the numeric offset:
  // return base + offsetStr;
```

</details>

---

### 2. `apps/memos-local-plugin/core/time.ts` (L57)

`Intl.DateTimeFormat` with `hour12: false` can produce `"24"` for the hour component at midnight in some environments instead of `"00"`, which is not valid ISO 8601. Guard against this edge case.

**Before:**
```
  const base = `${parts.year}-${parts.month}-${parts.day}T${parts.hour}:${parts.minute}:${parts.second}.${parts.fractionalSecond ?? "000"}`;
```

**After:**
```suggestion
  const hour = parts.hour === "24" ? "00" : parts.hour;
  const base = `${parts.year}-${parts.month}-${parts.day}T${hour}:${parts.minute}:${parts.second}.${parts.fractionalSecond ?? "000"}`;
```

</details>

---

### 3. `apps/memos-local-plugin/core/time.ts` (L80-L89)

`timeZoneName: "longOffset"` is a relatively recent addition to the `Intl.DateTimeFormat` spec (ES2022 / Chrome 95+, Node 18+). On older runtimes this silently falls back, and `fmt.formatToParts` may not return a `timeZoneName` part at all, causing `zone` to be `undefined` and the function to return `"+00:00"` for every timezone — a silent wrong answer. Add a comment documenting the minimum runtime requirement, or add an assertion / fallback for when the part is missing.

**Before:**
```
function offsetForTimeZone(d: Date, tz: string): string {
  const fmt = new Intl.DateTimeFormat("en-US", {
    timeZone: tz,
    timeZoneName: "longOffset",
    hour: "2-digit",
  });
  const zone = fmt.formatToParts(d).find((part) => part.type === "timeZoneName")?.value;
  if (!zone || zone === "GMT" || zone === "UTC") return "+00:00";
  return zone.replace(/^GMT/, "");
}
```

**After:**
```suggestion
function offsetForTimeZone(d: Date, tz: string): string {
  // Requires Node ≥18 / Chrome ≥95 for `longOffset` support.
  const fmt = new Intl.DateTimeFormat("en-US", {
    timeZone: tz,
    timeZoneName: "longOffset",
    hour: "2-digit",
  });
  const zone = fmt.formatToParts(d).find((part) => part.type === "timeZoneName")?.value;
  if (!zone || zone === "GMT" || zone === "UTC") return "+00:00";
  // Expected format: "GMT+HH:MM" or "GMT-HH:MM"
  const offset = zone.replace(/^GMT/, "");
  if (!/^[+-]\d{2}:\d{2}$/.test(offset)) {
    throw new Error(`Unexpected timeZoneName format: ${zone}`);
  }
  return offset;
}
```

</details>

Memtensor-AI:

**Branch:** `feat/log-timezone`

PR #2018, opened 2026-07-01

### LINKED ISSUE #1722: fix: memory_search returned error

Memtensor-AI:
## Description

Fixes #1722 (memory_search / turn.end returning `[timeout]` after the Hermes agent has been running for an hour or two) in the Python side of the MemOS Local plugin without touching the Node bridge or the JSON-RPC wire format.

Three cooperating Python defects made a dead / hung bridge subprocess repeatedly stall every memory tool for 30 s per call. The fix has four layers: (1) `MemosBridgeClient._read_loop` now runs in try/except/finally and, on any exit, calls a new `_abort_pending("bridge subprocess exited")` helper that wakes every parked JSON-RPC waiter with `BridgeError("transport_closed", ...)` in under a second; (2) `MemosBridgeClient.request` fast-fails with `transport_closed` when `Popen.poll()` reports the subprocess has exited, so we never write to a broken pipe that the OS would silently buffer; (3) `MemTensorProvider._start_bridge_keepalive` now uses `_should_reconnect_after_keepalive_failure(err)` to reconnect on `transport_closed`, `BridgeError("timeout", ...)`, or any error observed while the subprocess is dead — while explicitly NOT reconnecting on a live subprocess with a transient generic error, to avoid reconnect storms; (4) `MemTensorProvider.handle_tool_call` routes every read-path memory tool (`memos_search`, `memos_get`, `memos_timeline`, `memos_environment`, `memos_skill_list`, `memos_skill_get`) through the new `_bridge_request_with_retry` helper that reconnects + retries exactly once on `transport_closed`, mirroring the write-path pattern already used by `sync_turn`.

Tests: `apps/memos-local-plugin/tests/python/test_bridge_client.py` gains six new cases covering reader-thread abort, request fast-fail, keepalive reconnect on timeout / dead subprocess, the negative case for transient generic errors, and both read-path retry outcomes. Execution results: `python3 -m unittest tests.python.test_bridge_client` → 33 tests OK; `python3 -m unittest tests.python.test_hermes_provider_pipeline` → 16 tests OK; `python3 -m ruff check adapters/hermes/memos_provider/ tests/python/` → all clean; `python3 -m ruff format --check adapters/hermes/memos_provider/ tests/python/` → 5 files already formatted.

Committed as cc352cc7 on branch bugfix/autodev-1722 (base dev-20260624-v2.0.22); working branch pushed to origin, ready for the scheduler to open the PR. Specs archive (task.md + proposal / spec / design / verification-report) pushed to memos-autodev-specs main under 2026-07-01-1722-fix-memorysearch-returned-error/.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Automated tests are pending.

- [ ] Unit Test

## Checklist

## Reviewer Checklist
- [x] closes #1722
- [ ] Made sure Checks passed

### DIFF
--- a/apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py
+++ b/apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py
@@ -1280,7 +1280,7 @@ class MemTensorProvider(MemoryProvider):
                 }
                 if bool(args.get("sessionScope", False)):
                     params["sessionId"] = self._session_id
-                resp = self._bridge.request(
+                resp = self._bridge_request_with_retry(
                     "memory.search",
                     params,
                 )
@@ -1298,7 +1298,7 @@ class MemTensorProvider(MemoryProvider):
                 method = methods.get(kind)
                 if method is None:
                     return json.dumps({"error": f"unknown memory kind: {kind}"})
-                item = self._bridge.request(
+                item = self._bridge_request_with_retry(
                     method, {"id": item_id, "namespace": self._runtime_namespace()}
                 )
                 if not item:
@@ -1343,7 +1343,7 @@ class MemTensorProvider(MemoryProvider):
                     }
                 )
             if tool_name == "memos_timeline":
-                resp = self._bridge.request(
+                resp = self._bridge_request_with_retry(
                     "memory.timeline",
                     {
                         "episodeId": args.get("episodeId", self._episode_id),
@@ -1358,12 +1358,12 @@ class MemTensorProvider(MemoryProvider):
                 params = {"limit": limit, "namespace": self._runtime_namespace()}
                 if args.get("status"):
                     params["status"] = args["status"]
-                return json.dumps(self._bridge.request("skill.list", params))
+                return json.dumps(self._bridge_request_with_retry("skill.list", params))
             if tool_name == "memos_environment":
                 query = (args.get("query") or "").strip()
                 limit = self._int_arg(args, "limit", 5, 1, 30)
                 if not query:
-                    resp = self._bridge.request(
+                    resp = self._bridge_request_with_retry(
                         "memory.list_world_models",
                         {"limit": limit, "offset": 0, "namespace": self._runtime_namespace()},
                     )
@@ -1379,7 +1379,7 @@ class MemTensorProvider(MemoryProvider):
                             "queried": False,
                         }
                     )
-                resp = self._bridge.request(
+                resp = self._bridge_request_with_retry(
                     "memory.search",
                     {
                         "agent": "hermes",
@@ -1412,7 +1412,7 @@ class MemTensorProvider(MemoryProvider):
                 skill_id = (args.get("id") or "").strip()
                 if not skill_id:
                     return json.dumps({"error": "missing id"})
-                skill = self._bridge.request(
+                skill = self._bridge_request_with_retry(
                     "skill.get",
                     {
                         "id": skill_id,
@@ -1724,6 +1724,40 @@ class MemTensorProvider(MemoryProvider):
         msg = str(err).lower()
         return "broken pipe" in msg or "bridge closed" in msg or "transport_closed" in msg
 
+    def _should_reconnect_after_keepalive_failure(self, err: Exception) -> bool:
+        """Return True when a keepalive-detected failure should trigger reconnect.
+
+        Reconnect when any of the following holds — a superset of the
+        original "transport_closed only" gate:
+
+        1. ``transport_closed`` (existing behaviour: broken pipe /
+           subprocess close observed by the reader thread).
+        2. ``BridgeError("timeout", …)`` — the exact code raised when
+           the bridge stops answering. This is the primary #1722 signal:
+           the subprocess is alive but hung, so every 5 s keepalive
+           health-check times out and we would otherwise never recover.
+        3. The subprocess is dead: ``self._bridge._proc.poll()`` is
+           non-``None`` even if the error itself is generic (defends
+           against edge cases where a crash surfaces as a shape-parse
+           exception rather than a transport signal).
+        """
+        if self._is_transport_closed(err):
+            return True
+        if isinstance(err, BridgeError) and err.code == "timeout":
+            return True
+        bridge = self._bridge
+        if bridge is not None:
+            proc = getattr(bridge, "_proc", None)
+            try:
+                if proc is not None and proc.poll() is not None:
+                    return True
+            except Exception:
+                # Defensive: if poll() itself raises, do NOT reconnect —
+                # we cannot prove the subprocess is dead and reconnecting
+                # on every ambiguous error would create a storm.
+                return False
+        return False
+
     def _reconnect_bridge(self, session_id: str = "", *, timeout: float = 30.0) -> None:
         # Don't reconnect if we're shutting down
         if self._bridge_keepalive_stop.is_set():
@@ -1785,6 +1819,48 @@ class MemTensorProvider(MemoryProvider):
             self._bridge = None
             return False
 
+    def _bridge_request_with_retry(
+        self,
+        method: str,
+        params: dict[str, Any] | None = None,
+        *,
+        timeout: float | None = None,
+    ) -> dict[str, Any]:
+        """Issue a bridge request, reconnecting + retrying once on stale pipe.
+
+        Mirrors the write-path pattern already used by :meth:`sync_turn`.
+        Read-path memory tools (``memos_search``, ``memos_get``,
+        ``memos_timeline``, ``memos_environment``, ``memos_skill_list``,
+        ``memos_skill_get``) route through here so a hung / dead bridge
+        no longer surfaces a bare ``transport_closed`` on the *first*
+        user call after the keepalive tears it down.
+
+        Only ``transport_closed`` is retried — every other error
+        (``timeout``, ``internal``, …) is surfaced verbatim so callers
+        can decide how to react. Only a single retry is attempted; a
+        second failure raises so the tool response contains the error
+        text unchanged.
+        """
+        if self._bridge is None:
+            raise BridgeError("transport_closed", "bridge client is closed")
+        kwargs: dict[str, Any] = {}
+        if timeout is not None:
+            kwargs["timeout"] = timeout
+        try:
+            return self._bridge.request(method, params, **kwargs) or {}
+        except BridgeError as err:
+            if not self._is_transport_closed(err):
+                raise
+            logger.warning(
+                "MemOS: bridge transport closed during %s; reconnecting and retrying once — %s",
+                method,
+                err,
+            )
+            self._reconnect_bridge(self._session_id, timeout=30.0)
+            if self._bridge is None:
+                raise
+            return self._bridge.request(method, params, **kwargs) or {}
+
     def _start_bridge_keepalive(self) -> None:
         if self._bridge_keepalive_thread and self._bridge_keepalive_thread.is_alive():
             return
@@ -1798,8 +1874,12 @@ class MemTensorProvider(MemoryProvider):
                     assert self._bridge is not None
                     self._bridge.request("core.health", {}, timeout=10.0)
                 except Exception as err:
-                    if self._is_transport_closed(err):
-                        logger.info("MemOS: bridge keepalive reconnecting after transport close")
+                    if self._should_reconnect_after_keepalive_failure(err):
+                        logger.info(
+                            "MemOS: bridge keepalive reconnecting — %s: %s",
+                            err.__class__.__name__,
+                            err,
+                        )
                         with contextlib.suppress(Exception):
                             self._reconnect_bridge(self._session_id, timeout=10.0)
                     else:
--- a/apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py
+++ b/apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py
@@ -172,6 +172,18 @@ class MemosBridgeClient:
     ) -> dict[str, Any]:
         if self._closed:
             raise BridgeError("transport_closed", "bridge client is closed")
+        # Fast-fail path: if the subprocess is already gone, stdin.write
+        # would still succeed thanks to OS pipe buffering, and the
+        # caller would then park on ``waiter.wait(timeout)`` for the
+        # full 30 s before failing. Detecting the exit here surfaces
+        # the correct code (``transport_closed``) immediately, which
+        # ``_bridge_request_with_retry`` uses to trigger recovery.
+        exit_code = self._proc.poll()
+        if exit_code is not None:
+            raise BridgeError(
+                "transport_closed",
+                f"bridge subprocess exited (code={exit_code})",
+            )
         with self._lock:
             rpc_id = self._next_id
             self._next_id += 1
@@ -274,87 +286,117 @@ class MemosBridgeClient:
                 except subprocess.TimeoutExpired:
                     logger.error("MemOS: bridge process %d could not be killed", pid)
 
-        # 5. Clean up pending requests
-        with self._lock:
-            for entry in list(self._pending.values()):
-                entry["error"] = {
-                    "code": -32000,
-                    "message": "bridge closed",
-                    "data": {"code": "transport_closed"},
-                }
-                entry["event"].set()
-            self._pending.clear()
+        # 5. Clean up pending requests. Shares the abort path with the
+        # reader thread so any callers still parked on a waiter are
+        # released with ``transport_closed`` from a single code path.
+        self._abort_pending("bridge closed")
 
     # ─── Internals ──
 
     def _read_loop(self) -> None:
         assert self._proc.stdout is not None
-        for line in self._proc.stdout:
-            line = line.strip()
-            if not line:
-                continue
-            try:
-                msg = json.loads(line)
-            except json.JSONDecodeError:
-                logger.debug("bridge: malformed line: %r", line[:120])
-                continue
-            if "id" in msg and msg["id"] is not None and ("result" in msg or "error" in msg):
-                self._resolve(msg)
-                continue
-            if msg.get("method") == "events.notify":
-                for cb in list(self._events):
-                    try:
-                        cb(msg.get("params") or {})
-                    except Exception:
-                        logger.debug("event listener threw", exc_info=True)
-                continue
-            if msg.get("method") == "logs.forward":
-                for cb in list(self._logs):
-                    try:
-                        cb(msg.get("params") or {})
-                    except Exception:
-                        logger.debug("log listener threw", exc_info=True)
-                continue
-            # Reverse-direction request: the bridge is asking the
-            # adapter to do something (e.g. run a fallback LLM call
-            # via `host.llm.complete`). Dispatch to the registered
-            # handler and write the response back synchronously.
-            method = msg.get("method")
-            rpc_id = msg.get("id")
-            if (
-                isinstance(method, str)
-                and rpc_id is not None
-                and "result" not in msg
-                and "error" not in msg
-            ):
-                handler = self._host_handler_for(method)
-                if handler is None:
-                    self._send_response(
-                        rpc_id,
-                        error={
-                            "code": -32601,
-                            "message": f"method not found: {method}",
-                            "data": {"code": "unknown_method"},
-                        },
-                    )
+        try:
+            for line in self._proc.stdout:
+                line = line.strip()
+                if not line:
                     continue
-                params = msg.get("params") or {}
-                if not isinstance(params, dict):
-                    params = {}
                 try:
-                    result = handler(params)
-                    self._send_response(rpc_id, result=result)
-                except Exception as err:
-                    logger.warning("host handler %s failed: %s", method, err)
-                    self._send_response(
-                        rpc_id,
-                        error={
-                            "code": -32000,
-                            "message": str(err) or err.__class__.__name__,
-                            "data": {"code": "host_handler_failed"},
-                        },
-                    )
-                continue
+                    msg = json.loads(line)
+                except json.JSONDecodeError:
+                    logger.debug("bridge: malformed line: %r", line[:120])
+                    continue
+                if "id" in msg and msg["id"] is not None and ("result" in msg or "error" in msg):
+                    self._resolve(msg)
+                    continue
+                if msg.get("method") == "events.notify":
+                    for cb in list(self._events):
+                        try:
+                            cb(msg.get("params") or {})
+                        except Exception:
+                            logger.debug("event listener threw", exc_info=True)
+                    continue
+                if msg.get("method") == "logs.forward":
+                    for cb in list(self._logs):
+                        try:
+                            cb(msg.get("params") or {})
+                        except Exception:
+                            logger.debug("log listener threw", exc_info=True)
+                    continue
+                # Reverse-direction request: the bridge is asking the
+                # adapter to do something (e.g. run a fallback LLM call
+                # via `host.llm.complete`). Dispatch to the registered
+                # handler and write the response back synchronously.
+                method = msg.get("method")
+                rpc_id = msg.get("id")
+                if (
+                    isinstance(method, str)
+                    and rpc_id is not None
+                    and "result" not in msg
+                    and "error" not in msg
+                ):
+                    handler = self._host_handler_for(method)
+                    if handler is None:
+                        self._send_response(
+                            rpc_id,
+                            error={
+                                "code": -32601,
+                                "message": f"method not found: {method}",
+                                "data": {"code": "unknown_method"},
+                            },
+                        )
+                        continue
+                    params = msg.get("params") or {}
+                    if not isinstance(params, dict):
+                        params = {}
+                    try:
+                        result = handler(params)
+                        self._send_response(rpc_id, result=result)
+                    except Exception as err:
+                        logger.warning("host handler %s failed: %s", method, err)
+                        self._send_response(
+                            rpc_id,
+                            error={
+                                "code": -32000,
+                                "message": str(err) or err.__class__.__name__,
+                                "data": {"code": "host_handler_failed"},
+                            },
+                        )
+                    continue
+        except Exception:
+            # Any unexpected exception on the reader thread (e.g. the
+            # subprocess died mid-write and raised OSError inside the
+            # iterator) is logged and treated as EOF — the finally
+            # block below unblocks every parked waiter.
+            logger.debug("bridge reader thread crashed", exc_info=True)
+        finally:
+            # The reader thread has stopped consuming responses. Any
+            # pending JSON-RPC waiter would otherwise sleep for the
+            # full per-request timeout. Wake them all up with the
+            # correct code so callers can trigger reconnect + retry.
+            self._abort_pending("bridge subprocess exited")
+
+    def _abort_pending(self, message: str) -> None:
+        """Release every pending waiter with ``transport_closed``.
+
+        Called from the reader thread on EOF / exception, so any
+        request that had already sent its bytes but never received a
+        response is woken up immediately instead of parking on its
+        per-request timeout. ``close()`` also delegates here so the
+        cleanup is defined in exactly one place.
+        """
+        with self._host_handlers_cv:
+            self._closed = True
+            self._host_handlers_cv.notify_all()
+        with self._lock:
+            pending = list(self._pending.values())
+            self._pending.clear()
+        for entry in pending:
+            entry["error"] = {
+                "code": -32000,
+                "message": message,
+                "data": {"code": "transport_closed"},
+            }
+            entry["event"].set()
 
     def _host_handler_for(
         self,
--- a/apps/memos-local-plugin/tests/python/test_bridge_client.py
+++ b/apps/memos-local-plugin/tests/python/test_bridge_client.py
@@ -68,6 +68,12 @@ class FakePopen:
     def kill(self) -> None:
         pass
 
+    # ``request()`` short-circuits on a dead subprocess by consulting
+    # ``poll()``. The default fake process is "alive" (returns None); a
+    # specific test can monkeypatch this to simulate an exit.
+    def poll(self) -> int | None:
+        return None
+
 
 class _ServerStream(io.StringIO):
     """Script bridge responses as if coming from the Node subprocess."""
@@ -342,6 +348,68 @@ class BridgeClientTests(unittest.TestCase):
                     return msg
             time.sleep(0.01)
         self.fail("timed out waiting for client write")
+
+    def test_reader_exit_marks_pending_as_transport_closed(self) -> None:
+        """When the reader loop exits (subprocess died), pending waiters
+        must be released immediately with ``transport_closed`` instead of
+        parking on the caller's ``timeout``."""
+        client = MemosBridgeClient(bridge_path="/tmp/bridge.cts")
+        assert self._fake is not None
+
+        # Enqueue a request the server will NEVER answer, then simulate
+        # subprocess exit by draining the stdout stream and marking it
+        # done. The reader thread must observe EOF and abort every
+        # pending request.
+        results: dict[str, Exception | None] = {"err": None}
+
+        def _call() -> None:
+            try:
+                client.request("core.health", timeout=30.0)
+            except BridgeError as err:
+                results["err"] = err
+
+        # Craft a request whose response the fake server won't produce.
+        # We do that by temporarily swapping the fake's on_request so no
+        # answer is enqueued.
+        self._fake.stdout.on_request = lambda _raw: None  # type: ignore[assignment]
+
+        thread = threading.Thread(target=_call, daemon=True)
+        thread.start()
+
+        # Give the writer + reader a moment to register the pending id.
+        time.sleep(0.05)
+
+        # Simulate subprocess exit: no more bytes ever, iterator returns.
+        self._fake.stdout._done = True
+        # Nudge the reader thread out of its 50 ms wait.
+        self._fake.stdout._event.set()
+
+        thread.join(timeout=2.0)
+        self.assertFalse(thread.is_alive(), "waiter never released after EOF")
+
+        err = results["err"]
+        self.assertIsInstance(err, BridgeError)
+        assert isinstance(err, BridgeError)
+        self.assertEqual(err.code, "transport_closed")
+        client.close()
+
+    def test_request_fast_fails_when_subprocess_already_dead(self) -> None:
+        """A dead subprocess (``poll()`` non-None) must raise
+        ``transport_closed`` before we write to a broken pipe."""
+        client = MemosBridgeClient(bridge_path="/tmp/bridge.cts")
+        assert self._fake is not None
+
+        # Simulate subprocess exit (exit code 137 = OOM kill).
+        self._fake.poll = lambda: 137  # type: ignore[assignment]
+
+        with self.assertRaises(BridgeError) as ctx:
+            client.request("core.health", timeout=1.0)
+        self.assertEqual(ctx.exception.code, "transport_closed")
+        self.assertIn("137", str(ctx.exception))
+        # We must not have written anything to the dead stdin pipe.
+        stdin_writes = [line for line in self._fake._stdin_lines if line]
+        self.assertEqual(stdin_writes, [])
+        client.close()
 
 
 class MemTensorProviderTests(unittest.TestCase):
@@ -626,6 +694,167 @@ class MemTensorProviderTests(unittest.TestCase):
         self.assertIn("中东", retry_payload["userText"])
         self.assertIn("局势", retry_payload["agentText"])
         self.assertEqual(retry_payload["toolCalls"][0]["name"], "search_files")
+
+    # ─── Regression tests for #1722 ──────────────────────────────────
+    #
+    # A hung / dead bridge subprocess used to make every memory tool
+    # block for the full 30 s per-request timeout without ever
+    # recovering. The three layers below cover:
+    #
+    #   1. keepalive detects a hung bridge (timeout / dead subprocess)
+    #      and triggers a reconnect;
+    #   2. keepalive stays quiet for a live subprocess with a transient
+    #      parse error (no reconnect storms);
+    #   3. read-path tools reconnect and retry once on a stale bridge,
+    #      mirroring the write-path behaviour already covered above.
+
+    def test_keepalive_reconnects_on_health_timeout(self) -> None:
+        """``BridgeError('timeout', …)`` must count as reconnect-worthy."""
+
+        class TimingOutBridge:
+            def __init__(self) -> None:
+                self.calls: list[str] = []
+
+            def close(self) -> None:
+                pass
+
+            def request(self, method: str, params: dict | None = None, **_kw) -> dict:
+                self.calls.append(method)
+                if method == "core.health":
+                    raise BridgeError("timeout", "core.health did not respond within 10.0s")
+                return {}
+
+        p = self._provider_mod.MemTensorProvider()
+        # The helper is the sole decision point the keepalive consults;
+        # exercising it directly keeps this test hermetic.
+        err = BridgeError("timeout", "core.health did not respond within 10.0s")
+        p._bridge = TimingOutBridge()
+        self.assertTrue(p._should_reconnect_after_keepalive_failure(err))
+
+    def test_keepalive_reconnects_when_subprocess_is_dead(self) -> None:
+        """Any error whose subprocess reports ``poll() != None`` triggers reconnect."""
+
+        class DeadBridge:
+            def __init__(self) -> None:
+                class _Proc:
+                    def poll(self_inner) -> int:  # noqa: N805
+                        return 1
+
+                self._proc = _Proc()
+
+            def close(self) -> None:
+                pass
+
+            def request(self, method: str, params: dict | None = None, **_kw) -> dict:
+                raise BridgeError("internal", "shape parse failed")
+
+        p = self._provider_mod.MemTensorProvider()
+        p._bridge = DeadBridge()
+        err = BridgeError("internal", "shape parse failed")
+        self.assertTrue(p._should_reconnect_after_keepalive_failure(err))
+
+    def test_keepalive_does_not_reconnect_on_transient_generic_error(self) -> None:
+        """A live subprocess + generic error must NOT trigger reconnect
+        (avoids reconnect storms on transient parse noise)."""
+
+        class LiveNoisyBridge:
+            def __init__(self) -> None:
+                class _Proc:
+                    def poll(self_inner) -> None:  # noqa: N805
+                        return None
+
+                self._proc = _Proc()
+
+            def close(self) -> None:
+                pass
+
+        p = self._provider_mod.MemTensorProvider()
+        p._bridge = LiveNoisyBridge()
+        err = BridgeError("internal", "malformed response")
+        self.assertFalse(p._should_reconnect_after_keepalive_failure(err))
+
+    def test_handle_tool_call_retries_memos_search_after_transport_closed(self) -> None:
+        """Read-path memory tools must reconnect + retry once on stale bridge."""
+
+        class StaleBridge:
+            def __init__(self) -> None:
+                self.calls: list[str] = []
+
+            def close(self) -> None:
+                pass
+
+            def request(self, method: str, params: dict | None = None, **_kw) -> dict:
+                self.calls.append(method)
+                raise BridgeError("transport_closed", "[Errno 32] Broken pipe")
+
+        class HealthyBridge:
+            def __init__(self) -> None:
+                self.calls: list[tuple[str, dict]] = []
+
+            def register_host_handler(self, *_a, **_k) -> None:
+                return None
+
+            def request(self, method: str, params: dict | None = None, **_kw) -> dict:
+                payload = params or {}
+                self.calls.append((method, payload))
+                if method == "session.open":
+                    return {"sessionId": payload.get("sessionId", "sess")}
+                if method == "memory.search":
+                    return {"hits": [{"refId": "tr-after", "score": 0.9}]}
+                return {}
+
+        stale = StaleBridge()
+        healthy = HealthyBridge()
+        p = self._provider_mod.MemTensorProvider()
+        p._bridge = stale
+        p._session_id = "sess_search"
+        p._episode_id = "ep_search"
+
+        with patch("memos_provider.MemosBridgeClient", return_value=healthy):
+            raw = p.handle_tool_call("memos_search", {"query": "yesterday"})
+
+        parsed = json.loads(raw)
+        self.assertNotIn("error", parsed)
+        self.assertEqual(parsed["hits"][0]["refId"], "tr-after")
+        # First call went to the stale bridge; the retry hit the healthy one.
+        self.assertEqual(stale.calls, ["memory.search"])
+        self.assertIn("memory.search", [m for m, _ in healthy.calls])
+
+    def test_handle_tool_call_surfaces_second_transport_failure(self) -> None:
+        """If the retry ALSO fails, the tool must surface an error verbatim."""
+
+        class AlwaysStaleBridge:
+            def __init__(self) -> None:
+                self.calls = 0
+
+            def close(self) -> None:
+                pass
+
+            def request(self, method: str, params: dict | None = None, **_kw) -> dict:
+                self.calls += 1
+                raise BridgeError("transport_closed", "[Errno 32] Broken pipe")
+
+        class ReplacementBridge:
+            def register_host_handler(self, *_a, **_k) -> None:
+                return None
+
+            def request(self, method: str, params: dict | None = None, **_kw) -> dict:
+                if method == "session.open":
+                    return {"sessionId": (params or {}).get("sessionId", "sess")}
+                # Retry still hits transport_closed.
+                raise BridgeError("transport_closed", "[Errno 32] Broken pipe again")
+
+        p = self._provider_mod.MemTensorProvider()
+        p._bridge = AlwaysStaleBridge()
+        p._session_id = "sess_search"
+        p._episode_id = "ep_search"
+
+        with patch("memos_provider.MemosBridgeClient", return_value=ReplacementBridge()):
+            raw = p.handle_tool_call("memos_search", {"query": "yesterday"})
+
+        parsed = json.loads(raw)
+        self.assertIn("error", parsed)
+        self.assertIn("Broken pipe", parsed["error"])
 
     def test_get_config_schema_describes_known_fields(self) -> None:
         p = self._provider_mod.MemTensorProvider()


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2018
**Task:** `ff6c49378bec35ae`

---

### 1. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L252-L257)

**Race condition: unsynchronized `self._closed` check in `close()`**

`close()` reads `self._closed` (line 253) without holding any lock, while `_abort_pending` (called from the reader thread's `finally` block) writes `self._closed = True` under `_host_handlers_cv`. Two concurrent callers of `close()` can both pass the guard before either sets the flag, resulting in a double-close that calls `terminate()`/`kill()` on a subprocess that is already being shut down.

Suggested fix: perform the idempotency check inside the same `_host_handlers_cv` lock that protects the assignment:

```python
def close(self) -> None:
    with self._host_handlers_cv:
        if self._closed:
            return
        self._closed = True
        self._host_handlers_cv.notify_all()
    # ... rest of shutdown ...
```

---

### 2. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L387-L399)

**`_abort_pending` called from reader thread's `finally` block conflicts with `close()`'s own call to `_abort_pending`**

When the bridge subprocess exits gracefully (i.e., triggered by `close()` closing stdin), the event sequence is:
1. `close()` sets `self._closed = True` and closes stdin.
2. The reader thread observes EOF, its `finally` block calls `_abort_pending("bridge subprocess exited")`, which sets `self._closed = True` again and runs the pending-waiter cleanup.
3. `close()` then calls `_abort_pending("bridge closed")` a second time.

While the second call is functionally a no-op (pending dict is already empty), the `_host_handlers_cv.notify_all()` inside `_abort_pending` fires again unnecessarily and any code that inspects the `message` passed to `_abort_pending` (e.g. future logging improvements) may see the wrong message. Consider guarding `_abort_pending` with an internal idempotency flag (`_aborted`) distinct from `_closed`, or skip the `_host_handlers_cv` block in `_abort_pending` if `_closed` is already `True`.

---

### 3. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L365-L370)

**Unexpected reader-thread exceptions logged at `debug` level — will be invisible in production**

An OSError or other unexpected exception on the reader thread (e.g. the OS killed the subprocess with SIGKILL mid-read) is silently swallowed at `DEBUG` level. In production environments where the default log level is `WARNING` or `INFO`, this means a crashing bridge will leave no trace in logs, making repeated reconnect cycles very hard to diagnose.

Consider using `logger.warning` or `logger.error` here, consistent with how other subprocess failures are logged (e.g. SIGTERM/SIGKILL paths use `logger.warning` / `logger.error`).

---

### 4. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L379-L380)

Timing-based synchronization risk: the `time.sleep(0.05)` is used to ensure that `client.request()` has progressed far enough to register the pending waiter in `_pending` before EOF is injected. On a heavily loaded CI runner or slow VM, 50 ms may not be sufficient for the request thread to acquire `_lock`, insert into `_pending`, write to stdin, and release the lock — causing the reader's `_abort_pending` to fire before the entry exists, which means the waiter would then block for the full 30 s `timeout` and the `thread.join(timeout=2.0)` assertion would fail non-deterministically.

Consider replacing the fixed sleep with an active poll on the observable bridge state, e.g.:
```python
deadline = time.monotonic() + 2.0
while time.monotonic() < deadline:
    with client._lock:
        if client._pending:
            break
    time.sleep(0.005)
else:
    self.fail("request never registered in _pending")
```
This makes the synchronization deterministic and self-documenting.

---

### 5. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1844-L1862)

Thread-safety race condition: `self._bridge` is read and used without holding `_reconnect_lock`, while the keepalive thread calls `_reconnect_bridge` (which sets `self._bridge = None` during teardown) concurrently. Two specific windows exist:
1. Between the `is None` guard (first line) and the `self._bridge.request(...)` call below it — the keepalive thread could null out the bridge after the check passes, causing `AttributeError: 'NoneType' object has no attribute 'request'`.
2. Between the post-reconnect `if self._bridge is None:` check (second-to-last line) and the retry call on the last line — same race.

Suggestion: snapshot the bridge into a local variable to avoid the race:
```python
bridge = self._bridge
if bridge is None:
    raise BridgeError("transport_closed", "bridge client is closed")
try:
    return bridge.request(method, params, **kwargs) or {}
except BridgeError as err:
    ...
    bridge = self._bridge
    if bridge is None:
        raise
    return bridge.request(method, params, **kwargs) or {}
```

---

### 6. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1838-L1842)

Incomplete fix for issue #1722 (hung bridge): The keepalive thread now correctly reconnects on `BridgeError("timeout")` via `_should_reconnect_after_keepalive_failure`, but `_bridge_request_with_retry` only retries on `transport_closed`. In the primary #1722 scenario — bridge subprocess is alive but hung — a user tool call that hits the hung bridge will receive a `timeout` error that is NOT retried here. The user sees the error even though reconnect would succeed.

The docstring acknowledges this asymmetry ("callers can decide how to react"), but given the fix goal, consider also retrying on `BridgeError(code="timeout")` the same way `_should_reconnect_after_keepalive_failure` does, or at minimum document why the asymmetry is intentional from the user-facing perspective.

---

### 7. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L382-L385)

The `_done` flag is set on the `_ServerStream` instance but `_ServerStream.__iter__` checks `hasattr(self, '_done') and self._done` only after the 50ms `_event.wait` timeout elapses. By setting `_event` immediately after `_done = True`, the iterator should exit on the next loop iteration without waiting the full 50ms. However, this double-step setup (`_done = True` then `_event.set()`) is slightly fragile: if the reader thread is mid-iteration between reading a line and reaching the `_event.wait()` call, it will loop once more before checking `_done`. This is harmless in practice (one extra 0ms spin), but documenting that both steps are intentionally required would improve maintainability.

---

### 8. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L379-L380)

Timing-based synchronization risk: the `time.sleep(0.05)` is used to ensure that `client.request()` has progressed far enough to register the pending waiter in `_pending` before EOF is injected. On a heavily loaded CI runner or slow VM, 50 ms may not be sufficient for the request thread to acquire `_lock`, insert into `_pending`, write to stdin, and release the lock. If `_abort_pending` fires before the entry exists, the waiter then blocks for the full 30s `timeout`, causing the subsequent `thread.join(timeout=2.0)` assertion to fail non-deterministically.

Consider replacing the fixed sleep with an active poll on the observable bridge state:
```python
deadline = time.monotonic() + 2.0
while time.monotonic() < deadline:
    with client._lock:
        if client._pending:
            break
    time.sleep(0.005)
else:
    self.fail("request never registered in _pending")
```
This makes the synchronization deterministic and self-documenting.

---

### 9. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L379-L380)

Timing-based synchronization risk: the `time.sleep(0.05)` is used to ensure `client.request()` has registered the pending waiter in `_pending` before EOF is injected. On a heavily loaded CI runner, 50ms may not be sufficient for the request thread to acquire `_lock`, insert into `_pending`, write to stdin, and release the lock. If `_abort_pending` fires before the entry exists, the waiter blocks for the full 30s `timeout`, causing the `thread.join(timeout=2.0)` assertion to fail non-deterministically.

Consider replacing the fixed sleep with an active poll on the observable bridge state:
```python
deadline = time.monotonic() + 2.0
while time.monotonic() < deadline:
    with client._lock:
        if client._pending:
            break
    time.sleep(0.005)
else:
    self.fail("request never registered in _pending")
```
This makes the synchronization deterministic and self-documenting.

---

### 10. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L379-L380)

Flaky
-test
 risk
:
 
`time
.sleep
(0
.05
)
`
 is
 used
 to
 guarantee
 the
 
`_call
`
 thread
 has
 reached
 
`waiter
.wait
(
)
`
 (i
.e
.
 the
 RPC
 id
 is
 registered
 in
 
`client
._pending
`
) before
 EOF
 is
 injected
 via
 
`_done
 =
 True
`
.
 If
 that
 ordering
 is
 not
 met
,
 
`_abort_pending
`
 fires
 with
 an
 empty
 
`_pending
`
 dict
,
 and
 then
 the
 request
 thread
 adds
 its
 waiter
 *after
 the
 abort
*
 —
 leaving
 it
 to
 block
 for
 the
 full
 30s
 
`timeout
`
.
 The
 
`thread
.join
(timeout
=2
.0
)
`
 assertion
 below
 then
 fails
 non
-deterministically
 on
 slow
 CI
 runners
.

Replace
 the
 fixed
 sleep
 with
 a
 condition
-based
 poll
 that
 spins
 on
 the
 observable
 bridge
 state
:
```python
deadline = time.monotonic() + 2.0
while time.monotonic() < deadline:
    with client._lock:
        if client._pending:
            break
    time.sleep(0.005)
else:
    self.fail("request never registered in _pending")
```
This
 makes
 the
 synchronization
 deterministic
 and
 self
-documenting
 without
 relying
 on
 wall
-clock
 assumptions
.

**Before:**
```
        # Give the writer + reader a moment to register the pending id.
        time.sleep(0.05)
```

**After:**
```suggestion
        # Wait until the request is actually registered in _pending,
        # making the synchronization deterministic instead of relying on
        # a fixed wall-clock sleep that can be too short on loaded CI runners.
        deadline = time.monotonic() + 2.0
        while time.monotonic() < deadline:
            with client._lock:
                if client._pending:
                    break
            time.sleep(0.005)
        else:
            self.fail("request never registered in _pending")
```

</details>

---

### 11. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L806-L807)

Missing
 assertion
 that
 
`close
(
)
`
 was
 called
 on
 the
 stale
 bridge
 before
 replacement
.
 The
 production
 
`
_reconnect
_bridge
`
 method
 calls
 
`old
_bridge
.close
(
)
`
,
 but
 this
 test
 doesn
't
 verify
 it
.
 If
 that
 call
 were
 accidentally
 removed
 from
 
`
_reconnect
_bridge
`
,
 the
 test
 would
 still
 pass
 while
 a
 resource
 leak
 regression
 goes
 undetected
.
 Consider
 adding
:

`
`
`python
self
.assertTrue
(
 stale
.was
_closed
,
 
"stale
 bridge
.
 close
(
)
 must
 be
 called
 before
 reconnect
"
)

`
`
`
where
 
`StaleBridge
.close
(
)
`
 sets
 a
 
`was
_closed
`
 flag
.

---

### 12. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L751-L754)

This

 test

 verifies

 

`
_should
_reconnect
_after
_keepalive
_failure
`

 in

 isolation

,

 which

 is

 fine

 as

 a

 unit

 test

.

 However

,

 the

 method

 under

 test

 accesses

 

`self

.
_bridge

.
_proc

.poll

(

)

`

 via

 

`getattr

`

 with

 a

 

`None

`

 fallback

,

 which

 means

 a

 typo

 in

 the

 attribute

 name

 (e

.g

.

 

`_process

`

 instead

 of

 

`_proc

`

)

 would

 silently

 make

 

`proc

`

 resolve

 to

 

`None

`

 and

 return

 

`False

`

,

 causing

 this

 test

 to

 fail

 for

 the

 wrong

 reason

.

 Adding

 an

 explicit

 assertion

 that

 

`DeadBridge

.
_proc

.poll

(

)

 

==

 1

`

 is

 called

 (e

.g

.

 using

 a

 

`MagicMock

`

 for

 

`_proc

`

)

 would

 make

 the

 test

 verify

 the

 full

 call

 chain

 rather

 than

 just

 the

 final

 boolean

.

Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2018
**Task:** `ff6c49378bec35ae`

🔍 **OpenCodeReview** found **15** issue(s) in this PR.

---

### 1. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L451-L462)

**[High] TOCTOU gap between `_host_handlers_cv` and `_lock` can orphan a pending entry.**

`_abort_pending` releases `_host_handlers_cv` (with `_aborted=True` and `_closed=True` set) and *then* acquires `_lock` to drain `_pending`. Between those two acquisitions a concurrent `request()` call can:
1. Pass the `_closed` check (before `_aborted` was set, or just before it reads `_closed`).
2. Pass the `poll()` check.
3. Acquire `_lock`, insert its entry into `_pending`.
4. Release `_lock`.

Then `_abort_pending` acquires `_lock`, snapshots `entries = list(self._pending.values())` (which **includes** the new entry), calls `self._pending.clear()`, and sets the event. So in this specific race the waiter *will* be woken — that part is fine.

However there is a second, subtler race: `request()` inserts the entry **after** `_abort_pending` has already cleared `_pending` (i.e., between `self._pending.clear()` and `entry["event"].set()` or after `_abort_pending` has fully returned). In that case the orphaned entry is never woken and must wait the full `timeout`. The two-lock design (checking `_closed` under `_lock` and draining `_pending` under the same `_lock`) is almost right, but the gap still exists because `_aborted`/`_closed` are set under a *different* lock (`_host_handlers_cv`).

**Fix**: hold `_lock` for the entire drain-and-close operation, or — since `_lock` already guards `_pending` and `request()` re-checks `_closed` under `_lock` — move the `_aborted`/`_closed` flip inside the `_lock` block:
```python
def _abort_pending(self, error: dict[str, Any]) -> None:
    with self._lock:
        if self._aborted:
            return
        self._aborted = True
        self._closed = True
        entries = list(self._pending.values())
        self._pending.clear()
    # Wake host-handler waiters separately — no longer under _lock.
    with self._host_handlers_cv:
        self._host_handlers_cv.notify_all()
    for entry in entries:
        entry["error"] = dict(error)
        entry["event"].set()
```
With this ordering, any `request()` that acquires `_lock` after `_abort_pending` has set `_closed` will bail out immediately; any that already holds `_lock` when `_abort_pending` tries to acquire it will insert its entry before the clear, and that entry will be drained.

---

### 2. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L327-L340)

**[High] Exception in `close()` skips `_abort_pending`, leaving pending waiters parked until timeout.**

`_abort_pending(...)` is placed at the **end** of the `try` block in `close()`. If `self._proc.stdin.close()`, `self._proc.wait()`, `self._proc.terminate()`, or `self._proc.kill()` raises an unexpected exception (not `subprocess.TimeoutExpired`, which is already caught), the `except Exception` handler logs and returns, and `_abort_pending` is **never called**. Pending `request()` callers will park until their individual `timeout` fires — exactly the #1722 stall the PR aims to eliminate.

Since `_closed` is already set at the top of `close()` (under `_host_handlers_cv`), `_abort_pending` should be in a `finally` clause to guarantee it runs regardless of subprocess teardown failures:
```python
try:
    pid = self.pid
    with contextlib.suppress(Exception):
        self._proc.stdin.close()
    # ... wait / terminate / kill ...
except Exception:
    logger.exception("MemOS: unexpected error during bridge close")
finally:
    self._abort_pending(
        {
            "code": -32000,
            "message": "bridge closed",
            "data": {"code": "transport_closed"},
        }
    )
```

---

### 3. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L288-L293)

**[Medium] `close()` called after reader's `_abort_pending` fires will attempt subprocess teardown on an already-exited process.**

`_abort_pending()` sets `_aborted=True` and `_closed=True` but does **not** set `_close_called`. When the reader thread's `finally` clause fires `_abort_pending` (because the subprocess exited naturally) and `close()` is subsequently called, the `_close_called` guard in `close()` is still `False`. `close()` then proceeds to call `_proc.stdin.close()`, `_proc.wait()`, and potentially `_proc.terminate()`/`_proc.kill()` on an already-dead process. While `contextlib.suppress` and `TimeoutExpired` handling absorb most errors, sending `SIGTERM`/`SIGKILL` to a recycled PID is a real risk on long-running hosts.

Fix: check `self._aborted` (or both `_aborted` and `_close_called`) in the `close()` guard:
```python
with self._host_handlers_cv:
    if self._close_called or self._aborted:
        return
    self._close_called = True
    self._closed = True
    self._host_handlers_cv.notify_all()
```
This makes `close()` a no-op if the transport is already gone, consistent with the idempotency contract described in the comments.

---

### 4. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L413-L419)

**[Medium] Broad `except Exception` in `_read_loop` can mask programming errors as transport failures.**

The outer `except Exception` block is intended to catch pipe/IO errors that bubble up from `self._proc.stdout` iteration. However, `_send_response()` (called for reverse-RPC replies) acquires `_lock` and writes to stdin — an `AttributeError` or `TypeError` in that path would also bubble up here, silently terminate the reader thread, log a `WARNING`, and trigger `_abort_pending`, closing the entire bridge client due to a local programming mistake.

All the existing inner `except` blocks protect event/log callbacks and host handler dispatch, but `_send_response()` and `_resolve()` calls are not individually guarded. Consider narrowing the outer catch to IO-specific types:
```python
except (OSError, IOError):
    logger.warning("MemOS: bridge reader thread crashed", exc_info=True)
```
Or wrap individual `_send_response()`/`_resolve()` calls in `try/except Exception` so genuine programming bugs propagate naturally instead of being absorbed.

---

### 5. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1923-L1924)

The retry call at this line has no exception handler. If the freshly reconnected bridge also raises (e.g., another `BridgeError("timeout", ...)` or `transport_closed`), the exception propagates to the caller with no log entry and no recovery action. More importantly, `self._bridge` continues to reference the new (but already-broken) bridge, so the next caller will use it directly and likely hang again until the keepalive thread replaces it.

Consider wrapping the retry in a `try/except` that at minimum logs the failure at WARNING level, making the second failure's cause visible in logs. If the design intent is to allow at most one retry, explicitly documenting the post-retry state (and whether `self._bridge` should be cleared on second failure) would also improve correctness.

**Before:**
```
            result = new_bridge.request(method, params, **kwargs)
        return result if result is not None else {}
```

**After:**
```suggestion
            try:
                result = new_bridge.request(method, params, **kwargs)
            except Exception as retry_err:
                logger.warning(
                    "MemOS: bridge retry also failed for %s — %s",
                    method,
                    retry_err,
                )
                raise
        return result if result is not None else {}
```

</details>

---

### 6. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1916)

The reconnect timeout is hard-coded to `30.0` seconds regardless of the `timeout` parameter passed to `_bridge_request_with_retry`. This causes the keepalive thread's `core.health` call (which uses `timeout=10.0`) to block for up to 30 seconds during a reconnect triggered by the retry path, far exceeding its intended time budget. The total latency seen by any time-sensitive caller becomes unpredictable.

Consider deriving the reconnect timeout from the caller-supplied `timeout`, or accepting a separate `reconnect_timeout` parameter to make the budget explicit.

**Before:**
```
            self._reconnect_bridge(self._session_id, timeout=30.0, current_bridge=bridge)
```

**After:**
```suggestion
            reconnect_timeout = timeout if timeout is not None else 30.0
            self._reconnect_bridge(self._session_id, timeout=reconnect_timeout, current_bridge=bridge)
```

</details>

---

### 7. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1744-L1747)

`contextlib.suppress(OSError)` only suppresses OS-level errors from `poll()`. If `_proc` is not a `subprocess.Popen` instance (e.g., the attribute exists but holds a different type during an in-progress reconnect), `proc.poll()` will raise `AttributeError` or `TypeError`, which will propagate out of `_is_bridge_subprocess_dead` and crash the keepalive thread.

Since the purpose of this guard is defensive dead-process detection, consider broadening the suppression to `Exception` here, or adding an `isinstance(proc, subprocess.Popen)` check before calling `poll()`.

**Before:**
```
        with contextlib.suppress(OSError):
            if proc.poll() is not None:
                return True
        return False
```

**After:**
```suggestion
        with contextlib.suppress(Exception):
            if proc.poll() is not None:
                return True
        return False
```

</details>

---

### 8. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1924)

When a list-returning bridge method (such as `memory.search` via a different code path) legitimately returns `None` as an absent result, it is normalized to `{}` (an empty dict). If a caller later calls `.get("hits", [])` on this dict it works, but a caller expecting a list and iterating `for hit in result` will receive a dict, causing a `TypeError` or silent logical error.

The bridge's own `request()` method already normalises `None` to `{}` (as shown in `bridge_client.py`), so by the time the result reaches here it should never be `None` in practice. However, the defensive normalization here duplicates that logic and risks masking a future contract change. Consider removing it or normalizing to a type-appropriate sentinel (e.g. `[]` for list-returning methods), and documenting why `None` cannot arrive here.

**Before:**
```
        return result if result is not None else {}
```

**After:**
```suggestion
        # `bridge.request` already normalises None to {}; this is a belt-and-suspenders guard.
        return result if result is not None else {}
```

</details>

---

### 9. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L476-L477)

**[Correctness – Medium] `_exit_code` set directly bypasses `simulate_exit()`, leaving the stdout stream open and the reader thread running indefinitely.**

`self._fake._exit_code = 1` sets only the exit code but never calls `self.stdout.close_stream()`. The reader thread inside the `MemosBridgeClient` therefore never sees EOF and will keep spinning on 50 ms `_event.wait` timeouts. When `client.close()` later tries to join that thread it can hang until the daemon thread is finally abandoned by the process — making this test fragile on a loaded CI runner.

Replace the direct mutation with `simulate_exit()`, which atomically sets `_exit_code` *and* closes the stream:
```python
self._fake.simulate_exit(code=1)
```
If the intent is specifically to isolate the `poll()` short-circuit path (so the reader thread is intentionally left running), add a comment explaining that and verify `client.close()` won't block (i.e., the reader thread is a daemon thread).

**Before:**
```
        # Pretend the subprocess exited before our first request:
        self._fake._exit_code = 1
```

**After:**
```suggestion
        # Pretend the subprocess exited before our first request:
        self._fake.simulate_exit(code=1)
```

</details>

---

### 10. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L571-L573)

**[Correctness – Medium] The idempotency test relies on an implicit contract that `_abort_pending` drains `_pending`.**

The test inserts `e2` into `_pending` *after* the first `_abort_pending` call and asserts the second call leaves `e2` untouched. This only holds if the first call already removed all entries (making `_pending` empty when `e2` is added). If the production implementation ever iterates without popping, `e2` would still be present during the second call and receive `second_error`, silently breaking the assertion. Add an explicit drain assertion to surface this contract:

```python
client._abort_pending(first_error)
self.assertTrue(client._aborted)
self.assertEqual(len(client._pending), 0, "_abort_pending must drain _pending on first call")
```

---

### 11. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L82-L84)

**[Correctness – Low] `signal.SIGKILL` and `signal.SIGTERM` do not exist on Windows, causing `AttributeError` at call time.**

`signal.SIGKILL` is undefined on Windows (the `import signal` itself succeeds, but the attribute access fails). While the project currently targets POSIX, guarding with `getattr` costs nothing and prevents a confusing `AttributeError` if Windows CI is ever added:

```python
self._exit_code = -getattr(signal, "SIGKILL", 9)
```
```python
self._exit_code = -getattr(signal, "SIGTERM", 15)
```

**Before:**
```
            self._exit_code = -int(signal.SIGKILL)

    def terminate(self) -> None:
```

**After:**
```suggestion
            self._exit_code = -getattr(signal, "SIGKILL", 9)

    def terminate(self) -> None:
```

</details>

---

### 12. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L94)

**[Correctness – Low] `signal.SIGTERM` counterpart of the same Windows portability issue.**

**Before:**
```
            self._exit_code = -int(signal.SIGTERM)
```

**After:**
```suggestion
            self._exit_code = -getattr(signal, "SIGTERM", 15)
```

</details>

---

### 13. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L619-L629)

**[Correctness – Low] `call_count` dict is mutated from 6 concurrent threads without any synchronisation.**

`call_count["wait"] += 1` (and the `terminate`/`kill` variants) is not an atomic operation in CPython's general threading model — even though the GIL makes simple `dict` increments *usually* safe in practice, relying on this is an implicit assumption. If the test is ever ported or the production lock is held across the increment point, a data race could produce a count of 0 or 2 and make the `assertEqual(call_count["wait"], 1)` assertion non-deterministic.

Use `threading.Lock` or `collections.Counter` with a lock to make the counting bulletproof:
```python
_count_lock = threading.Lock()

def _counting_wait(*args, **kwargs):
    with _count_lock:
        call_count["wait"] += 1
    return original_wait(*args, **kwargs)
```

---

### 14. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L645-L647)

**[Maintainability – Low] `assertLessEqual(…, 1)` allows 0, but the comment says "Exactly one winner runs teardown" — the bound is misleading.**

`FakePopen.wait()` always returns immediately, so `terminate` and `kill` are correctly skipped by the production code's graceful-close path — making 0 the expected value, not 1. `assertLessEqual(…, 1)` passes for both 0 and 1, which means this assertion would not catch a hypothetical bug where both `terminate` *and* `kill` are called once each (total = 1 ≤ 1). Consider:
- Changing `assertLessEqual` to `assertEqual(..., 0)` to match the actual expected behaviour, **or**
- Adding an inline comment: `# FakePopen.wait() succeeds immediately → graceful path skips terminate/kill; 0 is the expected value, ≤1 guards against any double-invocation`.

---

### 15. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L1011-L1013)

**[Test Coverage – Medium] `EmptyListBridge` and `NoneBridge` stubs are missing a `close()` method.**

`_bridge_request_with_retry` may call `self._bridge.close()` on the retry/reconnect path before swapping in a new bridge. If it does, `AttributeError: 'EmptyListBridge' object has no attribute 'close'` would mask the actual assertion failure (i.e., you'd see the wrong error and not know the regression is present). Add defensive stubs:

```python
class EmptyListBridge:
    def close(self) -> None: pass
    def request(self, method, params=None, **_kwargs):
        return []
```

Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2018
**Task:** `ff6c49378bec35ae`

🔍 **OpenCodeReview** found **24** issue(s) in this PR.

---

### 1. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L291-L299)

**[C-01 · High] Mixed-lock write of `_closed` breaks the double-check in `request()`**

`close()` sets `self._closed = True` while holding `_host_handlers_cv` (line 292). But `request()` re-checks `self._closed` while holding `_lock` (line 203-204). These are **two different lock objects**, so the double-check pattern provides no coherent visibility guarantee — a thread inside `request()` that has already acquired `_lock` before `close()` flips `_closed` under `_host_handlers_cv` will proceed to insert into `_pending` **after** `_abort_pending` has already drained it, leaving the entry permanently orphaned until its per-request timeout fires.

The canonical fix is to also set `_closed = True` under `_lock` inside `close()` (or have `close()` call `_abort_pending` directly without the separate pre-set), which is exactly what `_abort_pending` already does — but `close()` sets the flag early (line 292) before `_abort_pending` is called in the `finally` block (line 353), creating the window.

**Suggested fix**: Remove the early `self._closed = True` from the `_host_handlers_cv` block in `close()` and rely solely on `_abort_pending`'s drain (which sets `_closed` under `_lock`) to enforce the invariant. The `_host_handlers_cv.notify_all()` can still fire early to unblock parked host-handler waiters.

---

### 2. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L298-L299)

**[C-02 · High] `_aborted` snapshot in `close()` uses the wrong lock**

`_aborted` is written under `self._lock` inside `_abort_pending()`, but is read at line 298 while holding `self._host_handlers_cv`. These are different locks with no ordering relationship, so there is no happens-before guarantee between the write and the read.

In practice on CPython the GIL makes this safe — the GIL is released and re-acquired on each lock acquire, creating an implicit memory barrier. However, the snapshot's stated purpose is to **skip subprocess teardown** when the reader already aborted: if the snapshot races and reads `False` when `_aborted` is actually `True`, `close()` redundantly attempts to `wait()`/`terminate()`/`kill()` on a process that is already dead or being torn down. Worse, on a non-CPython runtime (PyPy, GraalPy) or with future `free-threaded` CPython (PEP 703, `--disable-gil`), this is a genuine data race.

**Suggested fix**: Read `self._aborted` under `self._lock` instead:
```python
with self._lock:
    reader_already_aborted = self._aborted
```
or move the snapshot to after the `_host_handlers_cv` block closes, reading it under `_lock` before starting subprocess teardown.

---

### 3. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L203-L210)

**[C-03 · Medium] `self._proc.poll()` called while holding `_lock` — potential lock-order inversion**

`Popen.poll()` on CPython internally acquires a per-process POSIX lock (`_waitpid_lock` / `_popen_lock`) to read exit status. Holding `self._lock` while invoking `poll()` establishes the lock-order dependency:

> `_lock` → `_waitpid_lock`

If any other code path holds `_waitpid_lock` while trying to acquire `_lock` (e.g. a concurrent `communicate()` or `wait()` on the same `Popen` object from another thread), a classic ABBA deadlock results. The risk is low today since `wait()`/`communicate()` are only called from `close()` (which cannot overlap with `request()` after the `_close_called` guard), but the ordering is fragile and undocumented.

**Suggested fix**: Call `poll()` *before* acquiring `_lock`, save the result in a local, and re-use it inside the lock. The subsequent `_closed` re-check still covers the critical race:
```python
exit_code = self._proc.poll()  # outside _lock
with self._lock:
    if self._closed:
        raise BridgeError(...)
    if exit_code is not None:
        raise BridgeError(...)
    ...
```

---

### 4. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L432-L445)

**[C-04 · Medium] `ValueError` catch in `_read_loop` is broader than its comment implies**

The comment correctly identifies the *intended* `ValueError`: the `io` iterator raises `ValueError: I/O operation on closed file` when the underlying stream is closed externally. However, `ValueError` is also raised by:
- `json.loads()` in Python ≤ 3.4 for certain malformed inputs (fixed in 3.5+, but `json.JSONDecodeError` is a subclass of `ValueError`),
- Any future code added inside the loop body that manipulates sequences/dicts incorrectly.

The `json.JSONDecodeError` handler inside the loop already catches JSON parse errors before they propagate, but a future refactor adding a `ValueError`-raising operation inside the loop would be silently absorbed as a "pipe closed" event, converting a programmer bug into a masked transport error.

**Suggested fix**: Tighten the guard to check that the `ValueError` actually refers to a closed file:
```python
except ValueError as exc:
    if "closed file" not in str(exc) and "closed" not in str(exc).lower():
        raise
    logger.warning("MemOS: bridge reader thread crashed", exc_info=True)
```
Alternatively, restructure to catch only from the outermost `for` iterator rather than the full loop body.

---

### 5. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L483-L497)

**[C-05 · Low] Early return inside `_lock` in `_abort_pending` skips `_host_handlers_cv.notify_all()`**

When `_aborted` is already `True`, `_abort_pending` returns early from inside the `with self._lock:` block without ever notifying `_host_handlers_cv`. This is safe **only if** no thread can be parked in `_host_handler_for` after `_aborted` was set — which relies on `_closed` being visible to the `wait_for` predicate (`lambda: method in self._host_handlers or self._closed`).

If C-01's visibility gap exists (i.e. `_closed` is not coherently visible across `_host_handlers_cv` vs `_lock`), a thread could enter `_host_handler_for`'s `wait_for` loop after `_abort_pending` returns early, observe a stale `_closed = False`, and park until `HOST_HANDLER_WAIT_SECONDS` (5 s) expire with no notification arriving. This is a latent bug that becomes live if C-01 is left unaddressed.

---

### 6. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py`

**[T-01 · Medium] `test_abort_pending_is_idempotent_across_call_sites` leaves a dangling pending entry**

At the end of the idempotency test, `e2` is inserted into `client._pending[9102]` and intentionally not woken by the second `_abort_pending` call. The cleanup block (`client._pending.pop(9102, None)`) removes it before `client.close()` is called. However, if any assertion before that cleanup fails (e.g. `assertIsNone(e2["error"])`), the `pop` is skipped and `client.close()` will call `_abort_pending` which now tries to wake `e2.event`. Since `e2["event"]` is a real `threading.Event`, `.set()` on it is harmless — but the intent of leaving `e2` unresolved is to *prove* the second call was a no-op. An assertion failure mid-test will silently clean up via `close()`, potentially masking a broken invariant. Consider using `addCleanup` or a `try/finally` to ensure the pop happens unconditionally without obscuring the assertion failure.

---

### 7. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1928-L1933)

**P1 — reconnect_timeout too short for bridge startup when callers pass a tight deadline.**

`reconnect_timeout` is taken directly from the caller's `timeout`, which governs a single JSON-RPC round-trip. The bridge process startup + `session.open` handshake typically requires several seconds. When a tool call passes a short timeout (e.g. 2 s), the reconnect will almost certainly time out too, turning a transient failure into a permanent one for the rest of the session.

Consider applying a floor so the reconnect always has enough headroom:
```python
reconnect_timeout = max(timeout, 10.0) if timeout is not None else 30.0
```
Alternatively, always use `30.0` for `session.open` and only propagate the caller's `timeout` to the retry `request()` call.

**Before:**
```
        reconnect_timeout = timeout if timeout is not None else 30.0
            self._reconnect_bridge(
                self._session_id,
                timeout=reconnect_timeout,
                current_bridge=bridge,
            )
```

**After:**
```suggestion
            reconnect_timeout = max(timeout, 10.0) if timeout is not None else 30.0
            self._reconnect_bridge(
                self._session_id,
                timeout=reconnect_timeout,
                current_bridge=bridge,
            )
```

</details>

---

### 8. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1738-L1740)

**C1 — Private `_proc` attribute coupling will silently degrade liveness detection.**

`getattr(bridge, '_proc', None)` couples this method to `MemosBridgeClient`'s internal layout. If `_proc` is ever renamed (e.g. to `_process`), this silently returns `False` and dead-bridge detection stops working with no error or warning — exactly the class of silent regression that is hardest to diagnose in production.

Looking at `bridge_client.py`, it does not expose a public `is_alive()` or `poll()` API. Consider adding one to `MemosBridgeClient` (e.g. `def is_subprocess_running(self) -> bool`) so the coupling is contractual rather than structural.

**Before:**
```
        proc = getattr(bridge, "_proc", None)
        if proc is None:
            return False
```

**After:**
```suggestion
        proc = getattr(bridge, "_proc", None)
        if proc is None:
            # If _proc is absent (e.g. attribute renamed), fail open — treat
            # as alive so we don't suppress a potentially-needed reconnect.
            logger.debug("MemOS: _is_bridge_subprocess_dead: _proc not found on bridge instance")
            return False
```

</details>

---

### 9. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1958)

**M1 — `None` normalised to `{}` (dict) violates the `list[Any]` branch of the return type.**

When `bridge.request()` legitimately returns `None` for a list-typed method such as `memory.search`, this returns `{}`. Call sites that call `.get("hits", [])` on the result will not crash, but the return type annotation (`dict[str, Any] | list[Any]`) is violated — a `dict` is returned where `list` was expected. If any call site ever does `for item in result:` instead of `result.get(...)`, it will silently iterate over dict keys rather than search hits.

Consider preserving the distinction by returning `[]` for a `None` result on list-typed methods, or documenting that `None` is always normalised to `{}` and callers must never assume they receive a bare list from this wrapper.

**Before:**
```
        return result if result is not None else {}
```

**After:**
```suggestion
        # Preserve empty collections verbatim; only a genuinely absent
        # result (``None``) is normalised to ``{}``. Using ``result or {}``
        # here would silently collapse a zero-hit ``memory.search`` list
        # into a dict and break every downstream `for hit in resp["hits"]`
        # iteration.
        return result if result is not None else {}
```

</details>

---

### 10. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1928-L1933)

**P1 — `reconnect_timeout` inherits the caller's short tool deadline, almost guaranteeing reconnect failure.**

`reconnect_timeout = timeout if timeout is not None else 30.0` means any caller that passes a tight per-request `timeout` (e.g. 2 s to avoid stalling a UI) will also constrain `session.open` during the bridge startup handshake, which routinely takes several seconds. The reconnect will time out and `self._bridge` will remain `None`, turning a single transient transport failure into a permanent outage for the session.

Consider applying a floor so reconnect always has adequate headroom:
```python
reconnect_timeout = max(timeout, 10.0) if timeout is not None else 30.0
```
Alternatively, always use the full `30.0` for the `_reconnect_bridge` call and propagate the caller's `timeout` only to the retry `request()` invocation.

**Before:**
```
            reconnect_timeout = timeout if timeout is not None else 30.0
            self._reconnect_bridge(
                self._session_id,
                timeout=reconnect_timeout,
                current_bridge=bridge,
            )
```

**After:**
```suggestion
            reconnect_timeout = max(timeout, 10.0) if timeout is not None else 30.0
            self._reconnect_bridge(
                self._session_id,
                timeout=reconnect_timeout,
                current_bridge=bridge,
            )
```

</details>

---

### 11. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1738-L1740)

**C1 — Private `_proc` attribute coupling silently degrades dead-subprocess detection.**

`getattr(bridge, '_proc', None)` ties this health check to `MemosBridgeClient`'s internal layout. If `_proc` is renamed in a future refactor, `getattr` returns `None` silently and this method always returns `False` — dead-bridge detection stops working without any log line or test failure to surface the regression.

The `bridge_client.py` diff in this PR does not expose a public liveness API. Consider adding one (e.g. `def is_subprocess_running(self) -> bool`) so the contract is explicit and renaming `_proc` would cause a clear `AttributeError` at the call site rather than a silent no-op.

---

### 12. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1958)

**M1 — `None` response normalised to `{}` (dict) violates the `list[Any]` branch of the declared return type.**

When `bridge.request()` returns `None` for a list-typed method (e.g. `memory.search`), callers receive `{}`. Call sites using `.get("hits", [])` won't crash, but the return type annotation `dict[str, Any] | list[Any]` is violated. Any future call site that iterates `for hit in result:` directly (without `.get(...)`) will silently iterate over dict *keys* instead of search hits.

Consider either:
- Returning `[]` for `None` results on known list-typed methods, or
- Narrowing the wrapper's contract: document that it *always* returns a `dict` (never a bare `list`) and have callers use `.get("hits", [])` uniformly. Update the return annotation to `dict[str, Any]` accordingly.

---

### 13. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1738-L1740)

**C1 — Private `_proc` attribute access creates a silent regression risk.**

`getattr(bridge, '_proc', None)` ties dead-subprocess detection to `MemosBridgeClient`'s internal naming. If `_proc` is ever renamed, `getattr` silently returns `None`, this method always returns `False`, and the keepalive dead-process check stops working with no log, no error, and no failing test to surface the breakage.

The `bridge_client.py` diff in this PR does not expose a public liveness API. Consider adding one (e.g. `def is_subprocess_running(self) -> bool`) so the contract is explicit and a structural rename would cause a clear `AttributeError` at the call site rather than a silent no-op.

---

### 14. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1928-L1933)

**P1 — `reconnect_timeout` inherits the caller's short tool deadline, almost guaranteeing reconnect failure.**

`reconnect_timeout = timeout if timeout is not None else 30.0` propagates the per-request deadline directly into `_reconnect_bridge`. Bridge startup (Node process spawn + `session.open` handshake) typically takes several seconds. A tool call that passes a tight timeout (e.g. 2 s to avoid stalling a UI) will also constrain `session.open`, causing the reconnect to time out and leaving `self._bridge = None` — turning a single transient transport failure into a permanent outage for the session.

Consider applying a floor so the reconnect always has adequate headroom:
```python
reconnect_timeout = max(timeout, 10.0) if timeout is not None else 30.0
```
Alternatively, always use the full `30.0` for `_reconnect_bridge` and propagate the caller's `timeout` only to the retry `request()` call.

**Before:**
```
            reconnect_timeout = timeout if timeout is not None else 30.0
            self._reconnect_bridge(
                self._session_id,
                timeout=reconnect_timeout,
                current_bridge=bridge,
            )
```

**After:**
```suggestion
            reconnect_timeout = max(timeout, 10.0) if timeout is not None else 30.0
            self._reconnect_bridge(
                self._session_id,
                timeout=reconnect_timeout,
                current_bridge=bridge,
            )
```

</details>

---

### 15. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1958)

**M1 — `None` normalised to `{}` (dict) violates the `list[Any]` branch of the declared return type.**

When `bridge.request()` returns `None` for a list-typed method (e.g. `memory.search`), callers receive `{}`. Current call sites all use `.get("hits", [])` etc. so they don't crash today. However, the declared return type `dict[str, Any] | list[Any]` is violated: a `dict` is returned where the annotation says a `list` is possible. Any future call site that iterates `for hit in result:` directly (without `.get(...)`) would silently iterate over dict keys instead of search hits.

Consider either:
- Documenting that this wrapper always returns a `dict` and narrowing the annotation to `dict[str, Any]` (callers must always use `.get(...)`), or
- Returning `[]` for `None` results on list-typed methods so the type annotation stays accurate.

---

### 16. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1738-L1740)

**C1 — Private `_proc` attribute access creates a silent regression risk.**

`getattr(bridge, '_proc', None)` ties dead-subprocess detection to `MemosBridgeClient`'s internal naming. If `_proc` is ever renamed in a refactor, `getattr` silently returns `None`, this method always returns `False`, and the keepalive dead-process check stops working with no log, no error, and no failing test to surface the breakage.

The `bridge_client.py` diff in this PR does not expose a public liveness API. Consider adding one (e.g. `def is_subprocess_running(self) -> bool`) so the contract is explicit and a structural rename would cause a clear `AttributeError` at the call site rather than a silent no-op.

---

### 17. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1928-L1933)

**P1 — `reconnect_timeout` inherits the caller's short tool deadline, risking reconnect failure.**

`reconnect_timeout = timeout if timeout is not None else 30.0` propagates the per-request deadline directly into `_reconnect_bridge`. Bridge startup (Node process spawn + `session.open` handshake) typically takes several seconds. A tool call that passes a tight timeout (e.g. 2 s) will also constrain `session.open`, causing the reconnect to fail and leaving `self._bridge = None` — turning a transient transport failure into a permanent outage for the session.

Consider applying a floor so the reconnect always has adequate headroom:
```python
reconnect_timeout = max(timeout, 10.0) if timeout is not None else 30.0
```
Alternatively, always use the full `30.0` for `_reconnect_bridge` and propagate the caller's `timeout` only to the retry `request()` call.

**Before:**
```
            reconnect_timeout = timeout if timeout is not None else 30.0
            self._reconnect_bridge(
                self._session_id,
                timeout=reconnect_timeout,
                current_bridge=bridge,
            )
```

**After:**
```suggestion
            reconnect_timeout = max(timeout, 10.0) if timeout is not None else 30.0
            self._reconnect_bridge(
                self._session_id,
                timeout=reconnect_timeout,
                current_bridge=bridge,
            )
```

</details>

---

### 18. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1958)

**M1 — `None` normalised to `{}` (dict) violates the `list[Any]` branch of the declared return type.**

When `bridge.request()` returns `None` for a list-typed method (e.g. `memory.search`), callers receive `{}`. Current call sites all use `.get("hits", [])` etc. so they don't crash today. However, the declared return type `dict[str, Any] | list[Any]` is violated: a `dict` is returned where the annotation says a `list` is possible. Any future call site that iterates `for hit in result:` directly (without `.get(...)`) would silently iterate over dict keys instead of search hits.

Consider either:
- Documenting that this wrapper always returns a `dict` and narrowing the annotation to `dict[str, Any]` (callers must always use `.get(...)`), or
- Returning `[]` for `None` results on list-typed methods so the type annotation stays accurate.

---

### 19. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1738-L1740)

**C1 — Private `_proc` attribute access creates a silent regression risk.**

`getattr(bridge, '_proc', None)` ties dead-subprocess detection to `MemosBridgeClient`'s internal naming. If `_proc` is ever renamed in a refactor, `getattr` silently returns `None`, this method always returns `False`, and the keepalive dead-process check stops working with no log, no error, and no failing test to surface the breakage.

The `bridge_client.py` diff in this PR does not expose a public liveness API. Consider adding one (e.g. `def is_subprocess_running(self) -> bool`) so the contract is explicit and a structural rename would cause a clear `AttributeError` at the call site rather than a silent no-op.

---

### 20. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1928-L1933)

**P1 — `reconnect_timeout` inherits the caller's short tool deadline, risking reconnect failure.**

`reconnect_timeout = timeout if timeout is not None else 30.0` propagates the per-request deadline directly into `_reconnect_bridge`. Bridge startup (Node process spawn + `session.open` handshake) typically takes several seconds. A tool call that passes a tight timeout (e.g. 2 s to avoid stalling a UI) will also constrain `session.open`, causing the reconnect to fail and leaving `self._bridge = None` — turning a transient transport failure into a permanent outage for the session.

Consider applying a floor so the reconnect always has adequate headroom:
```python
reconnect_timeout = max(timeout, 10.0) if timeout is not None else 30.0
```
Alternatively, always use the full `30.0` for `_reconnect_bridge` and propagate the caller's `timeout` only to the retry `request()` call.

**Before:**
```
            reconnect_timeout = timeout if timeout is not None else 30.0
            self._reconnect_bridge(
                self._session_id,
                timeout=reconnect_timeout,
                current_bridge=bridge,
            )
```

**After:**
```suggestion
            reconnect_timeout = max(timeout, 10.0) if timeout is not None else 30.0
            self._reconnect_bridge(
                self._session_id,
                timeout=reconnect_timeout,
                current_bridge=bridge,
            )
```

</details>

---

### 21. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1958)

**M1 — `None` normalised to `{}` (dict) violates the `list[Any]` branch of the declared return type.**

When `bridge.request()` returns `None` for a list-typed method (e.g. `memory.search`), callers receive `{}`. Current call sites all use `.get("hits", [])` etc. so they don't crash today. However, the declared return type `dict[str, Any] | list[Any]` is violated: a `dict` is returned where the annotation says a `list` is possible. Any future call site that iterates `for hit in result:` directly (without `.get(...)`) would silently iterate over dict keys instead of search hits.

Consider either:
- Documenting that this wrapper always returns a `dict` and narrowing the annotation to `dict[str, Any]` (callers must always use `.get(...)`), or
- Returning `[]` for `None` results on list-typed methods so the type annotation stays accurate.

---

### 22. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1738-L1740)

**C1 — Private `_proc` attribute access creates a silent regression risk.**

`getattr(bridge, '_proc', None)` ties dead-subprocess detection to `MemosBridgeClient`'s internal naming. If `_proc` is ever renamed in a refactor, `getattr` silently returns `None`, this method always returns `False`, and the keepalive dead-process check stops working with no log, no error, and no failing test to surface the breakage.

The `bridge_client.py` diff in this PR does not expose a public liveness API. Consider adding one (e.g. `def is_subprocess_running(self) -> bool`) so the contract is explicit and a structural rename would cause a clear `AttributeError` at the call site rather than a silent no-op.

---

### 23. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1928-L1933)

**P1 — `reconnect_timeout` inherits the caller's short tool deadline, risking reconnect failure.**

`reconnect_timeout = timeout if timeout is not None else 30.0` propagates the per-request deadline directly into `_reconnect_bridge`. Bridge startup (Node process spawn + `session.open` handshake) typically takes several seconds. A tool call that passes a tight timeout (e.g. 2 s to avoid stalling a UI) will also constrain `session.open`, causing the reconnect to fail and leaving `self._bridge = None` — turning a transient transport failure into a permanent outage for the session.

Consider applying a floor so the reconnect always has adequate headroom:
```python
reconnect_timeout = max(timeout, 10.0) if timeout is not None else 30.0
```
Alternatively, always use the full `30.0` for `_reconnect_bridge` and propagate the caller's `timeout` only to the retry `request()` call.

**Before:**
```
            reconnect_timeout = timeout if timeout is not None else 30.0
            self._reconnect_bridge(
                self._session_id,
                timeout=reconnect_timeout,
                current_bridge=bridge,
            )
```

**After:**
```suggestion
            reconnect_timeout = max(timeout, 10.0) if timeout is not None else 30.0
            self._reconnect_bridge(
                self._session_id,
                timeout=reconnect_timeout,
                current_bridge=bridge,
            )
```

</details>

---

### 24. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1958)

**M1 — `None` normalised to `{}` (dict) violates the `list[Any]` branch of the declared return type.**

When `bridge.request()` returns `None` for a list-typed method (e.g. `memory.search`), callers receive `{}`. Current call sites all use `.get("hits", [])` etc. so they don't crash today. However, the declared return type `dict[str, Any] | list[Any]` is violated: a `dict` is returned where the annotation says a `list` is possible. Any future call site that iterates `for hit in result:` directly (without `.get(...)`) would silently iterate over dict keys instead of search hits.

Consider either:
- Documenting that this wrapper always returns a `dict` and narrowing the annotation to `dict[str, Any]` (callers must always use `.get(...)`), or
- Returning `[]` for `None` results on list-typed methods so the type annotation stays accurate.

Memtensor-AI:
## ⚠️ Open Code Review automatic OCR fix limit reached

Open Code Review still found **24** issue(s), but the automatic OCR fix loop has reached **2/2** attempts.

I stopped auto-fixing this PR to avoid an infinite loop. Please review the latest OCR comment and fix or dismiss the remaining findings manually.

_PR: https://github.com/MemTensor/MemOS/pull/2018_

PR #2027, opened 2026-07-02

### LINKED ISSUE #1722: fix: memory_search returned error

Memtensor-AI:
## Description

Fixes #1722 — `memory_search` returned `[timeout] memory.search did not respond within 30.0s` after the hermes plugin's Node bridge went silent, followed by the same failure on `turn.end`. Three cooperating Python-side defects: (a) `MemosBridgeClient._read_loop` iterated stdout with no try/finally so parked JSON-RPC waiters stayed asleep past EOF for the full 30 s per-request timeout; (b) `_start_bridge_keepalive` reconnected only on `transport_closed` and dropped `BridgeError("timeout", …)` — the exact code raised for the reported symptom — at DEBUG; (c) read-path memory tools (`memos_search`, `memos_get`, `memos_timeline`, `memos_environment`, `memos_skill_list`, `memos_skill_get`) called `_bridge.request` once with no retry, so even the first user-visible call to a stale bridge surfaced a bare failure.

Fix keeps the JSON-RPC wire protocol untouched. Reader loop now aborts every pending waiter with `transport_closed` on any exit via the new `_abort_pending` helper (also reused by `close()`). `request()` fast-fails on `Popen.poll() != None` before ever writing to a dead pipe. Keepalive delegates to a new `_should_reconnect_after_keepalive_failure` helper that reconnects on `transport_closed`, `BridgeError("timeout", …)`, and any error observed on an already-dead subprocess (while staying quiet for generic transient errors on a live subprocess, to prevent reconnect storms). Every read-path tool now goes through `_bridge_request_with_retry`, which mirrors the existing `sync_turn` write-path pattern — reconnect + retry once on transport_closed, surface any second failure verbatim.

Tests: added 8 new cases under `apps/memos-local-plugin/tests/python/test_bridge_client.py` covering reader-thread abort, fast-fail on dead subprocess, `close()` wake-pending, keepalive reconnect on `timeout` / dead subprocess / no-op on transient generic error, and `handle_tool_call` retry that both succeeds and surfaces a second transport failure. `python3 -m unittest tests.python.test_bridge_client` = 34 tests OK; `tests.python.test_hermes_provider_pipeline` = 16 tests OK; `ruff check` + `ruff format --check` clean.

Branch `bugfix/autodev-1722` pushed to origin at c65d96e2. opsp artifacts (task.md + proposal/spec/design/verification-report) archived to `memos-autodev-specs` main under `2026-07-02-1722-fix-memorysearch-returned-error/`.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Automated tests are pending.

- [ ] Unit Test

## Checklist

## Reviewer Checklist
- [x] closes #1722
- [ ] Made sure Checks passed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@
 ##############################################################################
 
 name = "MemoryOS"
-version = "2.0.19"
+version = "2.0.20"
 description = "Intelligence Begins with Memory"
 license = {text = "Apache-2.0"}
 readme = "README.md"
--- a/src/memos/__init__.py
+++ b/src/memos/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "2.0.19"
+__version__ = "2.0.20"
 
 from memos.configs.mem_cube import GeneralMemCubeConfig
 from memos.configs.mem_os import MOSConfig
--- a/src/memos/api/handlers/formatters_handler.py
+++ b/src/memos/api/handlers/formatters_handler.py
@@ -13,6 +13,13 @@ from memos.templates.instruction_completion import instruct_completion
 
 logger = get_logger(__name__)
 
+__all__ = [
+    "format_memory_item",
+    "post_process_textual_mem",
+    "rerank_knowledge_mem",
+    "to_iter",
+]
+
 
 def to_iter(running: Any) -> list[Any]:
     """
@@ -164,6 +171,7 @@ def rerank_knowledge_mem(
     text_mem: list[dict[str, Any]],
     top_k: int,
     file_mem_proportion: float = 0.5,
+    strip_conversation_sources: bool = False,
 ) -> list[dict[str, Any]]:
     """
     Rerank knowledge memories and keep conversation memories.
@@ -193,8 +201,9 @@ def rerank_knowledge_mem(
         item["memory"] = item["metadata"]["sources"][0]["content"]
         item["metadata"]["sources"] = []
 
-    for item in conversation_mem:
-        item.setdefault("metadata", {})["sources"] = []
+    if strip_conversation_sources:
+        for item in conversation_mem:
+            item.setdefault("metadata", {})["sources"] = []
 
     # deduplicate: remove items with duplicate memory content
     original_count = len(reranked_knowledge_mem)
--- a/src/memos/api/handlers/search_handler.py
+++ b/src/memos/api/handlers/search_handler.py
@@ -118,6 +118,23 @@ class SearchHandler(BaseHandler):
             top_k=search_req_local.top_k,
             file_mem_proportion=0.5,
         )
+        hooked_results = trigger_hook(
+            H.SEARCH_RESULTS_AFTER_RERANK,
+            handler=self,
+            search_req=search_req_local,
+            results=results,
+        )
+        if hooked_results is not None:
+            results = hooked_results
+
+        hooked_results = trigger_hook(
+            H.SEARCH_CONTEXT_RENDER,
+            handler=self,
+            search_req=search_req_local,
+            results=results,
+        )
+        if hooked_results is not None:
+            results = hooked_results
 
         self.logger.info(
             f"[SearchHandler] Final search results: count={len(results)} results={results}"
--- a/src/memos/api/product_models.py
+++ b/src/memos/api/product_models.py
@@ -422,6 +422,14 @@ class APISearchRequest(BaseRequest):
         ),
     )
 
+    rerank: bool = Field(
+        True,
+        description=(
+            "Whether to apply the textual memory reranker during search. "
+            "Set false to return retrieval-order candidates before post-search dedup/formatting."
+        ),
+    )
+
     pref_top_k: int = Field(
         6,
         ge=0,
@@ -462,6 +470,11 @@ class APISearchRequest(BaseRequest):
         3,
         ge=0,
         description="Number of skill memories to retrieve (top-K). Default: 3.",
+    )
+
+    context_format: str = Field(
+        "memory",
+        description="Optional search context format passed through to installed plugins.",
     )
 
     # ==== Filter conditions ====
--- a/src/memos/memories/textual/tree_text_memory/retrieve/searcher.py
+++ b/src/memos/memories/textual/tree_text_memory/retrieve/searcher.py
@@ -77,6 +77,30 @@ class Searcher:
         self.tokenizer = tokenizer
         self._usage_executor = ContextThreadPoolExecutor(max_workers=4, thread_name_prefix="usage")
 
+    def _maybe_rerank(
+        self,
+        enabled: bool,
+        *,
+        query: str,
+        graph_results: list[TextualMemoryItem],
+        top_k: int,
+        **kwargs,
+    ) -> list[tuple[TextualMemoryItem, float]]:
+        if not enabled or self.reranker is None:
+            return [(item, 0.0) for item in graph_results[:top_k]]
+        return self.reranker.rerank(
+            query=query,
+            graph_results=graph_results,
+            top_k=top_k,
+            **kwargs,
+        )
+
+    @staticmethod
+    def _query_embedding_for_rerank(enabled: bool, query_embedding):
+        if not enabled:
+            return None
+        return query_embedding[0]
+
     @timed
     def retrieve(
         self,
@@ -99,6 +123,7 @@ class Searcher:
         logger.info(
             f"[RECALL] Start query='{query}', top_k={top_k}, mode={mode}, memory_type={memory_type}, user_name={user_name}"
         )
+        rerank = bool(kwargs.get("rerank", True))
         parsed_goal, query_embedding, _context, query = self._parse_task(
             query,
             info,
@@ -125,6 +150,7 @@ class Searcher:
             skill_mem_top_k,
             include_preference_memory,
             pref_mem_top_k,
+            rerank,
         )
         return results
 
@@ -350,6 +376,7 @@ class Searcher:
         skill_mem_top_k: int = 3,
         include_preference_memory: bool = False,
         pref_mem_top_k: int = 6,
+        rerank: bool = True,
     ):
         """Run A/B/C/D/E/F retrieval paths in parallel"""
         tasks = []
@@ -372,6 +399,7 @@ class Searcher:
                     search_priority,
                     user_name,
                     id_filter,
+                    rerank=rerank,
                 )
             )
             tasks.append(
@@ -387,6 +415,7 @@ class Searcher:
                     user_name,
                     id_filter,
                     mode=mode,
+                    rerank=rerank,
                 )
             )
             tasks.append(
@@ -400,6 +429,7 @@ class Searcher:
                     mode,
                     memory_type,
                     user_name,
+                    rerank=rerank,
                 )
             )
             if self.use_fulltext:
@@ -415,6 +445,7 @@ class Searcher:
                         search_priority,
                         user_name,
                         id_filter,
+                        rerank=rerank,
                     )
                 )
             if search_tool_memory:
@@ -431,6 +462,7 @@ class Searcher:
                         user_name,
                         id_filter,
                         mode=mode,
+                        rerank=rerank,
                     )
                 )
             if include_skill_memory:
@@ -447,6 +479,7 @@ class Searcher:
                         user_name,
                         id_filter,
                         mode=mode,
+                        rerank=rerank,
                     )
                 )
             if include_preference_memory:
@@ -463,6 +496,7 @@ class Searcher:
                         user_name,
                         id_filter,
                         mode=mode,
+                        rerank=rerank,
                     )
                 )
             results = []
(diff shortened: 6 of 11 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2027
**Task:** `d24d1e93a3e30f26`

---

### 1. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L173-L190)

**Race condition (TOCTOU): `_closed` check is outside `_lock`, so a new pending entry can be orphaned.**

The sequence is:
1. Thread A reads `self._closed` → `False`, passes the guard at line 173.
2. Thread A calls `self._proc.poll()` → `None`, passes the fail-fast check.
3. `_abort_pending` (from reader-thread `finally`) runs on another thread: sets `_closed = True`, clears `_pending`, wakes all waiters.
4. Thread A acquires `_lock`, adds a new entry to `_pending` (line 190), and sends the payload.
5. Nobody will ever resolve that entry — it stalls until `timeout` fires.

The fix is to re-check `_closed` inside `_lock` *after* registering the pending entry, and remove it + raise immediately if `_closed` is now `True`. Alternatively, hold a single lock that guards both `_closed` and `_pending` throughout the registration path.

**Before:**
```
        if self._closed:
            raise BridgeError("transport_closed", "bridge client is closed")
        # Layer 2 (fix for #1722): fail-fast when the subprocess has
        # already exited. Otherwise the pipe write may silently succeed
        # into an OS buffer and the caller stalls for the full
        # per-request timeout before failing.
        exit_code = self._proc.poll() if self._proc is not None else None
        if exit_code is not None:
            raise BridgeError(
                "transport_closed",
                f"bridge subprocess exited (code={exit_code})",
            )
        with self._lock:
            rpc_id = self._next_id
            self._next_id += 1
            waiter = threading.Event()
            entry: dict[str, Any] = {"event": waiter, "result": None, "error": None}
            self._pending[rpc_id] = entry
```

**After:**
```suggestion
        with self._lock:
            # Re-check _closed inside _lock to close the TOCTOU window
            # between the guard above and the entry registration below.
            if self._closed:
                raise BridgeError("transport_closed", "bridge client is closed")
            exit_code = self._proc.poll() if self._proc is not None else None
            if exit_code is not None:
                raise BridgeError(
                    "transport_closed",
                    f"bridge subprocess exited (code={exit_code})",
                )
            rpc_id = self._next_id
            self._next_id += 1
            waiter = threading.Event()
            entry: dict[str, Any] = {"event": waiter, "result": None, "error": None}
            self._pending[rpc_id] = entry
```

</details>

---

### 2. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L294-L313)

**Race condition in `_abort_pending`: gap between releasing `_host_handlers_cv` and acquiring `_lock` can leave new `_pending` entries orphaned.**

`_closed` is set to `True` inside `_host_handlers_cv`, but `_pending` is cleared inside a *separate* `_lock`. Between the two lock acquisitions, a racing thread can:
1. Read `_closed == True` → False (before `_host_handlers_cv` is released), register an entry in `_pending` while `_abort_pending` still holds `_host_handlers_cv`.
2. `_abort_pending` releases `_host_handlers_cv`, acquires `_lock`, clears `_pending` — *but the new entry was added after the clear* if insertion races past the `_lock` acquisition order.

More importantly, there is no guarantee the `_closed = True` write (visible to `request()`) and the `_pending.clear()` appear atomic to a racing `request()` call. Consider consolidating `_closed` under `_lock` (or ensuring `_abort_pending` re-checks `_pending` after holding `_lock` for the full critical section).

---

### 3. `apps/memos-local-plugin/adapters/hermes/memos_provider/bridge_client.py` (L384-L385)

**Unexpected reader-loop exceptions are silently swallowed at DEBUG level.**

`except Exception` is deliberately broad, but `logger.debug(...)` means that in production — where DEBUG is typically off — all unexpected exceptions (e.g., `AttributeError`, `OSError` during line iteration, `MemoryError`) will be invisible. This makes post-mortem diagnosis very difficult. Consider logging at `WARNING` or higher for truly unexpected exceptions, reserving `DEBUG` only for the expected clean-EOF path.

**Before:**
```
        except Exception:
            logger.debug("bridge reader loop terminated with exception", exc_info=True)
```

**After:**
```suggestion
        except Exception:
            logger.warning("bridge reader loop terminated with exception", exc_info=True)
```

</details>

---

### 4. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1292-L1296)

**Inconsistency: timeout errors are not retried here, but they trigger reconnect in `_should_reconnect_after_keepalive_failure`.**

`_is_transport_closed` only matches `BridgeError(code='transport_closed', ...)`. A `BridgeError('timeout', ...)` — raised by `MemosBridgeClient.request` when the Node bridge stops answering (see `bridge_client.py` line 205) — will propagate uncaught from this helper without a retry or reconnect attempt, even though `_should_reconnect_after_keepalive_failure` explicitly treats `timeout` as a reconnect trigger.

This means a hung bridge causes `memos_search` / `skill.list` / etc. to return `{"error": "[timeout] memory.search did not respond within 30.0s"}` to the LLM on first occurrence instead of reconnecting and retrying, and the bridge is only cleaned up later by the keepalive thread — reproducing the original #1722 symptom for a different error code.

Suggested fix: extend the retry condition to include timeout:
```python
except BridgeError as err:
    if not self._is_transport_closed(err) and not (isinstance(err, BridgeError) and err.code == "timeout"):
        raise
```
Or better, extract the shared predicate used by both `_bridge_request_with_retry` and `_should_reconnect_after_keepalive_failure` into a single `_is_retriable_transport_error` helper to avoid this divergence.

---

### 5. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1302-L1304)

**TOCTOU race on `self._bridge` between `_reconnect_bridge` and the retry call.**

`_reconnect_bridge` acquires `_reconnect_lock` to write `self._bridge = new_bridge`, then returns and releases the lock. This helper then reads `self._bridge` outside that lock for the retry. The keepalive thread runs every 5 s and can concurrently call `_reconnect_bridge` again — which would set `self._bridge = None` briefly (line 1851) or replace it with another new client — between the `assert` and the `.request(...)` call.

Capture the bridge reference immediately after reconnect to avoid this window:
```python
br = self._bridge
if br is None:
    raise BridgeError("transport_closed", "reconnect did not produce a bridge")
return br.request(method, params, **kwargs)
```

---

### 6. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L379-L384)

**Racy synchronization via `time.sleep(0.05)`**: This sleep is used as a heuristic to wait until the background thread has written the request and parked it in `_pending`. On a slow or heavily loaded CI machine the 50 ms may not be enough — the EOF signal (`_done = True` + `_event.set()`) could arrive before the waiter is actually parked, meaning `_abort_pending` fires before the entry exists in `self._pending`, the worker then parks and waits the full `timeout=10.0 s`, and the `worker.join(timeout=2.0)` returns with the worker still alive, causing the test to fail with a misleading message.

A robust alternative is to use an `threading.Event` that `FakePopen`'s stdin write path sets when the first request arrives, replacing the sleep:
```python
request_received = threading.Event()
# ... monkeypatch stdin.write to set event ...
worker.start()
request_received.wait(timeout=2.0)  # deterministic
self._fake.stdout._done = True
...
```
This same pattern applies to `test_close_wakes_pending_waiters`.

---

### 7. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L409-L411)

**Busy-polling private `client._closed` without synchronization**: `_closed` is set inside `_host_handlers_cv` (a `threading.Condition`) in the production code, but the test reads it without holding that lock. While CPython's GIL makes bare boolean reads effectively atomic in practice, this is still technically a data race that could be flagged by threading sanitizers.

More importantly, this busy-wait loop exits via timeout (after 2 s) if the reader thread hasn't set `_closed` yet. In that case the subsequent `assertRaises(BridgeError)` may still pass — but for the wrong reason: the subprocess poll check at the top of `request()` also raises `transport_closed`, so the test can pass even if the reader thread never processed the EOF, masking a potential bug in `_abort_pending`.

Consider exposing a test-only synchronization hook (e.g., join the reader thread `client._reader.join(timeout=2.0)`) to deterministically wait for the reader loop to finish before asserting.

---

### 8. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L437-L441)

**`test_close_wakes_pending_waiters` leaks background threads**: `client.close()` is called as the operation under test, which is correct. However, `MemosBridgeClient.__init__` always spawns a reader thread (`self._reader`) and a stderr reader thread (`self._stderr_reader`). After `close()`, the stderr reader thread may still be alive (it loops over `self._proc.stderr` which `FakePopen.stderr` is an `io.StringIO` — the loop ends immediately, so this is likely fine in practice). Still, it's worth adding an assertion or a join to ensure both daemon threads have exited to avoid any cross-test interference.

---

### 9. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L379-L384)

**Racy synchronization via `time.sleep(0.05)`**: This sleep heuristic is unreliable on slow or heavily loaded CI machines. If the 50 ms elapses before the worker thread has written the request and parked it in `_pending`, the EOF signal (`_done = True` + `_event.set()`) will fire before the waiter exists, so `_abort_pending` finds an empty `_pending` dict, clears nothing, and the worker subsequently parks and waits the full 10 s — causing `worker.join(timeout=2.0)` to return with the worker still alive and a spurious test failure.

The same risk exists in `test_close_wakes_pending_waiters`.

A robust alternative is to synchronize on the arrival of the request at the fake stdin. The `FakePopen` stdin write path already appends to `_stdin_lines`; you could patch it to also set a `threading.Event`, e.g.:
```python
request_received = threading.Event()
original_write = self._fake.stdin.write
self._fake.stdin.write = lambda line: (original_write(line), request_received.set())
worker.start()
request_received.wait(timeout=2.0)  # deterministic
self._fake.stdout._done = True
...
```

---

### 10. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L409-L411)

**Busy-polling `client._closed` without a lock and ambiguous pass/fail semantics**: `_closed` is set under `_host_handlers_cv` (a `threading.Condition`) in the production code, but here it is read bare without that lock. While CPython's GIL makes this atomically safe in practice, it is technically a data race.

More critically, this busy-wait loop exits after 2 s even if the reader thread hasn't set `_closed` yet. In that case the subsequent `request()` call will still raise `transport_closed` — but via the `poll()` check at the top of `request()`, not via `_abort_pending`. This means the test can pass even if the reader-thread EOF→waiter notification path (`_abort_pending`) is entirely broken, defeating the purpose of the test.

Consider joining the reader thread directly after triggering EOF to ensure it has completed before asserting:
```python
client._reader.join(timeout=2.0)  # wait for reader to finish
self.assertFalse(client._reader.is_alive(), "reader thread did not exit")
```
This makes the test deterministic and specifically exercises the `_abort_pending` path.

---

### 11. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L379-L384)

**Racy synchronization via `time.sleep(0.05)`**: This sleep is an unreliable heuristic — on a slow or heavily loaded CI machine the 50 ms may elapse before the worker thread has written the request and parked it in `_pending`. In that case `_abort_pending` fires against an empty `_pending` dict (clearing nothing), the worker then parks and waits the full 10 s timeout, and `worker.join(timeout=2.0)` returns with the worker still alive — causing a spurious test failure.

The same issue exists in `test_close_wakes_pending_waiters` (line 435).

A deterministic alternative is to synchronize on the moment the request arrives at the fake stdin. The `FakePopen` stdin write path already accumulates lines in `_stdin_lines`; a small `threading.Event` patched into that path would let the test wait reliably:
```python
request_received = threading.Event()
original_write = self._fake.stdin.write
self._fake.stdin.write = lambda line: (original_write(line), request_received.set())
worker.start()
request_received.wait(timeout=2.0)  # deterministic
self._fake.stdout._done = True
...
```

---

### 12. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L409-L411)

**Busy-polling `client._closed` without a lock, with ambiguous pass semantics**: `_closed` is written under `_host_handlers_cv` (a `threading.Condition`) in the production code, but is read here without holding that lock — a technical data race (safe under CPython's GIL but fragile).

More importantly, this busy-wait exits after 2 s even if the reader thread hasn't set `_closed` yet. In that timeout case, the subsequent `request()` call will *still* raise `transport_closed` — but via the `poll()` check at the top of `request()`, **not** via `_abort_pending`. This means the test can pass even if the reader-thread EOF→waiter notification path (`_abort_pending`) is entirely broken, defeating the test's stated purpose.

A robust fix is to join the reader thread directly after triggering EOF:
```python
self._fake.stdout._done = True
self._fake.stdout._event.set()
client._reader.join(timeout=2.0)  # wait for reader to finish
self.assertFalse(client._reader.is_alive(), "reader thread did not exit")
# Now _closed is guaranteed to be set via _abort_pending
```

PR #2056, opened 2026-07-03
fix: reraise timed status errors without fallback

### LINKED ISSUE #1523: [Bug] timed_with_status decorator silently swallows exceptions and returns None

RerankerGuo:
## Description

Fixes #1523.

`timed_with_status` now re-raises the original exception when no fallback callback is configured. The fallback path still returns the fallback value, while the no-fallback path no longer silently returns `None` after logging a failed status.

Related Issue (Required): Fixes #1523

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Static regression check passed: the fallback branch still returns the fallback result, and the no-fallback branch contains an explicit `raise`.
- [x] `python3 -m py_compile src/memos/utils.py tests/test_utils_timing.py` passed.
- [ ] `PYTHONPATH=src python3 -m pytest tests/test_utils_timing.py -q` could not run because `pytest` is not installed: `No module named pytest`.
- [ ] `make format` could not run because `poetry` is not installed: `make: poetry: Command not found`.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have made corresponding changes to the documentation, if applicable
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/src/memos/utils.py
+++ b/src/memos/utils.py
@@ -179,6 +179,7 @@ def timed_with_status(
                 if fallback is not None and callable(fallback):
                     result = fallback(e, *args, **kwargs)
                     return result
+                raise
             finally:
                 elapsed_ms = (time.perf_counter() - start) * 1000.0
 
--- a/tests/test_utils_timing.py
+++ b/tests/test_utils_timing.py
@@ -318,7 +318,8 @@ class TestTimedWithStatusRegression:
             raise RuntimeError("bad")
 
         with caplog.at_level(logging.INFO):
-            fail_func()
+            with pytest.raises(RuntimeError, match="bad"):
+                fail_func()
         logs = _collect_timer_with_status_logs(caplog)
         assert len(logs) == 1
         assert "status: FAILED" in logs[0]


RerankerGuo:
Closing this draft because current main already re-raises timed_with_status failures when no fallback is configured and includes matching regression coverage.

PR #2064, opened 2026-07-06
fix(skill-crystallize): retry with context when LLM returns empty response

### LINKED ISSUE #2062: feat: skill crystallization retry with context — pass previous failure output to LLM on retry
## Problem

When `crystallizeDraft()` calls the LLM to promote an L2 policy into a Skill, the LLM sometimes returns an empty response (or malformed JSON). The error is logged as `skill.crystallize.failed` and the policy is skipped via `continue`.

On the next trigger event (`reward.updated`, `l2.policy.induced`, etc.), the same policy is re-evaluated and crystallization is attempted **from scratch** — the same prompt, the same LLM call, with no memory of the previous failure.

If the underlying issue (e.g. empty response for a particular prompt shape) persists, the policy enters an infinite retry loop with no progress.

## Suggestion

Implement a **retry-with-context** mechanism for failed crystallization attempts:

1. **On failure**: store the LLM's raw output (even if empty/malformed) as a temporary artifact linked to the policy.
2. **On subsequent retry**: inject the previous attempt as additional context in the LLM prompt — e.g. "The previous attempt produced the following output: ... Please correct this and generate a valid skill definition."
3. **On success**: clear the temporary artifact.

This mirrors the pattern used elsewhere in MemOS (e.g. `LlmClient.completeJson` already retries internally with `maxMalformedRetries`), but applies it at the orchestrator level so the context of *why* the previous attempt failed is visible to the model.

### Why it matters

- Without this, a policy that consistently triggers empty LLM responses will be retried infinitely on every pipeline cycle with zero signal for the model to adjust.
- With context, even a single retry can succeed if the model sees its own blank output and corrects course.
- The change is scoped to `crystallizeDraft()` and/or `runSkill()` — no upstream API changes needed.

### Implementation sketch

1. Add a `failureArtifact` KV store keyed by `policyId` in the skill repos (or reuse `kv` table).
2. In `crystallizeDraft()` catch block, save `{ raw, error, timestamp }` to the artifact.
3. In `packPrompt()`, check for existing artifact and inject a "correction" block to the LLM prompt.
4. In the success path, clear the artifact.

Would be happy to discuss further or submit a PR if the direction is agreed on.

34262315716:
## Problem

When `crystallizeDraft()` receives an empty response from the LLM (e.g., due to transient model issues or rate limiting), the function immediately fails with a generic `llm-failed` error. There is no retry mechanism and no diagnostic context, making it difficult to recover or debug.

## Changes

- Added a retry-with-correction-context block in the catch handler of `crystallizeDraft()`
- When `rawPreview` is available and the LLM call throws, the retry sends the original raw output plus the error message back to the LLM with a correction instruction
- If the retry also fails (including model refusal), the function returns a descriptive failure instead of crashing
- The retry preserves the same `llm.completeJson` schema hint (`skill-crystallize.v2`) for consistent parsing

## Related

Closes #2062

### DIFF
--- a/apps/memos-local-plugin/core/skill/crystallize.ts
+++ b/apps/memos-local-plugin/core/skill/crystallize.ts
@@ -176,7 +176,7 @@ export async function crystallizeDraft(
     }
 
     // ── retry with correction context if raw output is available ──
-    if (rawPreview) {
+    if (rawPreview !== null) {
       log.warn("skill.crystallize.retry", {
         policyId: input.policy.id,
         error: message,
--- a/apps/memos-local-plugin/core/skill/crystallize.ts
+++ b/apps/memos-local-plugin/core/skill/crystallize.ts
@@ -235,6 +235,10 @@ The error was: ${message}. Please correct this and generate a valid JSON skill d
           return { ok: false, skippedReason: "llm-refusal", modelRefusal };
         }
         if (deps.validate) deps.validate(draft);
+        log.warn("skill.crystallize.retry_succeeded", {
+          policyId: input.policy.id,
+          error: message,
+        });
         return { ok: true, draft };
       } catch (retryErr) {
         const retryMsg = retryErr instanceof Error ? retryErr.message : String(retryErr);
--- a/apps/memos-local-plugin/core/skill/crystallize.ts
+++ b/apps/memos-local-plugin/core/skill/crystallize.ts
@@ -9,7 +9,7 @@
  * traces we fail fast with `skipped_reason="no-evidence"`.
  */
 
-import type { LlmClient } from "../llm/types.js";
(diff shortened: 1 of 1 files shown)


Memtensor-AI:
Routing this PR to @syzsunshine219 for review/follow-up.

Memtensor-AI:

**Details:** A migrator test failed with 'table traces has no column named role', indicating a schema/migration mismatch unrelated to the diff. The diff only modifies crystallize.ts (skill retry logic) and does not touch storage, migrations, or the traces schema. [advisory, non-gating] AI-generated tests on branch test/auto-gen-73ac8e5c46e47e18-20260708215802: 32/32 passed — these do NOT affect the PR verdict; review the branch manually.
**Branch:** `fix/skill-crystallize-retry-with-context`

PR #2077, opened 2026-07-08
Fix #2076: local-plugin v2.0.8: gateway CPU 100% — synchronous full-table vector scan (scan

### LINKED ISSUE #2076: local-plugin v2.0.8: gateway CPU 100% — synchronous full-table vector scan (scanAndTopK) + unbounded trace re-insertion from dedup pagination cap
## Summary

Running the OpenClaw local memory plugin (`@memtensor/memos-local-plugin` v2.0.8), the gateway `node` process pinned **one CPU core at 100% for 40+ minutes** on startup, holding **4.2 GB RSS**. Investigation traced this to **two distinct bugs in the plugin**:

1. **Vector retrieval does a synchronous O(n) full-table scan on the Node main thread**, decoding every vector BLOB and computing cosine in-process — this starves the event loop (100% CPU, GC/IO threads idle).
2. **A pagination cap in the capture-side dedup read causes the same trace to be re-inserted indefinitely**, bloating the `traces` table (in our case **518,375 rows of which 437,792 — 84% — were exact duplicates**), which is what made bug #1 catastrophic.

Together: bug #2 grows the table without bound → bug #1 loads the whole thing synchronously on every dreaming/retrieval cycle → CPU melts.

## Environment

- `@memtensor/memos-local-plugin` **v2.0.8** (`MemOS Local Memory (V7)`)
- Node 26, better-sqlite3
- `dreaming.enabled: true`, `embedding.enabled: false`
- `memos.db` had grown to **6.8 GB** (`traces` table alone = 5.96 GB, ~4.2 GB of that in `vec_summary`/`vec_action` BLOBs)

## Evidence — native stack of the 100% CPU main thread (gdb)

```
#0  pread64
#1  unixRead                         (better_sqlite3.node)
#2  accessPayload                    (better_sqlite3.node)
#3  vdbeColumnFromOverflow           (better_sqlite3.node)
#4  sqlite3VdbeExec
#5  sqlite3_step
#6  Statement::JS_all(...)           <-- synchronous .all()
#7  Builtins_CallApiCallbackOptimizedNoProfiling  (libnode)
```

`strace -c` over 3s on the main thread: **56,637 `pread64` calls**, 99.97% of time — i.e. the thread is doing nothing but synchronously reading BLOB overflow pages off a multi-GB table.

## Bug #1 — `scanAndTopK` blocks the event loop

`dist/core/storage/vector.js:149`

```js
export function scanAndTopK(db, table, selectExtra, query, k, opts) {
    const { vecColumn, norm2Column, where, params, hardCap } = opts;
    const cols = ["id", vecColumn, ...(norm2Column ? [norm2Column] : []), ...selectExtra];
    const sql = [
        `SELECT ${cols.join(", ")} FROM ${table}`,
        where ? `WHERE ${where}` : "",
        `LIMIT ${hardCap ?? 100000}`,          // default cap = 100,000 rows
    ].filter(Boolean).join(" ");
    const rows = db.prepare(sql).all(params);  // <-- loads up to 100k rows w/ vector BLOBs synchronously
    const decoded = [];
    for (const r of rows) {
        const vec = decodeVector(r[vecColumn]); // <-- decode every BLOB on main thread
        ...
    }
    return topKCosine(query, decoded, k);       // <-- cosine over all of them on main thread
}
```

Called from `dist/core/storage/repos/traces.js:255` (`searchByVector`).

Problems:
- No ANN/vector index — retrieval is a brute-force O(n) scan over the entire table.
- `.all()` materializes up to `hardCap ?? 100000` rows, each carrying multi-KB vector BLOBs, into JS memory in one synchronous call (hence 4.2 GB RSS and the `JS_all` stack above).
- Decode + cosine also run synchronously on the main thread.

With a table in the hundreds-of-thousands of rows, every dreaming/retrieval cycle blocks the event loop long enough to peg a core and make the gateway unresponsive.

**Suggested fixes:** add a real vector index (e.g. `sqlite-vec`), or at minimum (a) stream/iterate instead of `.all()`, (b) do decode+cosine off the main thread (worker), (c) pre-filter candidates by cheap SQL predicates before loading BLOBs, (d) make the default `hardCap` much smaller and explicit.

## Bug #2 — dedup pagination cap → unbounded re-insertion

Documented in a code comment at `dist/core/storage/repos/traces.js` (`listAllForEpisode`):

```js
/**
 * Full episode-scoped trace fetch with NO pagination cap. The paginated
 * `list({ episodeId })` path silently truncates to 500 rows (default 50),
 * which breaks capture-side dedup when an episode has more than the cap
 * worth of steps — the next runLite/runReflect re-inserts everything past
 * the cap as "novel". Use this for any dedup / reconciliation read.
 */
```

If any code path still feeds dedup from the paginated `list({ episodeId })` (cap 500) instead of `listAllForEpisode`, then for any episode with >500 steps the dedup check cannot see the older traces, so `runLite`/`runReflect` treats them as novel and **re-inserts them every cycle**.

Observed in our DB — the damage is exactly this shape:

- `traces`: **518,375 rows**, but only **80,583** distinct by `(episode_id, turn_id, user_text, agent_text, tool_calls_json)` → **437,792 exact duplicates (84%)**.
- Duplicates concentrated in a few large episodes, e.g. one episode had **49,402 rows but only 153 distinct `turn_id`s / 588 distinct contents** — i.e. the same turns re-inserted hundreds of times.
- Duplicates span the full lifetime of the DB (2026-05-19 → 2026-07-08), consistent with a per-cycle re-insertion loop rather than a one-off.

**Suggested fix:** ensure all capture-side dedup/reconciliation reads use the uncapped `listAllForEpisode` (or a keyset-paginated full read), and/or add a `UNIQUE` constraint / `INSERT … ON CONFLICT DO NOTHING` on the natural trace identity so re-insertion is idempotent.

## Impact

- Gateway CPU pinned at 100%, unresponsive, 4.2 GB RSS.
- DB bloated 8× (≈850 MB → 6.8 GB) purely from duplicate rows + their vectors.

## Workaround we applied (data-side only — does not fix the plugin bugs)

Stopped the gateway, backed up, de-duplicated `traces` by full content identity (518,375 → 80,583 rows), rebuilt the `traces_fts` FTS5 index, `VACUUM`ed (6.8 GB → 1.3 GB), restarted. CPU settled to ~11%. But both root causes are in plugin code, so this will recur as data grows again.

One operational note that bit us during cleanup: `traces` has trigram FTS5 triggers (`traces_fts_ai/ad/au`), so a row-by-row `DELETE` of 437k rows is pathologically slow (didn't finish in 50 min). Dropping the triggers, doing the bulk delete, then rebuilding the FTS content and recreating the triggers was orders of magnitude faster. Not a bug per se, but worth being aware of for any maintenance tooling.

Happy to provide more detail or test a fix.

Memtensor-AI:
## Description

Fixed two coupled root causes in @memtensor/memos-local-plugin that made the gateway node process pin one CPU core at 100% with 4.2 GB RSS on startup (#2076).

Bug #1 — synchronous full-table vector scan in core/storage/vector.ts. `scanAndTopK` used `db.prepare(sql).all(params)`, loading up to `hardCap` (default 100_000) multi-KB vector BLOBs synchronously into JS memory. Fixed by streaming via `.iterate()` and maintaining the top-K min-heap on hits directly (only k entries + one just-decoded vector live in memory at a time), and lowering the default `hardCap` to 5_000 as a safer default (all production callers pass their own value explicitly). Peak RSS during a scan drops from O(hardCap × dim) to O(k × dim).

Bug #2 — dedup pagination cap in core/capture/capture.ts. Four dedup call sites (`runLite` / `runLightweight` / `runReflect` / `persistRows`) used the paginated `tracesRepo.list({ episodeId })`, which silently caps at 500 rows. Once an episode exceeded 500 traces the older rows became invisible to dedup and were re-inserted every cycle (the reporter's DB had 518_375 rows of which 84% were exact duplicates). Fixed by adding `tracesRepo.listAllForEpisode(episodeId)`, an uncapped read ordered by `ts ASC`, and switching the four dedup call sites to it. All other paginated `list({ episodeId, limit })` callers keep their explicit page-size contract.

Tests: 7 new regression tests (traces-listall.test.ts: 4 cases covering 750-row episode dedup, episode scoping, empty episode, and ts-ordering; vector-stream.test.ts: 3 cases covering top-K correctness parity, streaming under a large cap, and explicit-hardCap truncation). Full plugin unit suite: 1128 pass, 1 skipped. The 2 remaining failures (startup-recovery source-string check, migrator schema-drift `traces has no column named role`) are pre-existing on base branch dev-v2.0.23 and unrelated to this fix. Type check clean.

Related Issue (Required):  Fixes #2076

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Automated tests are pending.

- [ ] Unit Test

## Checklist

## Reviewer Checklist
- [x] closes #2076
- [ ] Made sure Checks passed

### DIFF
--- a/apps/memos-local-plugin/core/capture/capture.ts
+++ b/apps/memos-local-plugin/core/capture/capture.ts
@@ -133,12 +133,17 @@ export function createCaptureRunner(deps: CaptureDeps): CaptureRunner {
     // ─── Extract + dedup (skip steps we've already written this episode) ──
     const extractStart = now();
     const rawAll = extractSteps(input.episode);
-    // #2076: MUST use listAllForEpisode (uncapped). The paginated `list`
-    // path silently truncates to 500 rows, which breaks dedup once an
-    // episode grows past that and causes the tail to be re-inserted every
-    // cycle (bloating `traces` unboundedly + starving the vector scan).
-    const existingTraces = deps.tracesRepo.listAllForEpisode(input.episode.id);
-    const seenTs = new Set<number>(existingTraces.map((t) => t.ts));
+    // #2076: MUST use listDedupRowsForEpisode (uncapped, streaming, no
+    // BLOB projection). The paginated `list` path silently truncates to
+    // 500 rows, which breaks dedup once an episode grows past that and
+    // causes the tail to be re-inserted every cycle (bloating `traces`
+    // unboundedly + starving the vector scan). Using the narrow-column
+    // dedup projection here also keeps peak RSS proportional to scalar
+    // fields, not embedding footprint (open code review on #2077).
+    const existingDedupRows = deps.tracesRepo.listDedupRowsForEpisode(input.episode.id);
+    const seenTs = new Set<number>(existingDedupRows.map((t) => t.ts));
+    // Reused later by persistRows so we don't scan the episode twice.
+    const seenSignatures = new Set(existingDedupRows.map(traceIdentitySignature));
     const raw = rawAll.filter((s) => !seenTs.has(s.ts));
     const extractMs = now() - extractStart;
     log.debug("stage.extract.done", {
@@ -187,7 +192,7 @@ export function createCaptureRunner(deps: CaptureDeps): CaptureRunner {
     // Persist as new rows. Reflection / α deliberately empty.
     const persistStart = now();
     const rows = buildRows(scored, summaries, vecs, input.episode);
-    const persisted = await persistRows(rows, input, warnings);
+    const persisted = await persistRows(rows, input, warnings, {}, seenSignatures);
     if (!persisted) {
       // emit capture.failed handled inside persistRows on hard fail.
       return finalResult(
@@ -255,13 +260,16 @@ export function createCaptureRunner(deps: CaptureDeps): CaptureRunner {
 
     const extractStart = now();
     const rawAll = extractSteps(input.episode);
-    // #2076: uncapped dedup read — see runLite for the full rationale.
-    const existingTraces = deps.tracesRepo.listAllForEpisode(input.episode.id);
+    // #2076 + #2077 OCR: uncapped, narrow-projection dedup read.
+    // See `runLite` for the full rationale — one scan, no BLOBs.
+    const existingDedupRows = deps.tracesRepo.listDedupRowsForEpisode(input.episode.id);
     const seenTurnIds = new Set(
-      existingTraces
+      existingDedupRows
         .map((t) => t.turnId)
         .filter((v): v is number => typeof v === "number" && Number.isFinite(v)),
     );
+    // Reused later by persistRows so we don't scan the episode twice.
+    const seenSignatures = new Set(existingDedupRows.map(traceIdentitySignature));
     const rawByTurn = new Map<number, StepCandidate[]>();
     for (const step of rawAll) {
       const turnId = pickTurnId(step.meta, step.ts);
@@ -313,7 +321,7 @@ export function createCaptureRunner(deps: CaptureDeps): CaptureRunner {
     });
     const persisted = await persistRows(rows, input, warnings, {
       skipActionVectorRetry: true,
-    });
+    }, seenSignatures);
     if (!persisted) {
       return finalResult(
         input,
@@ -428,7 +436,12 @@ export function createCaptureRunner(deps: CaptureDeps): CaptureRunner {
       }));
       const { vecs } = await runEmbed(orphanScored, summaries, warnings);
       const orphanRows = buildRows(orphanScored, summaries, vecs, input.episode);
-      await persistRows(orphanRows, input, warnings);
+      // Reuse the `existing` scan we already ran above — the orphan
+      // persistRows only needs the signature Set, and re-running
+      // `listAllForEpisode` here (or its narrow sibling) would be a
+      // duplicate full-episode scan for no additional information.
+      const seenSignatures = new Set(existing.map(traceIdentitySignature));
+      await persistRows(orphanRows, input, warnings, {}, seenSignatures);
       for (const r of orphanRows) traceByTs.set(r.ts, r);
     }
 
@@ -775,13 +788,23 @@ export function createCaptureRunner(deps: CaptureDeps): CaptureRunner {
     input: CaptureInput,
     warnings: CaptureResult["warnings"],
     opts: { skipActionVectorRetry?: boolean } = {},
+    existingSignatures?: Set<string>,
   ): Promise<boolean> {
-    // #2076: uncapped dedup read — every row we ever wrote for this
-    // episode is a legitimate "already inserted" signature to skip.
-    // Using paginated `list` here missed all rows past the 500 cap
-    // and let the duplicate signatures re-insert every cycle.
-    const existingBeforeInsert = deps.tracesRepo.listAllForEpisode(input.episode.id);
-    const seenSignatures = new Set(existingBeforeInsert.map(traceIdentitySignature));
+    // #2076 + #2077 OCR: uncapped, narrow-projection dedup read. The
+    // paginated `list` path missed all rows past the 500 cap and let
+    // duplicate signatures re-insert every cycle. When the caller has
+    // already computed the signature set upstream (runLite /
+    // runLightweight / runReflect all do), skip the second full scan
+    // and reuse it — otherwise fall back to a fresh streaming scan.
+    // We clone the caller-supplied set so the intra-batch dedup below
+    // doesn't leak new signatures back into the caller's Set instance.
+    const seenSignatures = existingSignatures
+      ? new Set(existingSignatures)
+      : new Set(
+          deps.tracesRepo
+            .listDedupRowsForEpisode(input.episode.id)
+            .map(traceIdentitySignature),
+        );
     const uniqueRows = rows.filter((row) => {
       const signature = traceIdentitySignature(row);
       if (seenSignatures.has(signature)) return false;
@@ -902,7 +925,9 @@ export function createCaptureRunner(deps: CaptureDeps): CaptureRunner {
     return ["user", turnId, step.ts, step.userText.trim()].join("\x1f");
   }
 
-  function traceIdentitySignature(row: TraceRow): string {
+  function traceIdentitySignature(
+    row: Pick<TraceRow, "toolCalls" | "turnId" | "ts" | "agentText" | "userText">,
+  ): string {
     const tool = row.toolCalls[0];
     if (tool) {
       const hasRealTiming =
--- a/apps/memos-local-plugin/core/capture/capture.ts
+++ b/apps/memos-local-plugin/core/capture/capture.ts
@@ -133,7 +133,11 @@ export function createCaptureRunner(deps: CaptureDeps): CaptureRunner {
     // ─── Extract + dedup (skip steps we've already written this episode) ──
     const extractStart = now();
     const rawAll = extractSteps(input.episode);
-    const existingTraces = deps.tracesRepo.list({ episodeId: input.episode.id });
+    // #2076: MUST use listAllForEpisode (uncapped). The paginated `list`
+    // path silently truncates to 500 rows, which breaks dedup once an
+    // episode grows past that and causes the tail to be re-inserted every
+    // cycle (bloating `traces` unboundedly + starving the vector scan).
+    const existingTraces = deps.tracesRepo.listAllForEpisode(input.episode.id);
     const seenTs = new Set<number>(existingTraces.map((t) => t.ts));
     const raw = rawAll.filter((s) => !seenTs.has(s.ts));
     const extractMs = now() - extractStart;
@@ -251,7 +255,8 @@ export function createCaptureRunner(deps: CaptureDeps): CaptureRunner {
 
     const extractStart = now();
     const rawAll = extractSteps(input.episode);
-    const existingTraces = deps.tracesRepo.list({ episodeId: input.episode.id });
+    // #2076: uncapped dedup read — see runLite for the full rationale.
+    const existingTraces = deps.tracesRepo.listAllForEpisode(input.episode.id);
     const seenTurnIds = new Set(
       existingTraces
         .map((t) => t.turnId)
@@ -391,7 +396,9 @@ export function createCaptureRunner(deps: CaptureDeps): CaptureRunner {
     // Pair each normalized step with its already-persisted trace row
     // (matched by ts). If runLite was skipped for any step, fall back
     // to a fresh insert path so we don't lose data.
-    const existing = deps.tracesRepo.list({ episodeId: input.episode.id });
+    // #2076: uncapped dedup read — reflect must see every trace we
+    // ever wrote for this episode, not just the newest 500.
+    const existing = deps.tracesRepo.listAllForEpisode(input.episode.id);
     const traceByTs = new Map<number, (typeof existing)[number]>();
     for (const tr of existing) traceByTs.set(tr.ts, tr);
     const orphan = normalized.filter((s) => !traceByTs.has(s.ts));
@@ -769,7 +776,11 @@ export function createCaptureRunner(deps: CaptureDeps): CaptureRunner {
     warnings: CaptureResult["warnings"],
     opts: { skipActionVectorRetry?: boolean } = {},
   ): Promise<boolean> {
-    const existingBeforeInsert = deps.tracesRepo.list({ episodeId: input.episode.id });
+    // #2076: uncapped dedup read — every row we ever wrote for this
+    // episode is a legitimate "already inserted" signature to skip.
+    // Using paginated `list` here missed all rows past the 500 cap
+    // and let the duplicate signatures re-insert every cycle.
+    const existingBeforeInsert = deps.tracesRepo.listAllForEpisode(input.episode.id);
     const seenSignatures = new Set(existingBeforeInsert.map(traceIdentitySignature));
     const uniqueRows = rows.filter((row) => {
       const signature = traceIdentitySignature(row);
--- a/apps/memos-local-plugin/core/capture/capture.ts
+++ b/apps/memos-local-plugin/core/capture/capture.ts
@@ -404,9 +404,7 @@ export function createCaptureRunner(deps: CaptureDeps): CaptureRunner {
     // Pair each normalized step with its already-persisted trace row
     // (matched by ts). If runLite was skipped for any step, fall back
     // to a fresh insert path so we don't lose data.
-    // #2076: uncapped dedup read — reflect must see every trace we
-    // ever wrote for this episode, not just the newest 500.
-    const existing = deps.tracesRepo.listAllForEpisode(input.episode.id);
+    const existing = deps.tracesRepo.list({ episodeId: input.episode.id });
     const traceByTs = new Map<number, (typeof existing)[number]>();
     for (const tr of existing) traceByTs.set(tr.ts, tr);
     const orphan = normalized.filter((s) => !traceByTs.has(s.ts));
@@ -436,12 +434,7 @@ export function createCaptureRunner(deps: CaptureDeps): CaptureRunner {
       }));
       const { vecs } = await runEmbed(orphanScored, summaries, warnings);
       const orphanRows = buildRows(orphanScored, summaries, vecs, input.episode);
-      // Reuse the `existing` scan we already ran above — the orphan
-      // persistRows only needs the signature Set, and re-running
-      // `listAllForEpisode` here (or its narrow sibling) would be a
-      // duplicate full-episode scan for no additional information.
-      const seenSignatures = new Set(existing.map(traceIdentitySignature));
-      await persistRows(orphanRows, input, warnings, {}, seenSignatures);
+      await persistRows(orphanRows, input, warnings);
       for (const r of orphanRows) traceByTs.set(r.ts, r);
     }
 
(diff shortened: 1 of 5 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2077
**Task:** `91b3e78270801f79`
**Base:** `dev-v2.0.23`
**Head:** `bugfix/autodev-2076-20260708092153967`
**Head SHA:** `1b2dfca0c43a6602bab7d082272d51dfa32d6ac6`

---

### 1. `apps/memos-local-plugin/core/storage/repos/traces.ts` (L249-L258)

**Memory concern**: `listAllForEpisode` selects all 25 columns via the full `COLUMNS` list — including the `vec_summary` and `vec_action` embedding BLOB columns — and materialises every matching row into a single JS array via `.all()`. However, every call-site in `capture.ts` only needs the dedup identity fields (`ts`, `turn_id`, `user_text`, `agent_text`, `tool_calls_json`).

Loading all BLOB columns synchronously for an entire large episode is the same root class of problem that caused the 4.2 GB RSS spike fixed in `vector.ts`, where the solution was precisely to switch from `.all()` to `.iterate()` with a narrow column projection.

Recommendations:
1. **Project only the dedup columns** instead of the full `COLUMNS` list to avoid loading BLOBs at all.
2. **Use `.iterate()` over `.all()`** to stream one row at a time rather than materialising the full result set at once, consistent with the `vector.ts` streaming fix.

Example:
```ts
const DEDUP_COLS = ['ts', 'turn_id', 'user_text', 'agent_text', 'tool_calls_json'] as const;
const sql = `SELECT ${DEDUP_COLS.join(", ")} FROM traces WHERE episode_id = @episode_id ORDER BY ts ASC`;
const result: TraceRow[] = [];
for (const r of db.prepare<{ episode_id: string }, RawTraceRow>(sql).iterate({ episode_id: String(episodeId) })) {
  result.push(mapDedupRow(r));
}
return result;
```
Peak RSS would then be proportional to scalar fields only, not to the total size of all stored embeddings.

**Before:**
```
    listAllForEpisode(episodeId: EpisodeId | string): TraceRow[] {
      if (!episodeId) return [];
      const sql = `SELECT ${COLUMNS.join(
        ", ",
      )} FROM traces WHERE episode_id = @episode_id ORDER BY ts ASC`;
      const rows = db
        .prepare<{ episode_id: string }, RawTraceRow>(sql)
        .all({ episode_id: String(episodeId) });
      return rows.map(mapRow);
    },
```

**After:**
```suggestion
    listAllForEpisode(episodeId: EpisodeId | string): TraceRow[] {
      if (!episodeId) return [];
      // Only project columns needed for dedup — avoids loading
      // vec_summary / vec_action BLOBs that callers never inspect.
      const DEDUP_COLS = ['ts', 'turn_id', 'user_text', 'agent_text', 'tool_calls_json'] as const;
      const sql = `SELECT ${DEDUP_COLS.join(", ")} FROM traces WHERE episode_id = @episode_id ORDER BY ts ASC`;
      const result: TraceRow[] = [];
      for (const r of db
        .prepare<{ episode_id: string }, RawTraceRow>(sql)
        .iterate({ episode_id: String(episodeId) })) {
        result.push(mapDedupRow(r));
      }
      return result;
    },
```

</details>

---

### 2. `apps/memos-local-plugin/core/storage/vector.ts` (L255)

The new streaming loop silently `continue`s on dimension mismatches (`vec.length !== query.length`), whereas the old `topKCosine` path emitted a structured `log.warn('search.dim_mismatch', { expected, got, rowId })` for the same case.

This is a behavioural regression: in production, schema drift (re-embedding with a different model dimension) or data corruption will cause the function to silently return fewer than `k` results with no observable signal. The old warning was the primary mechanism for operators to detect such drift.

Suggestion: restore the warning log:
```ts
if (vec.length === 0 || vec.length !== query.length) {
  if (vec.length !== 0) {
    log.warn("search.dim_mismatch", {
      expected: query.length,
      got: vec.length,
      rowId: String(r["id"]),
    });
  }
  continue;
}
```

---

### 3. `apps/memos-local-plugin/core/storage/vector.ts` (L260-L272)

The meta-object construction block is duplicated verbatim in both branches of the `heap.length < k` / `score > heap[0]!.score` conditional. This violates DRY and means future edits to meta projection must be applied in two places, risking divergence.

The existing `pushBounded` helper in this file already encapsulates the heap-push logic. Consider either reusing `pushBounded` (which would require building meta before the call) or extracting meta construction above the branch:
```ts
const meta = selectExtra.length > 0
  ? (Object.fromEntries(selectExtra.map((c) => [c, r[c]])) as TMeta)
  : (undefined as TMeta);
if (heap.length < k) {
  heap.push({ id: String(r["id"]), score, meta });
  siftUp(heap, heap.length - 1);
} else if (score > heap[0]!.score) {
  heap[0] = { id: String(r["id"]), score, meta };
  siftDown(heap, 0);
}
```
Alternatively, the existing `pushBounded` helper already handles both branches correctly and could be reused directly.

---

### 4. `apps/memos-local-plugin/core/capture/capture.ts` (L783-L784)

`persistRows` calls `listAllForEpisode` unconditionally on every invocation, but its callers (`runLite`, `runLightweight`, `runReflect`) have each *already* called `listAllForEpisode` earlier in the same execution path for their own dedup step. This results in **2 full-episode table scans** per capture cycle — with all columns projected, including the large embedding BLOBs (`vec_summary`, `vec_action`) and `tool_calls_json` — which can be very expensive for large episodes and partially re-introduces the CPU pressure that #2076 aimed to fix.

Consider passing the already-fetched `existingTraces` result (or a pre-built signature `Set`) as a parameter to `persistRows`, so the second scan can be avoided:

```ts
async function persistRows(
  rows: TraceRow[],
  input: CaptureInput,
  warnings: CaptureResult["warnings"],
  opts: { skipActionVectorRetry?: boolean } = {},
  existingSignatures?: Set<string>,   // caller may pass its own scan result
): Promise<boolean> {
  const seenSignatures =
    existingSignatures ??
    new Set(deps.tracesRepo.listAllForEpisode(input.episode.id).map(traceIdentitySignature));
  ...
}
```
Callers that already hold `existingTraces` can then pass `new Set(existingTraces.map(traceIdentitySignature))`, eliminating the redundant DB round-trip.

---

### 5. `apps/memos-local-plugin/core/capture/capture.ts` (L140-L142)

The dedup `Set` built here (`seenTs`) is derived from `existingTraces` which fetches all columns including large BLOBs (`vec_summary`, `vec_action`) and `tool_calls_json`. However, only the `ts` field is actually needed at this point. Consider projecting only the required fields in `listAllForEpisode`, or creating a lighter sibling (e.g., `listTsForEpisode`) that only selects `ts` for the timestamp-based dedup check, to reduce memory pressure for very large episodes.

---

🧹 Filtered **15** low-confidence OCR finding(s) before posting/fix-loop (duplicate: 15).

Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2077
**Task:** `91b3e78270801f79`
**Base:** `dev-v2.0.23`
**Head:** `bugfix/autodev-2076-20260708092153967`
**Head SHA:** `3e25bc3372cfd210c43e2464d2482a4747e943a6`

---

### 1. `apps/memos-local-plugin/core/storage/vector.ts` (L278-L281)

A new `buildMeta` closure is allocated on every iteration of the streaming loop (up to `cap` iterations), even though it is only *called* on the subset of rows that beat the heap threshold. Each closure creation allocates a new function object on the JS heap, adding unnecessary GC pressure in the hot path — which is the very thing this refactor aims to reduce.

Because `selectExtra` is a loop-invariant constant, the meta-projection logic can be expressed as a plain function that accepts `r` as a parameter, defined once outside the loop. The two call sites then become straightforward function calls with no per-iteration allocation.

**Before:**
```
    const buildMeta = (): TMeta =>
      selectExtra.length > 0
        ? (Object.fromEntries(selectExtra.map((c) => [c, r[c]])) as TMeta)
        : (undefined as TMeta);
```

**After:**
```suggestion
  // Defined once, outside the loop — `selectExtra` is loop-invariant.
  const buildMeta = (row: ScanRow): TMeta =>
    selectExtra.length > 0
      ? (Object.fromEntries(selectExtra.map((c) => [c, row[c]])) as TMeta)
      : (undefined as TMeta);

  // ...then inside the loop:
  //   heap.push({ id: String(r["id"]), score, meta: buildMeta(r) });
  //   heap[0] = { id: String(r["id"]), score, meta: buildMeta(r) };
```

</details>

---

### 2. `apps/memos-local-plugin/core/storage/repos/traces.ts` (L308-L325)

The JSDoc claims this method "streams via `.iterate()` so peak memory scales with the scalar payload", but the implementation accumulates every row into the `rows: TraceDedupRow[]` array before returning. Using `.iterate()` to fill an array gives the same final heap footprint as calling `.all()` — both load the entire result set into JS memory. For the 518,375-row episodes described in #2076 this still materialises hundreds of thousands of objects at once.

Since all call-sites in `capture.ts` do `.map(...)` on the returned array, they can't lazily consume a generator today. But the JSDoc "streaming" claim is factually wrong and will mislead future maintainers. Either:
1. Change the comment to accurately describe the eager materialisation (the scalar-only projection is the real saving, not streaming), or
2. Change the return type to `IterableIterator<TraceDedupRow>` / a generator and update callers to avoid building the intermediate array.

**Before:**
```
    listDedupRowsForEpisode(episodeId: EpisodeId | string): TraceDedupRow[] {
      if (!episodeId) return [];
      const sql = `SELECT ${DEDUP_COLUMNS.join(
        ", ",
      )} FROM traces WHERE episode_id = @episode_id ORDER BY ts ASC`;
      const stmt = db.prepare<{ episode_id: string }, RawDedupRow>(sql);
      const rows: TraceDedupRow[] = [];
      for (const r of stmt.iterate({ episode_id: String(episodeId) })) {
        rows.push({
          ts: r.ts,
          turnId: r.turn_id,
          userText: r.user_text,
          agentText: r.agent_text,
          toolCalls: fromJsonText<ToolCallDTO[]>(r.tool_calls_json, []),
        });
      }
      return rows;
    },
```

**After:**
```suggestion
    /**
     * Narrow-projection episode fetch for dedup identity.
     *
     * Projects only the five scalar dedup-identity columns — never
     * `vec_summary` / `vec_action` BLOBs — so per-row size is small
     * even though the full result set is materialised into an array.
     * Uses `.iterate()` to avoid an internal `.all()` intermediate
     * allocation, but callers still receive a fully-materialised
     * `TraceDedupRow[]`.
     *
     * Same episode-scoping / `ts ASC` ordering contract as
     * `listAllForEpisode`; use this whenever you don't need the BLOBs.
     */
    listDedupRowsForEpisode(episodeId: EpisodeId | string): TraceDedupRow[] {
      if (!episodeId) return [];
      const sql = `SELECT ${DEDUP_COLUMNS.join(
        ", ",
      )} FROM traces WHERE episode_id = @episode_id ORDER BY ts ASC`;
      const stmt = db.prepare<{ episode_id: string }, RawDedupRow>(sql);
      const rows: TraceDedupRow[] = [];
      for (const r of stmt.iterate({ episode_id: String(episodeId) })) {
        rows.push({
          ts: r.ts,
          turnId: r.turn_id,
          userText: r.user_text,
          agentText: r.agent_text,
          toolCalls: fromJsonText<ToolCallDTO[]>(r.tool_calls_json, []),
        });
      }
      return rows;
    },
```

</details>

---

### 3. `apps/memos-local-plugin/core/storage/repos/traces.ts` (L261-L283)

The JSDoc for `listAllForEpisode` says "All hot fields required by dedup are projected, so a caller that only needs `ts`, `turnId`, or the identity signature can still iterate at full speed without paying the paginated round trips." This is misleading: the method selects ALL columns including `vec_summary` and `vec_action` BLOBs (the root cause of the 4.2 GB OOM in #2076). A developer reading this doc might reach for `listAllForEpisode` for dedup purposes, paying the full BLOB cost. The comment should steer dedup callers explicitly to `listDedupRowsForEpisode`.

**Before:**
```
    /**
     * Full episode-scoped trace fetch with NO pagination cap.
     *
     * The paginated `list({ episodeId })` path silently truncates to
     * `PageOptions.limit`, which caps at 500 by default. That cap
     * breaks capture-side dedup (#2076): when an episode grows past
     * the cap, the next runLite / runReflect only sees the newest
     * 500 rows, treats every older step as "novel", and re-inserts
     * the whole tail every cycle. In the reporter's 4.2 GB / 6.8 GB
     * failure, 518,375 trace rows had shrunk to 80,583 distinct
     * `(episode_id, turn_id, user_text, agent_text, tool_calls_json)`
     * signatures — 84 % duplicates driven by exactly this loop.
     *
     * Use this helper for any dedup / reconciliation read that must
     * see the whole episode. All hot fields required by dedup are
     * projected, so a caller that only needs `ts`, `turnId`, or the
     * identity signature can still iterate at full speed without
     * paying the paginated round trips.
     *
     * Rows are ordered by `ts ASC` so the causal chain matches the
     * order runLite / runReflect built.
     */
    listAllForEpisode(episodeId: EpisodeId | string): TraceRow[] {
```

**After:**
```suggestion
    /**
     * Full episode-scoped trace fetch with NO pagination cap.
     *
     * Fetches ALL columns including the large `vec_summary` and
     * `vec_action` BLOB columns. Only use this when the caller genuinely
     * needs the full `TraceRow` shape (e.g. `runReflect` which re-embeds
     * and updates every field).
     *
     * **For dedup-only reads** — where only `ts`, `turnId`, `userText`,
     * `agentText`, and `toolCalls` are needed — use
     * {@link listDedupRowsForEpisode} instead to avoid loading
     * multi-GB of BLOB data into JS memory.
     *
     * Rows are ordered by `ts ASC` so the causal chain matches the
     * order runLite / runReflect built.
     */
    listAllForEpisode(episodeId: EpisodeId | string): TraceRow[] {
```

</details>

---

### 4. `apps/memos-local-plugin/core/storage/repos/traces.ts` (L283-L287)

The JSDoc for this method says "All hot fields required by dedup are projected, so a caller that only needs `ts`, `turnId`, or the identity signature can still iterate at full speed without paying the paginated round trips." This is misleading: the method selects `${COLUMNS.join(", ")}` which includes `vec_summary` and `vec_action` — the large BLOB columns explicitly identified as the root cause of the 4.2 GB OOM in #2076. A developer reading this comment might reach for `listAllForEpisode` for a dedup-only read and silently pay the full BLOB cost.

The JSDoc should clearly state that all columns (including BLOBs) are fetched, and redirect dedup-only callers to `listDedupRowsForEpisode`.

---

### 5. `apps/memos-local-plugin/core/storage/repos/traces.ts` (L314-L324)

The JSDoc above claims this method "streams via `.iterate()` so peak memory scales with the scalar payload", but the implementation accumulates every row into the `rows: TraceDedupRow[]` array before returning. Using `.iterate()` to fill an array gives the same final heap footprint as `.all()` — both load the entire result set into JS memory at once. For a 518,375-row episode (the #2076 failure case), this still allocates hundreds of thousands of objects synchronously.

The real saving is the narrow column projection (no `vec_summary`/`vec_action` BLOBs), not streaming. The JSDoc should be corrected to avoid misleading future maintainers. If true streaming is desired, change the return type to `IterableIterator<TraceDedupRow>` and update call-sites to consume lazily.

---

🧹 Filtered **8** low-confidence OCR finding(s) before posting/fix-loop (duplicate: 7, existing_code_mismatch: 1).

Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2077
**Task:** `91b3e78270801f79`
**Base:** `dev-v2.0.23`
**Head:** `bugfix/autodev-2076-20260708092153967`
**Head SHA:** `98f779cc482670b347237728d3773f895ab1a380`

---

### 1. `apps/memos-local-plugin/core/storage/vector.ts` (L287-L293)

The streaming loop inlines the min-heap push logic (`siftUp` / `siftDown`) directly instead of calling the existing `pushBounded` helper defined at line 141, which already encapsulates exactly this pattern. This duplication means a future change to heap logic (e.g. tie-breaking policy, score comparison operator) must be updated in two places, creating a silent divergence risk.

Consider reusing `pushBounded`:
```ts
pushBounded(heap, { id: String(r["id"]), score, meta: buildMeta(r) }, k);
```

**Before:**
```
    if (heap.length < k) {
      heap.push({ id: String(r["id"]), score, meta: buildMeta(r) });
      siftUp(heap, heap.length - 1);
    } else if (score > heap[0]!.score) {
      heap[0] = { id: String(r["id"]), score, meta: buildMeta(r) };
      siftDown(heap, 0);
    }
```

**After:**
```suggestion
    pushBounded(heap, { id: String(r["id"]), score, meta: buildMeta(r) }, k);
```

</details>

---

### 2. `apps/memos-local-plugin/core/capture/capture.ts` (L136-L137)

The comment describes `listDedupRowsForEpisode` as 'streaming', but the implementation in `traces.ts` materialises all rows into a `TraceDedupRow[]` array before returning (it uses `stmt.iterate()` internally only to avoid better-sqlite3's intermediate `.all()` allocation, not to return a lazy iterator). The real benefit is the narrow column projection (no BLOB columns), not streaming. Calling this 'streaming' is misleading to future readers and could create false confidence about peak memory usage for very large episodes. Consider replacing 'streaming' with 'uncapped narrow-projection' or simply 'fully scanned' to match the actual behaviour documented in the `traces.ts` JSDoc.

Memtensor-AI:
## ⚠️ Open Code Review automatic OCR fix limit reached

Open Code Review still found **2** issue(s), but the automatic OCR fix loop has reached **2/2** attempts.

I stopped auto-fixing this PR to avoid an infinite loop. Please review the latest OCR comment and fix or dismiss the remaining findings manually.

_PR: https://github.com/MemTensor/MemOS/pull/2077_

Memtensor-AI:

All tests passed (25/25 executed). memos_local_plugin/unit: 25/25. Duration: 5s [advisory, non-gating] AI-generated tests on branch test/auto-gen-91b3e78270801f79-20260708230538: 0/58 passed, 58 failed — these do NOT affect the PR verdict; review the branch manually.

**Branch:** `bugfix/autodev-2076-20260708092153967`

Issue #1910, opened 2026-06-11
Bridge process leak: every turn spawns new bridge.cjs, 4+ processes accumulated


haojiajieyangde03-ctrl:
### Environment
- @memtensor/memos-local-plugin v2.0 (Hermes adapter)
- Hermes Agent v0.16.0, macOS 26.5.1
- Config: skillInjectionMode=full, embedding=local, LLM=deepseek-chat

### Problem
Every conversation turn spawns a new `bridge.cjs --agent=hermes --no-viewer` process. Old processes are not cleaned up. Over a session, 4+ bridge processes accumulate:

```
$ ps aux | grep bridge.cjs
cctv  509  0.0%  ~70MB  bridge.cjs --agent=hermes --daemon
cctv  2018 0.0%  ~92MB  bridge.cjs --agent=hermes --no-viewer
cctv  2058 0.0%  ~99MB  bridge.cjs --agent=hermes --no-viewer
cctv  6550 0.0%  ~343MB bridge.cjs --agent=hermes --no-viewer  ← 343MB!
```

The newest process has grown to 343MB RSS. Over longer sessions with many turns, this would lead to significant memory pressure (especially on 16GB macOS devices already constrained by 26.5.1 resource limits).

### Expected behavior
Bridge should be initialized once per session, with subsequent turns reusing the existing connection. Or at minimum, old bridge processes should be cleaned up when a new one spawns.

### Note
This may be related to NousResearch/hermes-agent#20939 which reports the same symptom from the Hermes side: "MemOS memory provider spawns new bridge.cts process on every single turn."

### Impact
- Memory leak: each process consumes 70-340MB RSS
- On long-running sessions (50+ turns), this could consume 3-10GB of RAM
- macOS 26.5.1 already imposes disk write limits; memory pressure from leaked processes compounds this

fancyboi999:
I've reproduced this locally and traced the root cause — claiming this one, fix PR incoming.

**TL;DR**: accumulation of live `--no-viewer` bridges = multiple live `MemosBridgeClient` instances in the Hermes host process. Two compounding sides:

1. **Host side (hermes-agent)**: before NousResearch/hermes-agent#27190, background review / delegate forks rebuilt the memory provider per fork (`load_memory_provider()` constructs a fresh `MemTensorProvider` on every call — module is cached, instance is not), and the abandoned fork never calls `shutdown()`. `oneshot.py` still lacks `skip_memory=True` today.
[...]

PR #2095, opened 2026-07-10
fix(hermes-adapter): read singular traceId from turn.end per core contract

fancyboi999:
## What

`MemTensorProvider._end_turn` reads the plural `traceIds` from the `turn.end` bridge response, but the core contract (`agent-contract/memory-core.ts` `onTurnEnd`, line 205) returns a **singular** `traceId`:

```ts
onTurnEnd(result: TurnResultDTO): Promise<{ traceId: string; episodeId: EpisodeId }>;
```

So against the real core, `_last_trace_id` is always empty and verifier feedback silently falls back to episode scope — trace-level binding never happens.

## Fix

Read the singular `traceId` first; keep the plural `traceIds` form as a fallback for forward compatibility. Pipeline test's `FakeBridge` updated to return the contract-shaped response, plus an assertion that the captured trace id round-trips into feedback submission.

## Provenance

Found via real-bridge E2E while investigating #1910. Originally part of #1911; the process-leak half of that PR was superseded by #2034 (thanks @CarltonXiang — the layered approach there is more robust than mine), so this PR carries only the remaining contract fix, rebased onto current `main`.

## Verification

- `ruff check` / `ruff format --check` clean on both changed files.
- `python -m unittest` over `apps/memos-local-plugin/tests/python/` — **77/77 pass** locally (macOS, Python 3.12).

### Heads-up: unrelated import breakage on current `main`

Running the plugin tests on current `main` fails before this change is even reached: `memos_provider/__init__.py:68` imports `MemosHttpClient` from `bridge_client`, but that class is not defined anywhere in the repo (`git grep "class MemosHttpClient"` → no hits; looks like the `pr/may27-fixes` merge brought in the usages without the class). I verified this PR's 77-test pass with a local throwaway stub for that class; the stub is **not** part of this PR. Filing a separate issue for it.

### DIFF
--- a/apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py
+++ b/apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py
@@ -2111,11 +2111,15 @@ class MemTensorProvider(MemoryProvider):
         if agent_thinking:
             payload["agentThinking"] = agent_thinking
         result = self._bridge.request("turn.end", payload, timeout=_LONG_RPC_TIMEOUT)
-        # Capture the trace ID for feedback submission
+        # Capture the trace ID for feedback submission. The core contract
+        # (memory-core.ts onTurnEnd) returns a singular `traceId`; accept
+        # the plural form too for forward compatibility.
         if result and isinstance(result, dict):
-            trace_ids = result.get("traceIds", [])
-            if trace_ids and len(trace_ids) > 0:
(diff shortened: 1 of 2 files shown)


fancyboi999:
Closing this for now at the author's request to keep the PR queue current.

The implementation branch and passing CI results will remain available. If the maintainers would like to revisit the Hermes trace binding fix, please comment here or reopen the PR. I am happy to rebase it if needed, address feedback, and rerun the full validation matrix.

PR #2099, opened 2026-07-12
fix: pidIsAlive self-PID check for stale lock detection

### LINKED ISSUE #2098: pidIsAlive misidentifies own PID as duplicate runtime when lock file is stale
## Bug Description

When the OpenClaw gateway exits abnormally (e.g. TerminateProcess, power failure, OOM), the openclaw-runtime.lock/ directory and its owner.json remain on disk. On restart, if Windows assigns the same PID to the new process (common due to PID recycling), pidIsAlive() checks the stale owner PID which happens to be the current process's own PID and incorrectly reports that another OpenClaw instance is running.

## Reproduction Steps

1. Start openclaw gateway (gets PID e.g. 12345)
2. Kill the process forcefully (not clean shutdown)
3. Start openclaw gateway again
4. If Windows assigns PID 12345 to the new process, the lock check fails

## Root Cause

Missing self-PID check in pidIsAlive() at apps/memos-local-plugin/adapters/openclaw/runtime-lock.ts:104

The function process.kill(pid, 0) when pid === process.pid always succeeds (you can always signal yourself), so the function returns true even though the lock is from a previous instance.

## Fix

Add one line in pidIsAlive() to exclude the current process PID:

if (pid === process.pid) return false;

## Environment
- memos-local-plugin: 2.0.5
- OpenClaw: 2026.6.10
- OS: Windows 11
- Node: 22.x

34262315716:
Fixes #2098

## Problem
When OpenClaw gateway exits abnormally, the openclaw-runtime.lock/ dir and owner.json remain on disk. On restart, if the OS reuses the same PID, pidIsAlive() signals itself and incorrectly reports duplicate runtime.

## Fix
Added `if (pid === process.pid) return false;` in pidIsAlive() to exclude self-PID from stale lock check. Zero side-effect: owner.pid can only equal process.pid when the lock is from a previous lifecycle of the same process.

## Files changed
- apps/memos-local-plugin/adapters/openclaw/runtime-lock.ts


shinetata (changes requested):
Thanks for the report — the Windows PID-recycle false positive is real.

However, putting `if (pid === process.pid) return false` inside `pidIsAlive()` breaks same-process re-acquire detection (see existing test “rejects a second live owner”). That test both acquires with `process.pid`; after this change the second acquire would reclaim the lock instead of throwing `DuplicateOpenClawRuntimeError`.

Please fix at the acquire site with held-lock tracking instead of changing `pidIsAlive` semantics, and add a regression test for a stale owner whose `pid === process.pid` (recycled PID) that should reclaim successfully. Happy to re-review after that.

syzsunshine219:
Addressed the requested changes in `8da7a110`:

- kept `pidIsAlive` semantics unchanged
- track lock directories actually held by this process
- reclaim an untracked owner with `pid === process.pid` as a recycled stale PID
- preserve same-process duplicate acquire detection and verify tracking cleanup after release

Validation:
- targeted runtime-lock and diagnostic tests: 8/8 passed
- full local-plugin suite: 1303 passed, 2 skipped
- lint and build passed

Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2099
**Task:** `2b4808357c7c23f6`
**Base:** `main`
**Head:** `fix/stale-pid-lock`

---

### 1. `apps/memos-local-plugin/adapters/openclaw/runtime-lock.ts` (L10)

The `heldRuntimeLockDirs` check fires before `owner` is read in the original flow, but here `owner` is read unconditionally on line 104 and then passed to `DuplicateOpenClawRuntimeError` on line 106 when the set already contains `lockDir`. This is fine since the constructor accepts `null`.

However, there is a more subtle problem: `heldRuntimeLockDirs` is module-level (process-global) state, but `acquireOpenClawRuntimeLock` supports an injected `pid` via `options.pid`. If callers pass a synthetic PID (e.g., in tests), the Set still uses the real module reference, so two test invocations sharing the same module instance will interfere with each other. Consider documenting that `heldRuntimeLockDirs` is keyed only by `lockDir`, not by `(lockDir, pid)`, so test isolation requires distinct `home.daemonDir` values per test.

---

### 2. `apps/memos-local-plugin/adapters/openclaw/runtime-lock.ts` (L137-L139)

The `heldRuntimeLockDirs` Set entry is added only after `writeFileSync` succeeds (line 137), but the duplicate-lock guard (line 105) is evaluated inside the `EEXIST` catch block of a `for(;;)` loop. If two synchronous calls to `acquireOpenClawRuntimeLock` are made with the same `lockDir` in the same tick (i.e., the first call's `mkdirSync` succeeds and breaks out, but before `heldRuntimeLockDirs.add` is reached at line 137, the code path is still executing synchronously and cannot be interleaved), this is safe in practice because Node.js JS execution is single-threaded.

However, the guard should ideally be placed immediately after `mkdirSync` succeeds (i.e., before the write) to prevent a second call from observing a lock directory that exists but isn't yet tracked in the Set. As written, a second call that hits `EEXIST` after the first call's `mkdirSync` but before its `heldRuntimeLockDirs.add` would fall through to the PID check instead of the Set check. In practice this cannot happen synchronously, but it is a fragile ordering assumption worth documenting or restructuring.

**Before:**
```
  heldRuntimeLockDirs.add(lockDir);

  let released = false;
```

**After:**
```suggestion
  // Track the held lock immediately so re-entrant calls within the same
  // process are caught by the Set guard before reaching the PID check.
  heldRuntimeLockDirs.add(lockDir);

  let released = false;
```

</details>

syzsunshine219:
Addressed the OCR finding in `c9cf4cb8`: the stale-owner identity check now compares against the normalized `pid` (`options.pid ?? process.pid`), and a regression test covers a live custom owner PID.

Revalidation: targeted 9/9; full local-plugin suite 1304 passed, 2 skipped; lint/build passed.

syzsunshine219:
Reviewed the two follow-up OCR notes; no additional code change is needed:

1. The Set is intentionally keyed by `lockDir`, because the lock protects that directory regardless of an injected owner PID. Tests already create a unique `tmpHome()`/`daemonDir` per case.
2. Acquisition is fully synchronous with no yield between `mkdirSync`, owner-file write, and `heldRuntimeLockDirs.add`. Adding the Set entry before the write would require rollback on `writeFileSync` failure and could otherwise leave a false in-memory owner. The current ordering records ownership only after durable owner metadata is written.

The OCR check itself completed successfully.

Memtensor-AI:

All tests passed (7/7 executed). memos_local_plugin/unit: 7/7. Duration: 2s

**Branch:** `fix/stale-pid-lock`

PR #2129, opened 2026-07-20
ci: automate OpenClaw local plugin release notes

Memtensor-AI:
## Summary

This PR wires the OpenClaw local plugin publish workflow into the Doc Agent release-notes draft flow.

It adds:

- evidence-backed release notes generation for `apps/memos-local-plugin`
- source-ref coverage checks for important local-plugin commits
- deterministic postprocessing for category/source-ref dedupe
- bilingual validation plus bounded LLM repair attempts
- retry handling for transient release workflow phases
- Linux sqlite prebuild installation for validation after `npm ci --ignore-scripts`
- prerelease GitHub Release marking for beta/next/alpha or semver prerelease versions
- a post-merge dry-run workflow that automatically runs the local-plugin workflow on `main` with `dry_run=true`

## Safety

- `dry_run=true` skips npm publish, tag creation, GitHub Release creation, and release PR creation.
- Formal docs sync remains gated on GitHub `release.published`.
- `release.published` prerelease events are skipped on the 106 Doc Agent side before docs PR creation.
- Production deployment is not part of this PR.

## Validation

- Local Node test: `node --test .github/scripts/draft-local-plugin-release-notes.test.mjs` passed, 14/14.
- GitHub Actions dry-run passed on this branch:
  - https://github.com/MemTensor/MemOS/actions/runs/29729728702
- Final inspection artifact summary:
  - `needs_review=false`
  - `missing_required_count=0`
  - `language_issue_count=0`
  - `validation_attempt_count=1`
  - `repair_attempt_count=0`
  - `item_count=8`

## Follow-up after merge

After this draft PR is reviewed and merged, the new post-merge workflow should automatically run the main-branch local-plugin dry-run using:

- `version=2.0.10`
- `tag=latest`
- `dry_run=true`

A real beta rehearsal should wait until there is an actual next release source; do not publish `2.0.11-beta.1` only for testing.

### DIFF
--- a/.github/scripts/draft-local-plugin-release-notes.mjs
+++ b/.github/scripts/draft-local-plugin-release-notes.mjs
@@ -1,0 +1,1353 @@
+#!/usr/bin/env node
+import { execFileSync } from "node:child_process";
+import { mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { tmpdir } from "node:os";
+import { pathToFileURL } from "node:url";
+
+const PRODUCT_PATH = "apps/memos-local-plugin";
+const PRODUCT_ID = "openclaw-local-plugin";
+const PRODUCT_TITLE = {
+  zh: "OpenClaw 本地插件",
+  en: "OpenClaw Local Plugin",
+};
+export const RELEASE_NOTE_GUIDANCE = {
+  category_policy: {
+    Added:
+      "Use for newly exposed user-facing capabilities, configuration entries, model slots, language support, or workflow modes.",
+    Improved:
+      "Use for performance optimization, context formatting, compatibility hardening, startup/link/install robustness, or reliability improvements when the evidence is not primarily a broken-behavior fix.",
+    Fixed:
+      "Use for concrete broken behavior, regressions, data recovery bugs, failed retries, endpoint probing anomalies, or configuration that existed but did not take effect.",
+  },
+  quality_policy: [
+    "Prefer Added / Improved / Fixed sections when the evidence supports all three; omit a section only when evidence is insufficient.",
+    "Do not collapse a new configuration capability and a bug fix in the same subsystem into one Fixed item; keep the added capability and the fixed behavior separate when both are evidenced.",
+    "Map vector scan CPU reductions, XML memory-context boundaries, and Hermes provider startup/link/install compatibility to Improved unless the evidence clearly describes a user-visible regression.",
+    "Keep bullets short, product-facing, and evidence-backed; do not mention internal file names unless they are the feature name users recognize.",
+  ],
+  translation_policy: [
+    "Treat text_cn as the canonical release-note wording first, then translate text_cn into text_en.",
+    "Do not independently invent English facts beyond the Chinese canonical bullet and its source_refs.",
+    "Keep category and source_refs identical between Chinese and English outputs.",
+    "text_cn must contain Chinese text; text_en must not contain Chinese/CJK characters.",
+  ],
+};
+const CURRENT_TAG_PREFIX = "memos-local-plugin-v";
+const TAG_PREFIXES = [CURRENT_TAG_PREFIX, "openclaw-local-plugin-v"];
+const RELEASE_NOTES_MARKER = "doc-agent-release-notes-json";
+const RELEASE_CATEGORY_ORDER = ["Added", "Improved", "Fixed"];
+const MAX_DRAFT_REPAIR_ATTEMPTS = 2;
+const RELEASE_TO_DOC_CATEGORY = {
+  Added: "New Features",
+  Improved: "Improvements",
+  Fixed: "Bug Fixes",
+};
+const CJK_RE = /[\u3040-\u30ff\u3400-\u9fff\uf900-\ufaff]/;
+
+function fail(message) {
+  throw new Error(String(message));
+}
+
+function warn(message) {
+  console.error(`::warning::${message}`);
+}
+
+function sh(args, options = {}) {
+  return execFileSync("git", args, {
+    cwd: process.cwd(),
+    encoding: "utf8",
+    stdio: ["ignore", "pipe", "pipe"],
+    ...options,
+  }).trim();
+}
+
+export function cleanVersion(raw) {
+  const value = String(raw || "").trim();
+  return value.startsWith("v") ? value.slice(1) : value;
+}
+
+export function displayVersion(raw) {
+  const value = cleanVersion(raw);
+  return value ? `v${value}` : "";
+}
+
+export function versionFromTag(tag) {
+  for (const prefix of TAG_PREFIXES) {
+    if (tag.startsWith(prefix)) {
+      return cleanVersion(tag.slice(prefix.length));
+    }
+  }
+  return "";
+}
+
+export function parseSemver(version) {
+  const cleaned = cleanVersion(version);
+  const match = cleaned.match(/^(\d+)\.(\d+)\.(\d+)(?:[-+]([0-9A-Za-z.-]+))?$/);
+  if (!match) return null;
+  return {
+    major: Number(match[1]),
+    minor: Number(match[2]),
+    patch: Number(match[3]),
+    prerelease: match[4] || "",
+  };
+}
+
+export function compareSemver(a, b) {
+  const av = parseSemver(a);
+  const bv = parseSemver(b);
+  if (!av || !bv) return String(a).localeCompare(String(b));
+  for (const key of ["major", "minor", "patch"]) {
+    if (av[key] !== bv[key]) return av[key] - bv[key];
+  }
+  if (av.prerelease === bv.prerelease) return 0;
+  if (!av.prerelease) return 1;
+  if (!bv.prerelease) return -1;
+  return av.prerelease.localeCompare(bv.prerelease);
+}
+
+function gitShowJson(ref, path) {
+  try {
+    return JSON.parse(sh(["show", `${ref}:${path}`]));
+  } catch {
+    return {};
+  }
+}
+
+function readJsonFile(path) {
+  try {
+    return JSON.parse(readFileSync(path, "utf8"));
+  } catch {
+    return {};
+  }
+}
+
+function tagInfo(ref) {
+  const text = sh(["show", "--no-patch", "--format=%H%n%ci%n%s", ref]);
+  const [sha = "", date = "", subject = ""] = text.split("\n");
+  return { tag: ref, sha, date, subject };
+}
+
+function refInfo(ref, tagLabel) {
+  const info = tagInfo(ref);
+  return { ...info, tag: tagLabel || ref, ref };
+}
+
+function refExists(ref) {
+  try {
+    sh(["rev-parse", "--verify", "--quiet", `${ref}^{commit}`]);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+export function resolveCurrentRef(
+  currentTag,
+  { requestedRef = process.env.RELEASE_EVIDENCE_REF || "", refExistsImpl = refExists } = {},
+) {
+  const explicit = String(requestedRef || "").trim();
+  if (explicit) {
+    if (!refExistsImpl(explicit)) {
+      fail(`RELEASE_EVIDENCE_REF does not exist or is not a commit: ${explicit}`);
+    }
+    return explicit;
+  }
+  if (currentTag && refExistsImpl(currentTag)) return currentTag;
+  return "HEAD";
+}
+
+function listProductTags() {
+  try {
+    sh(["fetch", "--tags", "--force", "origin"], { stdio: ["ignore", "ignore", "ignore"] });
+  } catch {
+    warn("Failed to fetch tags from origin; using locally available tags.");
+  }
+
+  const text = sh(["tag", "--list"]);
+  return text
+    .split("\n")
+    .map((tag) => tag.trim())
+    .filter(Boolean)
+    .map((tag) => ({ tag, version: versionFromTag(tag) }))
+    .filter((item) => item.version && parseSemver(item.version));
+}
+
+export function findPreviousTag(targetVersion, currentTag) {
+  const candidates = listProductTags()
+    .filter((item) => item.tag !== currentTag)
+    .filter((item) => compareSemver(item.version, targetVersion) < 0)
+    .sort((a, b) => {
+      const versionOrder = compareSemver(b.version, a.version);
+      if (versionOrder !== 0) return versionOrder;
+      const aPreferred = a.tag.startsWith(CURRENT_TAG_PREFIX) ? 1 : 0;
+      const bPreferred = b.tag.startsWith(CURRENT_TAG_PREFIX) ? 1 : 0;
+      return bPreferred - aPreferred;
+    });
+  return candidates[0]?.tag || "";
+}
+
+function parseCommits(previousTag, currentRef) {
+  const text = sh([
+    "log",
+    "--format=%H%x09%h%x09%s",
+    "--no-merges",
+    `${previousTag}..${currentRef}`,
+    "--",
+    PRODUCT_PATH,
+  ]);
+  return text
+    .split("\n")
+    .filter(Boolean)
+    .map((line) => {
+      const [sha = "", shortSha = "", subject = ""] = line.split("\t");
+      return { sha, short_sha: shortSha, subject };
+    });
+}
+
+function parseChangedFiles(previousTag, currentRef) {
+  const text = sh(["diff", "--name-status", `${previousTag}..${currentRef}`, "--", PRODUCT_PATH]);
+  return text
+    .split("\n")
+    .filter(Boolean)
+    .map((line) => {
+      const parts = line.split("\t");
+      const item = { status: parts[0], path: parts[parts.length - 1] };
+      if (parts.length === 3) item.old_path = parts[1];
+      return item;
+    });
+}
+
+function packageChanges(previousTag, currentRef) {
+  const path = `${PRODUCT_PATH}/package.json`;
+  const before = gitShowJson(previousTag, path);
+  const after = currentRef === "HEAD" ? readJsonFile(path) : gitShowJson(currentRef, path);
+  return ["name", "version"]
+    .filter((field) => before[field] !== after[field])
+    .map((field) => ({ field, before: before[field], after: after[field] }));
+}
+
+function extractPullRequests(commits, repo = process.env.GITHUB_REPOSITORY || "MemTensor/MemOS") {
+  const seen = new Set();
+  for (const commit of commits) {
+    for (const match of String(commit.subject || "").matchAll(/#(\d+)/g)) {
+      seen.add(match[1]);
+    }
+  }
+  return [...seen].map((number) => ({
+    number,
+    url: `https://github.com/${repo}/pull/${number}`,
+  }));
+}
+
+function refsForGuidance(commit) {
+  const refs = [];
+  if (commit.short_sha) refs.push(commit.short_sha);
+  for (const match of String(commit.subject || "").matchAll(/#(\d+)/g)) {
+    const value = `#${match[1]}`;
+    if (!refs.includes(value)) refs.push(value);
+  }
+  return refs;
+}
+
+function categoryHintForSubject(subject) {
+  const value = String(subject || "");
+  const lower = value.toLowerCase();
+  if (
+    lower.startsWith("release:") ||
+    /^fix(\([^)]+\))?:\s*ruff\b/i.test(value) ||
+    /review feedback/i.test(value)
+  ) {
+    return null;
+  }
+  if (/^(feat|feature|add)(\([^)]+\))?:|^add\s+/i.test(value)) {
+    return {
+      category: "Added",
+      reason: "new user-facing capability or configuration support",
+    };
+  }
+  if (
+    /cpu|full-table vector scan|vector scan|xml boundary|memory context with xml|hermes|provider link|bridge package version|startup poll|double-spawn/i.test(
+      value,
+    )
+  ) {
+    return {
+      category: "Improved",
+      reason: "performance, context-boundary, or integration robustness improvement",
+    };
+  }
+  if (/^(perf|performance|refactor)(\([^)]+\))?:/i.test(value)) {
+    return {
+      category: "Improved",
+      reason: "performance or implementation improvement",
+    };
+  }
+  if (
+    /lightweightmemory|skip evolution|endpoint paths|openai_compatible|empty response|error response|dirty-reward|orphan trace|session exemption|loopback/i.test(
+      lower,
+    )
+  ) {
+    return {
+      category: "Fixed",
+      reason: "specific broken behavior, recovery boundary, retry, auth, or endpoint probing fix",
+    };
+  }
+  if (/^(fix|hotfix|bugfix)(\([^)]+\))?:|^fix\s+#\d+/i.test(value)) {
+    return {
+      category: "Fixed",
+      reason: "specific bug fix",
+    };
+  }
+  return null;
+}
+
+export function categoryHintsForCommits(commits) {
+  return commits
+    .map((commit) => {
+      const hint = categoryHintForSubject(commit.subject);
+      const sourceRefs = refsForGuidance(commit);
+      if (!hint || sourceRefs.length === 0) return null;
+      return {
+        ...hint,
+        source_refs: sourceRefs,
+        subject: commit.subject,
+      };
+    })
+    .filter(Boolean);
+}
+
+export function releaseNoteGuidanceForCommits(commits) {
+  return {
+    ...RELEASE_NOTE_GUIDANCE,
+    source_ref_category_hints: categoryHintsForCommits(commits),
+    source_ref_hint_policy:
+      "Treat source_ref_category_hints as advisory classification hints grounded in evidence. " +
+      "Use them to avoid moving performance/compatibility work into Fixed merely because the commit subject starts with fix.",
+  };
+}
+
+export function collectEvidence({ targetVersion, currentTag, previousTag, currentRef = "HEAD" }) {
+  const commits = parseCommits(previousTag, currentRef);
+  const changedFiles = parseChangedFiles(previousTag, currentRef);
+  const diffRange = `${previousTag}..${currentRef}`;
+  const repo = process.env.GITHUB_REPOSITORY || "MemTensor/MemOS";
+  const importantDiff = sh([
+    "diff",
+    "--unified=2",
+    diffRange,
+    "--",
+    PRODUCT_PATH,
+  ]).slice(0, 24000);
+
+  return {
+    product_id: PRODUCT_ID,
+    product_title: PRODUCT_TITLE,
+    release_note_guidance: releaseNoteGuidanceForCommits(commits),
+    repo,
+    previous_tag: previousTag,
+    current_tag: currentTag,
+    current_ref: currentRef,
+    diff_range: diffRange,
+    target_version: displayVersion(targetVersion),
+    git_ref: sh(["rev-parse", "--short=12", currentRef]),
+    previous: tagInfo(previousTag),
+    current: refInfo(currentRef, currentTag),
+    commits,
+    pull_requests: extractPullRequests(commits, repo),
+    changed_files: changedFiles,
+    diff_stat: sh(["diff", "--stat", diffRange, "--", PRODUCT_PATH]),
+    important_diff: {
+      [`${PRODUCT_PATH}/**`]: importantDiff,
+    },
+    package_changes: packageChanges(previousTag, currentRef),
+    test_changes: changedFiles.filter((item) => item.path.includes("/tests/")),
+    docs_changes: changedFiles.filter((item) => item.path.includes("/docs/")),
+  };
+}
+
+export function evidenceForInspection(evidence) {
+  const guidance = evidence?.release_note_guidance || {};
+  const {
+    release_note_guidance: _releaseNoteGuidance,
+    important_diff: _importantDiff,
+    ...publicEvidence
+  } = evidence || {};
+  return {
+    ...publicEvidence,
+    release_note_guidance: {
+      source_ref_category_hints: Array.isArray(guidance.source_ref_category_hints)
+        ? guidance.source_ref_category_hints
+        : [],
+    },
+    redactions: {
+      important_diff: "omitted from public workflow artifacts; sent only to the configured draft service",
+      release_note_prompt_guidance: "omitted from public workflow artifacts",
+    },
+  };
+}
+
+export function draftForInspection(draft) {
+  return {
+    ok: Boolean(draft?.ok),
+    needs_review: Boolean(draft?.needs_review),
+    confidence: draft?.confidence || "",
+    release_items: Array.isArray(draft?.release_items) ? draft.release_items : [],
+    coverage: {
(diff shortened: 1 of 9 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2129
**Task:** `e813b5f074914288`
**Base:** `main`
**Head:** `docs-sync/local-plugin-draft-release-notes`

🔍 **OpenCodeReview** found **30** issue(s) in this PR.

---

### 1. `.github/workflows/memos-local-plugin-post-merge-dry-run.yml` (L14-L16)

These write permissions are over-provisioned for a workflow that always calls the reusable workflow with `dry_run: true`. A dry run never pushes tags, creates GitHub Releases, or opens PRs, so `contents: write` and `pull-requests: write` are unused. Prefer the minimum required permissions — for a dry run that only downloads/uploads artifacts and posts a step summary, `contents: read` is sufficient at the top level.

**Before:**
```
permissions:
  contents: write
  pull-requests: write
```

**After:**
```suggestion
permissions:
  contents: read
```

</details>

---

### 2. `.github/workflows/memos-local-plugin-post-merge-dry-run.yml` (L23)

The version is hardcoded to `"2.0.10"`. As the package version advances, this dry-run will keep testing against a stale version string that may no longer exist as the `latest` npm dist-tag. Consider deriving the version dynamically from the package manifest so the dry-run always reflects the current release candidate:
```yaml
- id: pkg
  run: echo "version=$(node -p "require('./apps/memos-local-plugin/package.json').version")" >> $GITHUB_OUTPUT
```
and then using `${{ steps.pkg.outputs.version }}`.

---

### 3. `.github/workflows/memos-local-plugin-post-merge-dry-run.yml` (L18-L21)

The `dry-run` job (and therefore the entire workflow) has no `timeout-minutes` set. Since it delegates to a reusable workflow that can run multi-platform matrix builds, npm installs, tests, and external service calls, a hung runner could consume resources indefinitely. Add a suitable timeout (e.g. `timeout-minutes: 30`).

**Before:**
```
jobs:
  dry-run:
    uses: ./.github/workflows/memos-local-plugin-publish.yml
```

**After:**
```suggestion
jobs:
  dry-run:
    if: ${{ github.repository == 'MemTensor/MemOS' }}
    timeout-minutes: 30
    uses: ./.github/workflows/memos-local-plugin-publish.yml
```

</details>

---

### 4. `.github/workflows/memos-local-plugin-post-merge-dry-run.yml` (L3-L7)

There is no `concurrency` group defined for this workflow. Multiple rapid pushes to `main` that match the path filters will each queue a run. Additionally, the called workflow (`memos-local-plugin-publish.yml`) already holds a global `concurrency: group: memos-local-plugin-publish` lock, so queued dry-run jobs could silently block or be blocked by an active real publish run. Adding a dedicated concurrency group with `cancel-in-progress: true` ensures only the latest dry-run matters and avoids unintended serialization with the publish workflow.

**Before:**
```
on:
  push:
    branches:
      - main
    paths:
```

**After:**
```suggestion
on:
  push:
    branches:
      - main
    paths:
      - ".github/workflows/memos-local-plugin-publish.yml"
      - ".github/workflows/memos-local-plugin-post-merge-dry-run.yml"
      - ".github/scripts/draft-local-plugin-release-notes.mjs"
      - ".github/scripts/draft-local-plugin-release-notes.test.mjs"
      - ".github/scripts/retry.sh"

concurrency:
  group: memos-local-plugin-post-merge-dry-run
  cancel-in-progress: true
```

</details>

---

### 5. `.github/scripts/retry.sh` (L26-L28)

The `*)` fallback case silently treats any unrecognized argument (including typos like `--attempt` or `--Delay`) as the start of the command, instead of failing. This means a misconfigured call such as `retry.sh --attempt 5 -- my-cmd` will silently execute `--attempt` as the command and fail with a confusing error, rather than producing a clear "unrecognized option" message.

Consider rejecting unrecognized options that begin with `--` explicitly:

**Before:**
```
    *)
      break
      ;;
```

**After:**
```suggestion
    --*)
      echo "::error::retry.sh: unrecognized option '$1'" >&2
      exit 2
      ;;
    *)
      break
      ;;
```

</details>

---

### 6. `.github/scripts/retry.sh` (L37)

The `--attempts` regex `^[1-9][0-9]*$` accepts arbitrarily large integers (e.g., `999`). Combined with the backoff formula `delay * attempt`, a caller setting `RETRY_ATTEMPTS=100` and `RETRY_DELAY_SECONDS=600` would compute a sleep of up to `600 * 99 = 59,400 seconds` (~16.5 hours) on the penultimate attempt, silently consuming CI runner time until the job-level timeout fires. Consider adding an upper-bound check.

**Before:**
```
if ! [[ "${attempts}" =~ ^[1-9][0-9]*$ ]]; then
```

**After:**
```suggestion
if ! [[ "${attempts}" =~ ^[1-9][0-9]*$ ]] || [ "${attempts}" -gt 20 ]; then
  echo "::error::Invalid retry attempts: ${attempts} (must be 1–20)" >&2
  exit 2
fi
```

</details>

---

### 7. `.github/scripts/retry.sh` (L64-L66)

The linear backoff formula `delay * attempt` has no ceiling. While current workflow callers use small defaults (delay=5, attempts≤3, max sleep=10s), callers using environment variables `RETRY_DELAY_SECONDS` or `--delay` with large values can produce unexpectedly long sleeps. Consider capping the computed sleep duration to a reasonable maximum (e.g., 300s).

**Before:**
```
  sleep_seconds=$((delay * attempt))
  echo "::warning::${label} failed with exit code ${status}; retrying in ${sleep_seconds}s." >&2
  sleep "${sleep_seconds}"
```

**After:**
```suggestion
  sleep_seconds=$((delay * attempt))
  max_sleep=300
  if [ "${sleep_seconds}" -gt "${max_sleep}" ]; then
    sleep_seconds=${max_sleep}
  fi
  echo "::warning::${label} failed with exit code ${status}; retrying in ${sleep_seconds}s." >&2
  sleep "${sleep_seconds}"
```

</details>

---

### 8. `.github/workflows/python-release.yml` (L12-L14)

The `deploy` job has no `timeout-minutes` set. Without a timeout, a hung step (e.g., a stalled `poetry install` or a PyPI upload) will consume runner resources indefinitely until GitHub's default 6-hour ceiling is hit. For a publish job that normally completes in a few minutes, a conservative ceiling (e.g., 15–20 minutes) is strongly recommended.

**Before:**
```
    if: startsWith(github.event.release.tag_name, 'v')

    runs-on: ubuntu-latest
```

**After:**
```suggestion
    if: startsWith(github.event.release.tag_name, 'v')

    runs-on: ubuntu-latest
    timeout-minutes: 20
```

</details>

---

### 9. `.github/workflows/python-release.yml` (L24-L26)

Dependencies are installed on every run without any caching. `actions/setup-python` supports built-in Poetry caching via the `cache` input, which can significantly speed up repeated runs.

**Before:**
```
    - name: Install dependencies
      run: |
        poetry install --no-interaction
```

**After:**
```suggestion
    - name: Set up Python
      uses: actions/setup-python@v5
      with:
        python-version: '3.10'
        cache: 'poetry'
    - name: Install dependencies
      run: |
        poetry install --no-interaction
```

</details>

---

### 10. `.github/scripts/draft-local-plugin-release-notes.test.mjs` (L531-L534)

Two issues with this cleanup pattern:

1. **`process.env = previous` does not reliably restore environment state.** In Node.js, `process.env` is a native object with special getter/setter semantics. Assigning a plain object snapshot to `process.env` replaces the reference rather than restoring individual keys. In practice, environment keys set via `Object.assign(process.env, ...)` may persist in the native env after reassignment. The correct approach is to delete added keys and restore changed ones individually:
```js
  } finally {
    for (const key of ['RELEASE_FAILURE_PHASE', 'RELEASE_FAILURE_ATTEMPT_DIR', 'RELEASE_VERSION', 'RELEASE_TAG', 'DOC_AGENT_RELEASE_FAILURE_URL', 'DOC_AGENT_RELEASE_NOTES_DRAFT_TOKEN']) {
      if (key in previous) process.env[key] = previous[key];
      else delete process.env[key];
    }
    rmSync(directory, { recursive: true, force: true });
  }
```

2. **`rmSync` is called after `process.env = previous` in the same `finally` block.** If `rmSync` throws (e.g., permission error), the exception in `finally` will propagate and the `process.env` assignment will still have run — but since assignment is ineffective (see above), any `Object.assign`-set env vars will still leak into subsequent tests. Restore env vars last, or use separate nested `try/finally` blocks.

---

### 11. `.github/scripts/draft-local-plugin-release-notes.test.mjs` (L583-L588)

Same `process.env = previous` anti-pattern: assigning a plain object snapshot to `process.env` does not reliably restore native env state in Node.js. Keys set with `Object.assign(process.env, ...)` should be restored or deleted individually. This pattern repeats across all tests that mutate `process.env` (lines ~535, ~573, ~597, ~631, ~654). If two such tests happen to run concurrently (Node's `--test` runner uses parallelism by default across files), there is also a race condition on shared env var state.

---

### 12. `.github/scripts/draft-local-plugin-release-notes.test.mjs` (L249-L252)

This assertion uses optional chaining on `Array.find()`. If no item with `记忆恢复` in its `text_cn` exists (e.g., after a refactor that renames the rewrite template), the expression evaluates to `undefined`, not `false`. `assert.equal(undefined, false)` with strict mode will **fail** — but for the wrong reason, masking the actual regression.

To make the intent explicit and get a meaningful failure message, find the item first and assert on it directly:
```js
  const memoryItem = processed.release_items.find((item) => item.text_cn.includes("记忆恢复"));
  assert.ok(memoryItem, "expected an item with 记忆恢复 in text_cn");
  assert.equal(memoryItem.text_cn.includes("XML"), false);
```

---

### 13. `.github/scripts/draft-local-plugin-release-notes.test.mjs` (L508-L510)

The test verifies that `repair_attempt` increments correctly but does not assert that `previous_release_items` in each repair request reflects the **most recently failed** draft rather than the original. Since the same `crossedDraft` object is reused for every `requestImpl` return, the repair context always receives the same items regardless of which attempt's output is used. If the implementation incorrectly passes the original first-attempt items instead of the most recent failed items (a plausible regression in `repairContextFromValidation`), this test would not catch it.

Consider asserting that `requests[2].release_notes_repair_context.previous_release_items` equals the items from `requests[1]`'s context (or the draft returned on attempt 2), not just the initial items.

---

### 14. `.github/scripts/draft-local-plugin-release-notes.test.mjs` (L112-L117)

Source refs for `hints[1]` and `hints[3]` are not asserted. In particular, hint[3] comes from the subject `'fix(memos-local-plugin): handle full endpoint paths in openai_compatible probe'` which has no PR number, so its `source_refs` should be exactly `["78ae7a53"]`. If the implementation incorrectly injects extra refs for this commit, it goes undetected.

Add explicit assertions:
```js
  assert.deepEqual(hints[1].source_refs, ["9deb941e", "#1959"]);
  assert.deepEqual(hints[3].source_refs, ["78ae7a53"]);
```

---

### 15. `.github/scripts/draft-local-plugin-release-notes.test.mjs` (L366-L370)

This SHA is 38 hex characters, not the standard 40. While this is likely intentional for brevity, the implementation's `normalizeSourceRef` accepts `/^[a-fA-F0-9]{7,40}$/`, so it will pass normalization. However, any future code path that validates `sha.length === 40` (e.g., when constructing GitHub API URLs from the full SHA) would behave differently for this fixture vs. real 40-character SHAs. Use a proper 40-character SHA for accuracy:
```js
        sha: "abc1234000000000000000000000000000000000",  // 40 chars
```

---

### 16. `.github/scripts/draft-local-plugin-release-notes.mjs` (L1131-L1143)

Both `requiredUrlFromEnv` and `optionalUrlFromEnv` allow plain `http:` URLs. The bearer token and the full git diff (up to 24 KB) are transmitted to these URLs. Permitting cleartext HTTP means both the credential and potentially sensitive diff content can be intercepted. In a CI environment where network traffic may be observable, this should be `https:` only.

Suggestion: change the protocol check to enforce HTTPS exclusively.

**Before:**
```
    if (!/^https?:$/.test(parsed.protocol)) fail(`${name} must be an HTTP(S) URL.`);
  } catch {
    fail(`${name} must be an HTTP(S) URL.`);
  }
  return value;
}

function optionalUrlFromEnv(name) {
  const value = String(process.env[name] || "").trim();
  if (!value) return "";
  try {
    const parsed = new URL(value);
    if (!/^https?:$/.test(parsed.protocol)) fail(`${name} must be an HTTP(S) URL.`);
```

**After:**
```suggestion
    if (parsed.protocol !== "https:") fail(`${name} must be an HTTPS URL.`);
  } catch {
    fail(`${name} must be an HTTPS URL.`);
  }
  return value;
}

function optionalUrlFromEnv(name) {
  const value = String(process.env[name] || "").trim();
  if (!value) return "";
  try {
    const parsed = new URL(value);
    if (parsed.protocol !== "https:") fail(`${name} must be an HTTPS URL.`);
```

</details>

---

### 17. `.github/scripts/draft-local-plugin-release-notes.mjs` (L1187-L1191)

`attemptDir` is read directly from the environment variable `RELEASE_FAILURE_ATTEMPT_DIR` without any path validation. If a misconfigured workflow or an attacker with env-var control sets it to an arbitrary path (e.g., `/proc/self/environ`, `/etc/passwd`, or a relative traversal like `../../.ssh`), `readFileSync(join(attemptDir, '1.log'), 'utf8')` will silently read that file and its contents will be transmitted to the failure-report endpoint via `reportFailure`. The `try/catch` suppresses read errors, so only readable files are exfiltrated—making the attack quiet.

Suggestion: Validate that `attemptDir` is an absolute path that resolves inside the expected runner temp directory (e.g., `RUNNER_TEMP`), or at minimum check that `resolve(attemptDir)` starts with a known safe prefix before using it.

**Before:**
```
  const attempts = [1, 2, 3].map((attempt) => {
    let message = "attempt log is unavailable";
    try { message = readFileSync(join(attemptDir, `${attempt}.log`), "utf8"); } catch {}
    return { error_code: phase.toUpperCase().replace(/[^A-Z0-9]+/g, "_"), message: cleanError(message), retryable: true };
  });
```

**After:**
```suggestion
  const resolvedDir = join(attemptDir);
  const safeBase = process.env.RUNNER_TEMP || tmpdir();
  if (!resolvedDir.startsWith(safeBase + "/") && resolvedDir !== safeBase) {
    fail(`RELEASE_FAILURE_ATTEMPT_DIR must be inside RUNNER_TEMP; got: ${resolvedDir}`);
  }
  const attempts = [1, 2, 3].map((attempt) => {
    let message = "attempt log is unavailable";
    try { message = readFileSync(join(resolvedDir, `${attempt}.log`), "utf8"); } catch {}
    return { error_code: phase.toUpperCase().replace(/[^A-Z0-9]+/g, "_"), message: cleanError(message), retryable: true };
  });
```

</details>

---

### 18. `.github/scripts/draft-local-plugin-release-notes.mjs` (L418-L424)

The multiline heredoc delimiter `__DOC_AGENT_EOF__` is hardcoded. If any `value` written to `GITHUB_OUTPUT` itself contains the literal string `__DOC_AGENT_EOF__` (e.g., a release-notes markdown body that happens to include this string, or a future draft response), GitHub Actions will misparse the output file and truncate or corrupt all subsequent output variables. This is a known footgun with static delimiters.

Suggestion: Generate a random, unguessable delimiter per write (similar to what GitHub's own toolkits do), or at minimum assert that `value` does not contain the delimiter before writing.

**Before:**
```
  const outputPath = process.env.GITHUB_OUTPUT;
  if (!outputPath) return;
  writeFileSync(outputPath, `${name}<<__DOC_AGENT_EOF__\n${value}\n__DOC_AGENT_EOF__\n`, {
    flag: "a",
  });
}
```

**After:**
```suggestion
function appendOutput(name, value) {
  const outputPath = process.env.GITHUB_OUTPUT;
  if (!outputPath) return;
  const delimiter = `__DOC_AGENT_EOF_${Math.random().toString(36).slice(2).toUpperCase()}__`;
  writeFileSync(outputPath, `${name}<<${delimiter}\n${value}\n${delimiter}\n`, {
    flag: "a",
  });
}
```

</details>

---

### 19. `.github/scripts/draft-local-plugin-release-notes.mjs` (L1018-L1033)

`rewriteKnownReleaseItem` is called twice per item: once before deduplication and once after. The second pass operates on text that was already rewritten by the first pass. Since the rewrite rules match against `text_cn`/`text_en` content (in addition to commit subjects), an item rewritten in the first pass to contain e.g. `"XML"` or `"CPU"` could match a *different* rewrite rule in the second pass, silently substituting a different hardcoded bullet. This produces incorrect release notes without any warning.

The first pass should only be used to determine the category (reclassification); the actual text rewrite should happen once, after deduplication, on stable final items.

**Before:**
```
  let items = inputItems.map((item) => {
    const hintedCategory = bestHintCategoryForItem(item, index);
    const category = hintedCategory || item.category;
    if (category !== item.category) reclassifiedItems += 1;
    return rewriteKnownReleaseItem({ ...item, category }, index);
  });

  const deduped = dedupeSourceRefsByBestCategory(items, index);
  items = dedupeReleaseItems(
    deduped.items.map((item) => {
      const hintedCategory = bestHintCategoryForItem(item, index);
      const category = hintedCategory || item.category;
      if (category !== item.category) reclassifiedItems += 1;
      return rewriteKnownReleaseItem({ ...item, category }, index);
    }),
  );
```

**After:**
```suggestion
  let items = inputItems.map((item) => {
    const hintedCategory = bestHintCategoryForItem(item, index);
    const category = hintedCategory || item.category;
    if (category !== item.category) reclassifiedItems += 1;
    return { ...item, category }; // defer rewrite until after deduplication
  });

  const deduped = dedupeSourceRefsByBestCategory(items, index);
  items = dedupeReleaseItems(
    deduped.items.map((item) => {
      const hintedCategory = bestHintCategoryForItem(item, index);
      const category = hintedCategory || item.category;
      if (category !== item.category) reclassifiedItems += 1;
      return rewriteKnownReleaseItem({ ...item, category }, index);
    }),
  );
```

</details>

---

### 20. `.github/scripts/draft-local-plugin-release-notes.mjs` (L858-L866)

`invalid_source_ref` issues are deliberately excluded from `repairableKinds`, so a draft with invalid refs is not sent for repair—it is returned to the caller with `ok=false`. However, the `requestDraft` function calls `fail(message)` when `payload.needs_review` is true, which throws an error caught by the retry `catch` block. If the error is not marked `retryable`, the outer `requestValidatedDraft` loop receives the thrown error from `requestImpl` rather than a structured `finalDraft`. This means `finalDraft` remains `null` for the first iteration and `requestValidatedDraft` returns `null` instead of the last-known `finalDraft`. The caller in `main()` then calls `draft.ok` on `null` and throws a `TypeError`.

The issue occurs because `requestDraft` calls `fail()` (which throws) for the `needs_review` case rather than returning a structured response; `requestValidatedDraft` has no guard against a `null` return from `requestImpl`.

**Before:**
```
  const repairableKinds = new Set(["language", "missing_required_source"]);
  const repairable =
    issues.length > 0 &&
    issues.every((issue) => repairableKinds.has(issue.kind)) &&
    Array.isArray(draft?.release_items) &&
    draft.release_items.length > 0;

  return {
    ok: Boolean(draft?.ok) && !draft?.needs_review && issues.length === 0,
```

**After:**
```suggestion
  // Guard against requestImpl throwing or returning null
  let finalDraft = { ok: false, needs_review: true, release_items: [], validation_report: {}, repair_attempts: [], validation_attempt_count: 0, repair_attempt_count: 0 };

  for (let repairAttempt = 0; repairAttempt <= maxRepairAttempts; repairAttempt += 1) {
```

</details>

---

### 21. `.github/scripts/draft-local-plugin-release-notes.mjs` (L1114-L1122)

The `.slice(0, 600)` truncation is applied *after* all sanitization regex passes. If the raw error string is longer than 600 characters, the sliced tail contains the unsanitized suffix of the original string after whitespace collapsing. For example, a long URL or bearer token near position 590 in the collapsed string could survive in the truncated output. The sanitization regexes should be applied before or the slice should be applied to the input to ensure only already-sanitized content is returned.

**Before:**
```
function cleanError(value) {
  return String(value || "")
    .replace(/Bearer\s+\S+/gi, "Bearer ***")
    .replace(/sk-[A-Za-z0-9_-]+/g, "sk-***")
    .replace(/https?:\/\/[^\s"'<>]+/gi, "https://***")
    .replace(/\b\d{1,3}(?:\.\d{1,3}){3}(?::\d+)?\b/g, "***")
    .replace(/\s+/g, " ")
    .slice(0, 600);
}
```

**After:**
```suggestion
function cleanError(value) {
  return String(value || "")
    .slice(0, 2000)  // bound input length before regex passes
    .replace(/Bearer\s+\S+/gi, "Bearer ***")
    .replace(/sk-[A-Za-z0-9_-]+/g, "sk-***")
    .replace(/https?:\/\/[^\s"'<>]+/gi, "https://***")
    .replace(/\b\d{1,3}(?:\.\d{1,3}){3}(?::\d+)?\b/g, "***")
    .replace(/\s+/g, " ")
    .slice(0, 600);
}
```

</details>

---

### 22. `.github/workflows/memos-local-plugin-publish.yml` (L293-L299)

Boolean comparison is broken for `workflow_call`. When a caller passes `dry_run: true` via `workflow_call`, the GitHub Actions expression engine receives a boolean `true`, so `inputs.dry_run == true` works. However, the `workflow_dispatch` path declares `dry_run` without an explicit `type:` key, meaning it is treated as a string input, and `inputs.dry_run == true` (boolean) will always be `false` — the dry-run guard never fires for manual dispatches, and a live publish runs even when the user selected `dry_run: true` in the UI.

The fix is to use string comparison consistently: `inputs.dry_run == 'true'`. Apply the same correction to the `'Create release tag and PR'` step.

**Before:**
```
      - name: Stop before publish in dry run
        if: ${{ inputs.dry_run == true }}
        run: |
          echo "::notice::dry_run=true; skipped npm publish, tag, GitHub Release, and release PR."

      - name: Publish to npm
        if: ${{ inputs.dry_run != true }}
```

**After:**
```suggestion
      - name: Stop before publish in dry run
        if: ${{ inputs.dry_run == 'true' }}
        run: |
          echo "::notice::dry_run=true; skipped npm publish, tag, GitHub Release, and release PR."

      - name: Publish to npm
        if: ${{ inputs.dry_run != 'true' }}
```

</details>

---

### 23. `.github/workflows/memos-local-plugin-publish.yml` (L545-L549)

The `'Stop before publish in dry run'` step prints a notice but does not exit the job; subsequent steps keep running unless their own `if:` condition also uses the same (broken) `inputs.dry_run != true` expression. If the dry-run guard logic is ever fixed, this step still only emits a notice — it must call `exit 0` (or rely entirely on the per-step `if:` conditions) to actually stop execution. Consider replacing it with a cleaner pattern: set a job output from the `dry_run` input and gate all publish steps on that output.

**Before:**
```
              git tag "${release_tag}"
              bash .github/scripts/retry.sh --label "push release tag" -- git push origin "refs/tags/${release_tag}"
          fi

          create_release_if_missing
```

**After:**
```suggestion
      - name: Stop before publish in dry run
        if: ${{ inputs.dry_run == 'true' }}
        run: |
          echo "::notice::dry_run=true; skipped npm publish, tag, GitHub Release, and release PR."
          exit 0
```

</details>

---

### 24. `.github/workflows/memos-local-plugin-publish.yml` (L381-L384)

The `working-directory` default is `apps/memos-local-plugin` (set in the top-level `defaults.run`). This step has no explicit `working-directory` override, so `../../.github/scripts/draft-local-plugin-release-notes.mjs` resolves correctly. However, this is inconsistent with every other step in the same `publish` job that uses `.github/scripts/retry.sh` (repo-root-relative paths via `working-directory: .` overrides). The two path conventions existing side-by-side are a latent maintenance hazard: if the package is ever moved, the relative path here will silently fail (failure is already suppressed with `|| echo`), leaving the exhausted-retry notification permanently broken without any obvious error.

**Before:**
```
              RELEASE_FAILURE_PHASE=npm-publish \
                RELEASE_FAILURE_ATTEMPT_DIR="${attempt_dir}" \
                node ../../.github/scripts/draft-local-plugin-release-notes.mjs \
                || echo "::warning::Failed to send the exhausted-retry notification."
```

**After:**
```suggestion
              RELEASE_FAILURE_PHASE=npm-publish \
                RELEASE_FAILURE_ATTEMPT_DIR="${attempt_dir}" \
                node .github/scripts/draft-local-plugin-release-notes.mjs \
                || echo "::warning::Failed to send the exhausted-retry notification."
```

</details>

---

### 25. `.github/workflows/memos-local-plugin-publish.yml` (L312-L313)

The `attempt` variable inside `npm_version_exists` is not declared `local`. Bash functions share the calling shell's variable namespace, so when `npm_version_exists` is called inside the outer `for attempt in 1 2 3` publish retry loop (line ~367), the function's loop will clobber the outer `attempt` counter. After the function returns, the outer loop's `attempt` will hold whatever value the inner loop ended on (likely `3`), which can cause the outer loop to skip remaining retry iterations or immediately trigger the three-attempt failure path.

**Before:**
```
          npm_version_exists() {
            for attempt in 1 2 3; do
```

**After:**
```suggestion
          npm_version_exists() {
            local attempt status
            for attempt in 1 2 3; do
```

</details>

---

### 26. `.github/workflows/memos-local-plugin-publish.yml` (L473)

The GitHub Release title hardcodes `"OpenClaw Local Plugin"`. The rest of this workflow (step summaries, artifact names, workflow name, the companion post-merge-dry-run workflow) all use the name `"MemOS local plugin"` / `@memtensor/memos-local-plugin`. `OpenClaw` appears to be a prior product name. This will create confusingly named GitHub Releases that do not match the npm package or repo branding.

**Before:**
```
              --title "OpenClaw Local Plugin v${RELEASE_VERSION}" \
```

**After:**
```suggestion
              --title "MemOS Local Plugin v${RELEASE_VERSION}" \
```

</details>

---

### 27. `.github/workflows/memos-local-plugin-publish.yml` (L418-L437)

When `git ls-remote` returns a non-zero exit on the third attempt, the function calls `exit ${status}`. But this function is always called inside a command substitution: `remote_branch_sha="$(remote_branch_sha ...)"`. `exit` inside `$(...)` only terminates the subshell; the outer shell sees the substitution return an empty string and continues. The caller then treats the empty string as "branch does not exist" and skips the push guard, potentially allowing a force-push to an existing branch. The function needs to signal failure through a side channel (e.g., a temp file or a non-substitution call pattern) rather than relying on `exit`.

**Before:**
```
remote_branch_sha() {
            local branch="$1"
            local out="${RUNNER_TEMP}/memos-local-plugin-remote-branch.txt"
            for attempt in 1 2 3; do
              set +e
              git ls-remote --heads origin "${branch}" >"${out}" 2>&1
              status=$?
              set -e
              if [ "${status}" = 0 ]; then
                awk '{print $1}' "${out}"
                return 0
              fi
              sed -n '1,120p' "${out}"
              if [ "${attempt}" = 3 ]; then
                echo "::error::Failed to check remote branch ${branch} after three attempts."
                exit "${status}"
              fi
              sleep "$((attempt * 5))"
            done
          }
```

**After:**
```suggestion
          remote_branch_sha() {
            local branch="$1"
            local out="${RUNNER_TEMP}/memos-local-plugin-remote-branch.txt"
            local status attempt
            for attempt in 1 2 3; do
              set +e
              git ls-remote --heads origin "${branch}" >"${out}" 2>&1
              status=$?
              set -e
              if [ "${status}" = 0 ]; then
                awk '{print $1}' "${out}"
                return 0
              fi
              sed -n '1,120p' "${out}"
              if [ "${attempt}" = 3 ]; then
                echo "::error::Failed to check remote branch ${branch} after three attempts." >&2
                # Write a sentinel so the caller can detect network failure even from $(...)
                echo "__REMOTE_BRANCH_SHA_ERROR__"
                return 1
              fi
              sleep "$((attempt * 5))"
            done
          }
```

</details>

---

### 28. `.github/workflows/memos-local-plugin-publish.yml` (L76-L88)

`macos-15-intel` is not a standard GitHub-hosted runner label. GitHub's Intel macOS runners are `macos-13` or `macos-12`; the `macos-14` and `macos-15` labels refer to Apple Silicon runners. If `macos-15-intel` is not a registered self-hosted runner label in this organization, the `darwin-x64` matrix leg will queue indefinitely and eventually time out, blocking the entire release (the `publish` job `needs: build-prebuilds`).

---

### 29. `.github/workflows/memos-local-plugin-publish.yml` (L76-L81)

Neither the `build-prebuilds` job nor the `publish` job declares `timeout-minutes`. The `build-prebuilds` matrix includes native compilation steps (`npm rebuild better-sqlite3`) that can hang on a slow or misconfigured runner, and the `publish` job performs multiple npm registry calls and git operations. Without an explicit timeout, a stuck run can consume the full GitHub Actions 6-hour default, blocking all other runs due to the single `concurrency` group (`memos-local-plugin-publish` with `cancel-in-progress: false`).

**Before:**
```
  build-prebuilds:
    strategy:
      matrix:
        include:
          - os: macos-14
            platform: darwin-arm64
```

**After:**
```suggestion
  build-prebuilds:
    timeout-minutes: 30
    strategy:
      matrix:
        include:
          - os: macos-14
            platform: darwin-arm64
```

</details>

---

### 30. `.github/workflows/memos-local-plugin-publish.yml` (L99-L106)

Neither `actions/setup-node` nor a standalone `actions/cache` step caches the npm cache directory. Each of the four matrix runners performs a full `npm ci` download on every run. Adding `cache: 'npm'` to the `actions/setup-node` step would cache `~/.npm` and significantly reduce both install time and exposure to transient npm registry outages.

**Before:**
```
      - name: Install dependencies
        shell: bash
        run: bash ../../.github/scripts/retry.sh --label "npm ci" -- npm ci

      - name: Rebuild for x64 under Rosetta (darwin-x64 only)
        if: matrix.platform == 'darwin-x64'
        shell: bash
        run: bash ../../.github/scripts/retry.sh --label "npm rebuild better-sqlite3 under Rosetta" -- arch -x86_64 npm rebuild better-sqlite3
```

**After:**
```suggestion
      - uses: actions/setup-node@v4
        with:
          node-version: 22
          cache: 'npm'
          cache-dependency-path: apps/memos-local-plugin/package-lock.json
```

</details>

Memtensor-AI:

All tests passed (10/10 executed, 1 skipped). memos_github_open_source/smoke: 1/1, memos_local_plugin/unit: 9 passed, 1 skipped. Duration: 8s [advisory, non-gating] AI-generated tests on branch test/auto-gen-61d7ba4da9694435-20260720180404: 87/87 passed — these do NOT affect the PR verdict; review the branch manually.

**Branch:** `docs-sync/local-plugin-draft-release-notes`

Memtensor-AI:

All tests passed (10/10 executed, 1 skipped). memos_github_open_source/smoke: 1/1, memos_local_plugin/unit: 9 passed, 1 skipped. Duration: 8s

**Branch:** `docs-sync/local-plugin-draft-release-notes`

Memtensor-AI:

All tests passed (10/10 executed, 1 skipped). memos_github_open_source/smoke: 1/1, memos_local_plugin/unit: 9 passed, 1 skipped. Duration: 8s [advisory, non-gating] AI-generated tests on branch test/auto-gen-86974d3efe1a2697-20260721115403: 39/39 passed — these do NOT affect the PR verdict; review the branch manually.

**Branch:** `docs-sync/local-plugin-draft-release-notes`

Memtensor-AI:

All tests passed (10/10 executed, 1 skipped). memos_github_open_source/smoke: 1/1, memos_local_plugin/unit: 9 passed, 1 skipped. Duration: 8s [advisory, non-gating] AI-generated tests on branch test/auto-gen-e813b5f074914288-20260721155342: 57/58 passed, 1 failed — these do NOT affect the PR verdict; review the branch manually.

**Branch:** `docs-sync/local-plugin-draft-release-notes`

PR #2138, opened 2026-07-22
Dev v2.0.25 sdk

Zyntrael:
## Description

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Unit Test

## Checklist

## Reviewer Checklist
- [ ] Made sure Checks passed

### DIFF
--- a/.github/scripts/draft-local-plugin-release-notes.mjs
+++ b/.github/scripts/draft-local-plugin-release-notes.mjs
@@ -1,0 +1,1353 @@
+#!/usr/bin/env node
+import { execFileSync } from "node:child_process";
+import { mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { tmpdir } from "node:os";
+import { pathToFileURL } from "node:url";
+
+const PRODUCT_PATH = "apps/memos-local-plugin";
+const PRODUCT_ID = "openclaw-local-plugin";
+const PRODUCT_TITLE = {
+  zh: "OpenClaw 本地插件",
+  en: "OpenClaw Local Plugin",
+};
+export const RELEASE_NOTE_GUIDANCE = {
+  category_policy: {
+    Added:
+      "Use for newly exposed user-facing capabilities, configuration entries, model slots, language support, or workflow modes.",
+    Improved:
+      "Use for performance optimization, context formatting, compatibility hardening, startup/link/install robustness, or reliability improvements when the evidence is not primarily a broken-behavior fix.",
+    Fixed:
+      "Use for concrete broken behavior, regressions, data recovery bugs, failed retries, endpoint probing anomalies, or configuration that existed but did not take effect.",
+  },
+  quality_policy: [
+    "Prefer Added / Improved / Fixed sections when the evidence supports all three; omit a section only when evidence is insufficient.",
+    "Do not collapse a new configuration capability and a bug fix in the same subsystem into one Fixed item; keep the added capability and the fixed behavior separate when both are evidenced.",
+    "Map vector scan CPU reductions, XML memory-context boundaries, and Hermes provider startup/link/install compatibility to Improved unless the evidence clearly describes a user-visible regression.",
+    "Keep bullets short, product-facing, and evidence-backed; do not mention internal file names unless they are the feature name users recognize.",
+  ],
+  translation_policy: [
+    "Treat text_cn as the canonical release-note wording first, then translate text_cn into text_en.",
+    "Do not independently invent English facts beyond the Chinese canonical bullet and its source_refs.",
+    "Keep category and source_refs identical between Chinese and English outputs.",
+    "text_cn must contain Chinese text; text_en must not contain Chinese/CJK characters.",
+  ],
+};
+const CURRENT_TAG_PREFIX = "memos-local-plugin-v";
+const TAG_PREFIXES = [CURRENT_TAG_PREFIX, "openclaw-local-plugin-v"];
+const RELEASE_NOTES_MARKER = "doc-agent-release-notes-json";
+const RELEASE_CATEGORY_ORDER = ["Added", "Improved", "Fixed"];
+const MAX_DRAFT_REPAIR_ATTEMPTS = 2;
+const RELEASE_TO_DOC_CATEGORY = {
+  Added: "New Features",
+  Improved: "Improvements",
+  Fixed: "Bug Fixes",
+};
+const CJK_RE = /[\u3040-\u30ff\u3400-\u9fff\uf900-\ufaff]/;
+
+function fail(message) {
+  throw new Error(String(message));
+}
+
+function warn(message) {
+  console.error(`::warning::${message}`);
+}
+
+function sh(args, options = {}) {
+  return execFileSync("git", args, {
+    cwd: process.cwd(),
+    encoding: "utf8",
+    stdio: ["ignore", "pipe", "pipe"],
+    ...options,
+  }).trim();
+}
+
+export function cleanVersion(raw) {
+  const value = String(raw || "").trim();
+  return value.startsWith("v") ? value.slice(1) : value;
+}
+
+export function displayVersion(raw) {
+  const value = cleanVersion(raw);
+  return value ? `v${value}` : "";
+}
+
+export function versionFromTag(tag) {
+  for (const prefix of TAG_PREFIXES) {
+    if (tag.startsWith(prefix)) {
+      return cleanVersion(tag.slice(prefix.length));
+    }
+  }
+  return "";
+}
+
+export function parseSemver(version) {
+  const cleaned = cleanVersion(version);
+  const match = cleaned.match(/^(\d+)\.(\d+)\.(\d+)(?:[-+]([0-9A-Za-z.-]+))?$/);
+  if (!match) return null;
+  return {
+    major: Number(match[1]),
+    minor: Number(match[2]),
+    patch: Number(match[3]),
+    prerelease: match[4] || "",
+  };
+}
+
+export function compareSemver(a, b) {
+  const av = parseSemver(a);
+  const bv = parseSemver(b);
+  if (!av || !bv) return String(a).localeCompare(String(b));
+  for (const key of ["major", "minor", "patch"]) {
+    if (av[key] !== bv[key]) return av[key] - bv[key];
+  }
+  if (av.prerelease === bv.prerelease) return 0;
+  if (!av.prerelease) return 1;
+  if (!bv.prerelease) return -1;
+  return av.prerelease.localeCompare(bv.prerelease);
+}
+
+function gitShowJson(ref, path) {
+  try {
+    return JSON.parse(sh(["show", `${ref}:${path}`]));
+  } catch {
+    return {};
+  }
+}
+
+function readJsonFile(path) {
+  try {
+    return JSON.parse(readFileSync(path, "utf8"));
+  } catch {
+    return {};
+  }
+}
+
+function tagInfo(ref) {
+  const text = sh(["show", "--no-patch", "--format=%H%n%ci%n%s", ref]);
+  const [sha = "", date = "", subject = ""] = text.split("\n");
+  return { tag: ref, sha, date, subject };
+}
+
+function refInfo(ref, tagLabel) {
+  const info = tagInfo(ref);
+  return { ...info, tag: tagLabel || ref, ref };
+}
+
+function refExists(ref) {
+  try {
+    sh(["rev-parse", "--verify", "--quiet", `${ref}^{commit}`]);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+export function resolveCurrentRef(
+  currentTag,
+  { requestedRef = process.env.RELEASE_EVIDENCE_REF || "", refExistsImpl = refExists } = {},
+) {
+  const explicit = String(requestedRef || "").trim();
+  if (explicit) {
+    if (!refExistsImpl(explicit)) {
+      fail(`RELEASE_EVIDENCE_REF does not exist or is not a commit: ${explicit}`);
+    }
+    return explicit;
+  }
+  if (currentTag && refExistsImpl(currentTag)) return currentTag;
+  return "HEAD";
+}
+
+function listProductTags() {
+  try {
+    sh(["fetch", "--tags", "--force", "origin"], { stdio: ["ignore", "ignore", "ignore"] });
+  } catch {
+    warn("Failed to fetch tags from origin; using locally available tags.");
+  }
+
+  const text = sh(["tag", "--list"]);
+  return text
+    .split("\n")
+    .map((tag) => tag.trim())
+    .filter(Boolean)
+    .map((tag) => ({ tag, version: versionFromTag(tag) }))
+    .filter((item) => item.version && parseSemver(item.version));
+}
+
+export function findPreviousTag(targetVersion, currentTag) {
+  const candidates = listProductTags()
+    .filter((item) => item.tag !== currentTag)
+    .filter((item) => compareSemver(item.version, targetVersion) < 0)
+    .sort((a, b) => {
+      const versionOrder = compareSemver(b.version, a.version);
+      if (versionOrder !== 0) return versionOrder;
+      const aPreferred = a.tag.startsWith(CURRENT_TAG_PREFIX) ? 1 : 0;
+      const bPreferred = b.tag.startsWith(CURRENT_TAG_PREFIX) ? 1 : 0;
+      return bPreferred - aPreferred;
+    });
+  return candidates[0]?.tag || "";
+}
+
+function parseCommits(previousTag, currentRef) {
+  const text = sh([
+    "log",
+    "--format=%H%x09%h%x09%s",
+    "--no-merges",
+    `${previousTag}..${currentRef}`,
+    "--",
+    PRODUCT_PATH,
+  ]);
+  return text
+    .split("\n")
+    .filter(Boolean)
+    .map((line) => {
+      const [sha = "", shortSha = "", subject = ""] = line.split("\t");
+      return { sha, short_sha: shortSha, subject };
+    });
+}
+
+function parseChangedFiles(previousTag, currentRef) {
+  const text = sh(["diff", "--name-status", `${previousTag}..${currentRef}`, "--", PRODUCT_PATH]);
+  return text
+    .split("\n")
+    .filter(Boolean)
+    .map((line) => {
+      const parts = line.split("\t");
+      const item = { status: parts[0], path: parts[parts.length - 1] };
+      if (parts.length === 3) item.old_path = parts[1];
+      return item;
+    });
+}
+
+function packageChanges(previousTag, currentRef) {
+  const path = `${PRODUCT_PATH}/package.json`;
+  const before = gitShowJson(previousTag, path);
+  const after = currentRef === "HEAD" ? readJsonFile(path) : gitShowJson(currentRef, path);
+  return ["name", "version"]
+    .filter((field) => before[field] !== after[field])
+    .map((field) => ({ field, before: before[field], after: after[field] }));
+}
+
+function extractPullRequests(commits, repo = process.env.GITHUB_REPOSITORY || "MemTensor/MemOS") {
+  const seen = new Set();
+  for (const commit of commits) {
+    for (const match of String(commit.subject || "").matchAll(/#(\d+)/g)) {
+      seen.add(match[1]);
+    }
+  }
+  return [...seen].map((number) => ({
+    number,
+    url: `https://github.com/${repo}/pull/${number}`,
+  }));
+}
+
+function refsForGuidance(commit) {
+  const refs = [];
+  if (commit.short_sha) refs.push(commit.short_sha);
+  for (const match of String(commit.subject || "").matchAll(/#(\d+)/g)) {
+    const value = `#${match[1]}`;
+    if (!refs.includes(value)) refs.push(value);
+  }
+  return refs;
+}
+
+function categoryHintForSubject(subject) {
+  const value = String(subject || "");
+  const lower = value.toLowerCase();
+  if (
+    lower.startsWith("release:") ||
+    /^fix(\([^)]+\))?:\s*ruff\b/i.test(value) ||
+    /review feedback/i.test(value)
+  ) {
+    return null;
+  }
+  if (/^(feat|feature|add)(\([^)]+\))?:|^add\s+/i.test(value)) {
+    return {
+      category: "Added",
+      reason: "new user-facing capability or configuration support",
+    };
+  }
+  if (
+    /cpu|full-table vector scan|vector scan|xml boundary|memory context with xml|hermes|provider link|bridge package version|startup poll|double-spawn/i.test(
+      value,
+    )
+  ) {
+    return {
+      category: "Improved",
+      reason: "performance, context-boundary, or integration robustness improvement",
+    };
+  }
+  if (/^(perf|performance|refactor)(\([^)]+\))?:/i.test(value)) {
+    return {
+      category: "Improved",
+      reason: "performance or implementation improvement",
+    };
+  }
+  if (
+    /lightweightmemory|skip evolution|endpoint paths|openai_compatible|empty response|error response|dirty-reward|orphan trace|session exemption|loopback/i.test(
+      lower,
+    )
+  ) {
+    return {
+      category: "Fixed",
+      reason: "specific broken behavior, recovery boundary, retry, auth, or endpoint probing fix",
+    };
+  }
+  if (/^(fix|hotfix|bugfix)(\([^)]+\))?:|^fix\s+#\d+/i.test(value)) {
+    return {
+      category: "Fixed",
+      reason: "specific bug fix",
+    };
+  }
+  return null;
+}
+
+export function categoryHintsForCommits(commits) {
+  return commits
+    .map((commit) => {
+      const hint = categoryHintForSubject(commit.subject);
+      const sourceRefs = refsForGuidance(commit);
+      if (!hint || sourceRefs.length === 0) return null;
+      return {
+        ...hint,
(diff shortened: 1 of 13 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2138
**Task:** `3d521927b26d5c6f`
**Base:** `main`
**Head:** `dev-v2.0.25-sdk`

🔍 **OpenCodeReview** found **26** issue(s) in this PR.

---

### 1. `.github/workflows/memos-local-plugin-post-merge-dry-run.yml` (L23)

The `version` input is hardcoded to `"2.0.10"`. As the project advances, this value will become stale and the dry-run will silently test an already-released, outdated version rather than the next upcoming one. Consider deriving the version dynamically — e.g., reading it from `package.json` or `pyproject.toml` at runtime — so the dry-run always reflects the current development snapshot.

Example:
```yaml
version: ${{ needs.read-version.outputs.version }}
```
or, if a simple shell step is acceptable inside the called workflow's `with:` block, pass a placeholder like `"0.0.0-dry-run"` that flags the intent clearly.

**Before:**
```
      version: "2.0.10"
```

**After:**
```suggestion
      # Derive dynamically, e.g. from a prior job that reads package.json
      version: ${{ needs.read-version.outputs.version }}
```

</details>

---

### 2. `.github/workflows/memos-local-plugin-post-merge-dry-run.yml` (L14-L16)

`contents: write` and `pull-requests: write` are unnecessary for a dry-run workflow. With `dry_run: true` hardcoded, the called workflow explicitly skips every step that requires write access (tag push, GitHub Release creation, PR creation). Granting write permissions violates the principle of least privilege; if this workflow were ever accidentally triggered or the callee logic changed, it would have more authority than needed.

For a read-only dry-run, `contents: read` (needed for checkout) is sufficient.

**Before:**
```
permissions:
  contents: write
  pull-requests: write
```

**After:**
```suggestion
permissions:
  contents: read
```

</details>

---

### 3. `.github/workflows/memos-local-plugin-post-merge-dry-run.yml` (L19-L21)

The `dry-run` job has no `timeout-minutes`. The called workflow (`memos-local-plugin-publish.yml`) spawns multi-platform matrix builds, npm operations, and external API calls — all of which can hang indefinitely on transient failures. Without a timeout on the calling job, a stalled runner will consume quota until the default 6-hour GitHub Actions limit is hit.

Consider setting a realistic upper bound, e.g. `timeout-minutes: 60`.

**Before:**
```
  dry-run:
    uses: ./.github/workflows/memos-local-plugin-publish.yml
```

**After:**
```suggestion
  dry-run:
    if: ${{ github.repository == 'MemTensor/MemOS' }}
    timeout-minutes: 60
    uses: ./.github/workflows/memos-local-plugin-publish.yml
```

</details>

---

### 4. `.github/workflows/memos-local-plugin-post-merge-dry-run.yml` (L3-L7)

No `concurrency` group is defined on this workflow. Rapid successive pushes that touch the listed paths (e.g., a series of script fixes) would queue multiple dry-run runs simultaneously. While the *called* workflow already defines `concurrency: group: memos-local-plugin-publish` (which would serialize executions), relying on that inner lock means extra runs pile up rather than being cancelled outright.

Adding a `concurrency` block here with `cancel-in-progress: true` would discard superseded dry-runs immediately.

**Before:**
```
on:
  push:
    branches:
      - main
    paths:
```

**After:**
```suggestion
concurrency:
  group: memos-local-plugin-post-merge-dry-run
  cancel-in-progress: true

on:
  push:
    branches:
      - main
    paths:
```

</details>

---

### 5. `.github/scripts/retry.sh` (L10-L21)

**Bug: Unbound variable crash on incomplete flag pairs under `set -u`**

With `set -u` active from line 2, if a recognized flag (`--attempts`, `--delay`, or `--label`) is passed as the last argument without its value (e.g., `retry.sh --attempts`), the shell will attempt to expand `$2` which is unset and immediately abort with an "unbound variable" error — bypassing the intended `exit 2` error message below. The guard at line 31 is never reached in this case.

Suggestion: Validate that `$2` is set before using it in each case branch, or use `${2:-}` with a follow-up check:
```bash
--attempts)
  if [ $# -lt 2 ]; then
    echo "::error::--attempts requires a value." >&2
    exit 2
  fi
  attempts="$2"
  shift 2
  ;;
```

---

### 6. `.github/scripts/retry.sh` (L26-L30)

**Bug: Typo in flag name silently becomes the command to execute**

The `*)` wildcard case silently `break`s out of the argument-parsing loop and treats the unrecognized token as the start of the command. For example, `retry.sh --attemps 5 -- my-command` will attempt to execute `--attemps` as a binary. Since the unrecognized flag starts with `--`, most shells will produce a confusing "command not found" error rather than a usage hint.

Suggestion: Distinguish between `--`-prefixed tokens (likely a flag typo) and bare tokens (start of the command):
```bash
    --*)
      echo "::error::Unknown option: $1" >&2
      exit 2
      ;;
    *)
      break
      ;;
```

---

### 7. `.github/scripts/retry.sh` (L42-L45)

**Issue: Zero delay is accepted without warning, risking rapid-fire retries against rate-limited services**

The validation regex `^[0-9]+$` permits `delay=0`, which results in `sleep 0` between every retry. The workflows wrap network-dependent operations (npm publish, git push, npm ci) with this script. If a caller accidentally passes `--delay 0` or sets `RETRY_DELAY_SECONDS=0` in the environment, all retry attempts will fire back-to-back with no pause, which can exacerbate rate-limiting or transient failures.

Suggestion: Require at least 1 second for delay, or add an explicit note in the validation message that zero is a valid intentional choice:
```bash
if ! [[ "${delay}" =~ ^[1-9][0-9]*$ ]]; then
  echo "::error::Invalid retry delay: ${delay} (must be a positive integer)" >&2
  exit 2
fi
```
If zero-delay is intentionally supported (e.g., for unit testing), document it explicitly.

---

### 8. `.github/scripts/draft-local-plugin-release-notes.test.mjs` (L531-L534)

Assigning a plain object to `process.env` does NOT restore the real Node.js process environment. `process.env` is a special proxy object backed by the OS environment; overwriting the reference in the local closure has no effect on the underlying environment. Environment variables set via `Object.assign(process.env, { ... })` (lines 517–524) will therefore leak into subsequent tests.

The correct cleanup is to explicitly delete keys that were added and restore keys that were changed:

**Before:**
```
  } finally {
    process.env = previous;
    rmSync(directory, { recursive: true, force: true });
  }
```

**After:**
```suggestion
  } finally {
    // Restore only the keys that were mutated/added
    for (const key of ["RELEASE_FAILURE_PHASE", "RELEASE_FAILURE_ATTEMPT_DIR", "RELEASE_VERSION", "RELEASE_TAG", "DOC_AGENT_RELEASE_FAILURE_URL", "DOC_AGENT_RELEASE_NOTES_DRAFT_TOKEN"]) {
      if (key in previous) {
        process.env[key] = previous[key];
      } else {
        delete process.env[key];
      }
    }
    rmSync(directory, { recursive: true, force: true });
  }
```

</details>

---

### 9. `.github/scripts/draft-local-plugin-release-notes.test.mjs` (L583-L588)

Same `process.env = previous` anti-pattern as in the test above: the assignment replaces only the local reference, not the real process environment. Mutated env vars (`DOC_AGENT_RELEASE_NOTES_DRAFT_URL`, `DOC_AGENT_RELEASE_NOTES_DRAFT_TOKEN`) leak into subsequent tests.

Use per-key delete/restore instead:
```js
} finally {
  for (const key of ["DOC_AGENT_RELEASE_NOTES_DRAFT_URL", "DOC_AGENT_RELEASE_NOTES_DRAFT_TOKEN"]) {
    if (key in previous) process.env[key] = previous[key];
    else delete process.env[key];
  }
}
```
The same fix applies to all five tests in this file that use this pattern.

---

### 10. `.github/scripts/draft-local-plugin-release-notes.test.mjs` (L249-L252)

The optional chaining `?.` on the result of `find()` means that if no item contains `"记忆恢复"`, the expression evaluates to `undefined`. With `node:assert/strict`, `assert.equal(undefined, false)` will throw (`undefined !== false`), making the test fail with a confusing equality error rather than a clear message.

More importantly, this form is semantically ambiguous: a future refactor that renames the postprocessed Chinese text would cause the `find()` to return `undefined`, the assertion would *fail* loudly (strict mode) instead of passing — which is the opposite of masking the bug. While the current behavior is technically correct (the item does exist and does not contain "XML"), the fragile guard still makes the intent unclear and error messages unhelpful.

Prefer an explicit lookup with a dedicated assertion:

**Before:**
```
  assert.equal(
    processed.release_items.find((item) => item.text_cn.includes("记忆恢复"))?.text_cn.includes("XML"),
    false,
  );
```

**After:**
```suggestion
  const memoryRecoveryItem = processed.release_items.find((item) => item.text_cn.includes("记忆恢复"));
  assert.ok(memoryRecoveryItem, "expected a release item whose text_cn contains 记忆恢复");
  assert.ok(!memoryRecoveryItem.text_cn.includes("XML"), "merged Fixed item should not contain XML after context-boundary deduplication");
```

</details>

---

### 11. `.github/scripts/draft-local-plugin-release-notes.mjs` (L1187-L1199)

Path traversal risk: `attemptDir` is taken directly from `process.env.RELEASE_FAILURE_ATTEMPT_DIR` with no validation. A value like `../../etc` would cause the script to read files outside the intended log directory, and their contents would then be sent to the external failure-report URL after `cleanError` redaction. Even in CI, defense-in-depth validation should be applied.

Suggested fix: resolve `attemptDir` to an absolute path and assert it stays within `RUNNER_TEMP` (or another known-safe base) before constructing child paths.

**Before:**
```
  const attempts = [1, 2, 3].map((attempt) => {
    let message = "attempt log is unavailable";
    try { message = readFileSync(join(attemptDir, `${attempt}.log`), "utf8"); } catch {}
    return { error_code: phase.toUpperCase().replace(/[^A-Z0-9]+/g, "_"), message: cleanError(message), retryable: true };
  });
  return reportFailure({
    evidence: {
      repo: process.env.GITHUB_REPOSITORY || "MemTensor/MemOS",
      target_version: displayVersion(process.env.RELEASE_VERSION),
      current_tag: process.env.RELEASE_TAG || `${CURRENT_TAG_PREFIX}${cleanVersion(process.env.RELEASE_VERSION)}`,
    },
    attempts,
    finalError: attempts[2].message,
```

**After:**
```suggestion
  const safeBase = process.env.RUNNER_TEMP || tmpdir();
  const resolvedAttemptDir = join(safeBase, basename(attemptDir));
  if (!resolvedAttemptDir.startsWith(safeBase + sep)) {
    fail(`RELEASE_FAILURE_ATTEMPT_DIR resolves outside the expected base: ${resolvedAttemptDir}`);
  }
  const attempts = [1, 2, 3].map((attempt) => {
    let message = "attempt log is unavailable";
    try { message = readFileSync(join(resolvedAttemptDir, `${attempt}.log`), "utf8"); } catch {}
    return { error_code: phase.toUpperCase().replace(/[^A-Z0-9]+/g, "_"), message: cleanError(message), retryable: true };
  });
```

</details>

---

### 12. `.github/scripts/draft-local-plugin-release-notes.mjs` (L1225)

Credential leakage via `previous_errors`: each `item.message` stored in `attempts` is already the result of `cleanError()`, which strips `Bearer <token>` patterns. However, `cleanError` only catches `Bearer\s+\S+` — it would miss tokens that appear without the `Bearer` prefix (e.g., in a network error that formats the header differently). Additionally the same token is also used as-is to authenticate the failure-report endpoint (`reportFailure`), so a single compromised endpoint URL receives the secret. The risk is low-to-medium given that GitHub Actions automatically masks registered secrets in log output, but the defense-in-depth guarantee depends entirely on that runner-level masking being active.

---

### 13. `.github/scripts/draft-local-plugin-release-notes.mjs` (L334-L340)

Raw diff content sent to external service without redaction: the first 24 000 characters of the git diff for `apps/memos-local-plugin` are captured verbatim and later transmitted to `DOC_AGENT_RELEASE_NOTES_DRAFT_URL` as part of the evidence payload. If any committed file within that path contains credentials, tokens, private keys, or otherwise sensitive source code, that content will be sent to the third-party AI drafting service with no filtering. The `cleanError` function only redacts error message strings, not the evidence payload. This is a structural design concern: the script should either (a) confirm the draft service is a trusted internal endpoint and document that requirement clearly, or (b) apply allowlist-based diff filtering before transmission.

---

### 14. `.github/scripts/draft-local-plugin-release-notes.mjs` (L1245-L1263)

Server-supplied `item.error` is passed directly into `reportFailure` without first running it through `cleanError`. If the server response contains a token, URL, or other sensitive string inside `payload.attempts[].error`, it will be forwarded to the failure-report endpoint and potentially logged by `warn()` in a subsequent retry path.

**Before:**
```
  if (!payload.ok || payload.needs_review) {
        const serverAttempts = Array.isArray(payload.attempts) ? payload.attempts : [];
        const coverage = payload.coverage ? JSON.stringify(payload.coverage) : "";
        const warnings = Array.isArray(payload.warnings) ? payload.warnings.join("; ") : "";
        const message = `Release notes draft needs review. ${coverage} ${warnings}`.trim();
        if (serverAttempts.length >= 3) {
          await reportFailure({
            evidence,
            attempts: serverAttempts.map((item) => ({
              error_code: "DRAFT_VALIDATION",
              message: item.error || message,
              retryable: false,
            })),
            finalError: message,
            fetchImpl,
          });
        }
        fail(message);
      }
```

**After:**
```suggestion
  if (!payload.ok || payload.needs_review) {
        const serverAttempts = Array.isArray(payload.attempts) ? payload.attempts : [];
        const coverage = payload.coverage ? JSON.stringify(payload.coverage) : "";
        const warnings = Array.isArray(payload.warnings) ? payload.warnings.join("; ") : "";
        const message = `Release notes draft needs review. ${coverage} ${warnings}`.trim();
        if (serverAttempts.length >= 3) {
          await reportFailure({
            evidence,
            attempts: serverAttempts.map((item) => ({
              error_code: "DRAFT_VALIDATION",
              message: cleanError(item.error || message),
              retryable: false,
            })),
            finalError: message,
            fetchImpl,
          });
        }
        fail(message);
      }
```

</details>

---

### 15. `.github/scripts/draft-local-plugin-release-notes.mjs` (L1282)

No per-request timeout on `fetchImpl` calls: Node.js `fetch` has no built-in timeout. If the draft service hangs (e.g., during a network partition), the script will block indefinitely at each of the three attempts in `requestDraft`, and additionally at each repair attempt in `requestValidatedDraft`. With `maxRepairAttempts=2`, this can block up to 9 `fetch` calls without bound. Although the workflow YAML has a job-level timeout, an `AbortSignal.timeout()` per request is strongly recommended to fail fast and surface the problem clearly rather than exhausting the entire CI runner budget.

**Before:**
```
    await sleep(250 * 2 ** (attempt - 1));
```

**After:**
```suggestion
      const controller = new AbortController();
      const timeoutId = setTimeout(() => controller.abort(), 60_000); // 60 s per request
      try {
        const response = await fetchImpl(url, {
          method: "POST",
          headers: { "content-type": "application/json", authorization: `Bearer ${token}` },
          body: JSON.stringify({ ... }),
          signal: controller.signal,
        });
      } finally {
        clearTimeout(timeoutId);
      }
      // ... or equivalently: signal: AbortSignal.timeout(60_000)
```

</details>

---

### 16. `.github/scripts/draft-local-plugin-release-notes.mjs` (L418-L424)

GITHUB_OUTPUT multiline delimiter collision: if any output `value` contains the string `__DOC_AGENT_EOF__` on a line by itself (possible for `release_notes_markdown` content, which is AI-generated and unfiltered), the GitHub Actions output parser will treat it as the end-of-value marker prematurely, corrupting that output and all subsequent outputs in the same step. Use a random or content-hashed delimiter to eliminate this risk.

**Before:**
```
  const outputPath = process.env.GITHUB_OUTPUT;
  if (!outputPath) return;
  writeFileSync(outputPath, `${name}<<__DOC_AGENT_EOF__\n${value}\n__DOC_AGENT_EOF__\n`, {
    flag: "a",
  });
}
```

**After:**
```suggestion
function appendOutput(name, value) {
  const outputPath = process.env.GITHUB_OUTPUT;
  if (!outputPath) return;
  // Use a random delimiter to prevent collision with output values (GitHub Actions spec allows any unique string)
  const delimiter = `__DOC_AGENT_EOF_${Math.random().toString(36).slice(2).toUpperCase()}__`;
  writeFileSync(outputPath, `${name}<<${delimiter}\n${value}\n${delimiter}\n`, {
    flag: "a",
  });
}
```

</details>

---

### 17. `.github/workflows/memos-local-plugin-publish.yml` (L473)

Wrong product name: the GitHub Release title uses "OpenClaw Local Plugin" but every other reference in this workflow (tags, branches, PR titles, job names) uses "memos-local-plugin" / "MemOS Local Plugin". This looks like a copy-paste leftover from a different project. The published release will carry an incorrect title visible to all users.

Suggested fix: align with the established naming convention.

**Before:**
```
              --title "OpenClaw Local Plugin v${RELEASE_VERSION}" \
```

**After:**
```suggestion
              --title "MemOS Local Plugin v${RELEASE_VERSION}" \
```

</details>

---

### 18. `.github/workflows/memos-local-plugin-publish.yml` (L293-L296)

The "Stop before publish in dry run" step only prints a notice but does NOT halt the job. The subsequent publish-gating steps (`Publish to npm`, `Create release tag and PR`) rely on their own independent `if: ${{ inputs.dry_run != true }}` guards — if either guard were accidentally removed in a future edit, `dry_run=true` runs would silently publish. For an explicit, fail-safe dry-run stop, add `exit 1` (or use `fail` with a non-error notice convention) so the job itself stops here and the remaining steps are skipped by the natural job-failure mechanism.

**Before:**
```
      - name: Stop before publish in dry run
        if: ${{ inputs.dry_run == true }}
        run: |
          echo "::notice::dry_run=true; skipped npm publish, tag, GitHub Release, and release PR."
```

**After:**
```suggestion
      - name: Stop before publish in dry run
        if: ${{ inputs.dry_run == true }}
        run: |
          echo "::notice::dry_run=true; skipped npm publish, tag, GitHub Release, and release PR."
          exit 1
        continue-on-error: false
```

</details>

---

### 19. `.github/workflows/memos-local-plugin-publish.yml` (L529-L540)

The TOCTOU-safe `--force-with-lease=<ref>:<expected-sha>` guard has been replaced by a check-then-push pattern that is not atomic. Between `git ls-remote` (inside `remote_branch_sha`) and the actual `git push`, another runner or concurrent job could update the remote branch, causing a silent overwrite or a confusing push rejection with no retry.

The old pattern was:
```bash
git push --force-with-lease="refs/heads/${release_branch}:${remote_branch_sha}" origin "HEAD:refs/heads/${release_branch}"
```
This atomically refuses to push if the remote moved. Consider restoring it (passing the fetched SHA to `--force-with-lease`) or at minimum documenting why the race is acceptable here.

---

### 20. `.github/workflows/memos-local-plugin-publish.yml` (L21-L30)

Under `workflow_dispatch`, `required: true` with a non-empty `default:` is contradictory: the UI will pre-fill the default, so the field is never truly "required" in the sense that the runner blocks on it. More importantly, the `if:` conditions throughout the job (`inputs.dry_run == true` / `inputs.dry_run != true`) work correctly when `type: boolean` is declared (GHA evaluates it natively), but if this workflow is ever called from a shell script via the REST API without the `type:` hint, the value arrives as a string `"true"`/`"false"` and the `==` comparison will still work — so the current setup is acceptable. However, keeping `required: true` misleadingly suggests the UI will reject a missing value even though a default is provided. Consider changing to `required: false` to match the `workflow_call` declaration.

**Before:**
```
      dry_run:
        description: "Draft release notes and build artifacts only. Skip npm publish, tag, GitHub Release, and release PR."
        required: true
        type: boolean
        default: true
      recover_existing_npm_release:
        description: "Allow reconstructing a missing tag/Release for an npm version. Keep false for normal releases."
        required: true
        type: boolean
        default: false
```

**After:**
```suggestion
      dry_run:
        description: "Draft release notes and build artifacts only. Skip npm publish, tag, GitHub Release, and release PR."
        required: false
        type: boolean
        default: true
      recover_existing_npm_release:
        description: "Allow reconstructing a missing tag/Release for an npm version. Keep false for normal releases."
        required: false
        type: boolean
        default: false
```

</details>

---

### 21. `.github/workflows/memos-local-plugin-publish.yml` (L76-L88)

Neither `build-prebuilds` nor `publish` declare a `timeout-minutes`. The `publish` job performs multiple network-retried operations (npm publish × 3, GitHub Release creation × 3, AI drafting service × 3, `git push` × 3). Without a timeout, a hung network call or a slow AI service can hold the concurrency group lock indefinitely and exhaust free runner minutes. Add a reasonable cap, e.g. 30 minutes for `build-prebuilds` and 45 minutes for `publish`.

**Before:**
```
  build-prebuilds:
    strategy:
      matrix:
        include:
          - os: macos-14
            platform: darwin-arm64
          - os: macos-15-intel
            platform: darwin-x64
          - os: ubuntu-latest
            platform: linux-x64
          - os: windows-latest
            platform: win32-x64
    runs-on: ${{ matrix.os }}
```

**After:**
```suggestion
  build-prebuilds:
    timeout-minutes: 30
    strategy:
      matrix:
        include:
          - os: macos-14
            platform: darwin-arm64
          - os: macos-15-intel
            platform: darwin-x64
          - os: ubuntu-latest
            platform: linux-x64
          - os: windows-latest
            platform: win32-x64
    runs-on: ${{ matrix.os }}
```

</details>

---

### 22. `.github/workflows/memos-local-plugin-publish.yml` (L122-L124)

The `publish` job also lacks `timeout-minutes`. Given the number of retried network operations it performs (AI drafting, npm publish, GitHub API calls, git push), it should have an explicit upper bound.

**Before:**
```
  publish:
    needs: build-prebuilds
    runs-on: ubuntu-latest
```

**After:**
```suggestion
  publish:
    needs: build-prebuilds
    runs-on: ubuntu-latest
    timeout-minutes: 45
```

</details>

---

### 23. `.github/workflows/memos-local-plugin-publish.yml` (L111-L114)

The inner `bash -euo pipefail -c '...'` is passed as a single quoted argument through the `retry.sh` `--` separator. `retry.sh` then runs `"$@"` which expands to `bash -euo pipefail -c '<multiline>'`. This works correctly with the current `retry.sh` implementation (it uses `"$@"` faithfully). However, the inner script body is a single-quoted here-string embedded inside a double-quoted YAML block, which means YAML itself sees it as a plain multi-line string — this is fine. No functional issue, but consider using a named script file or a step-level shell for better readability and easier debugging.

---

### 24. `.github/workflows/memos-local-plugin-publish.yml` (L131-L134)

Dependency installation (`npm ci`) runs without caching on every workflow run. `actions/setup-node` supports a built-in cache via `cache: 'npm'` which caches `~/.npm` and avoids downloading all packages on each run. For a matrix of 4 platforms × each run, this adds up quickly.

**Before:**
```
      - uses: actions/setup-node@v4
        with:
          node-version: 22
          registry-url: https://registry.npmjs.org
```

**After:**
```suggestion
      - uses: actions/setup-node@v4
        with:
          node-version: 22
          registry-url: https://registry.npmjs.org
          cache: 'npm'
          cache-dependency-path: apps/memos-local-plugin/package-lock.json
```

</details>

---

### 25. `.github/scripts/draft-local-plugin-release-notes.mjs` (L438)

Overbroad numeric-to-PR-ref conversion: `normalizeSourceRef` promotes any 2+-digit string to a PR reference by prepending `#`. This will misidentify version numbers (e.g., `10`, `22`), line counts, or other numeric values that appear in commit subjects as GitHub PR numbers, creating invalid `source_refs` in release note items. The same pattern in `refsForGuidance` / `extractPullRequests` (`/#(\d+)/g`) is at least guarded by a `#` prefix, but the normalizer accepts raw numeric strings which may reach it from `normalizeSourceRefs` when given an array from the AI draft.

**Before:**
```
if (/^\d{2,}$/.test(text)) return `#${text}`;
```

**After:**
```suggestion
function normalizeSourceRef(value) {
  const text = String(value || "").trim().replace(/^[`[(\s]+|[`)\],.;\s]+$/g, "");
  // Only accept PR refs that already carry the '#' prefix; do not promote bare numbers
  if (/^#\d+$/.test(text)) return text;
  if (/^[a-fA-F0-9]{7,40}$/.test(text)) return text.toLowerCase();
  return "";
}
```

</details>

---

### 26. `tests/api/test_client.py` (L494)

Neither the `prebuild` job nor the `publish` job declares a `permissions` block. GitHub Actions defaults to broad access (`contents: write`, `pull-requests: write`, etc.) when no `permissions` key is present. The `publish` job uses `GH_TOKEN: ${{ github.token }}` for `gh release create`, `gh pr create`, and `git push`, meaning it runs with write-all access. This violates the least-privilege principle.

Add minimal, explicit permissions to each job:
```yaml
# prebuild job
permissions:
  contents: read

# publish job
permissions:
  contents: write
  pull-requests: write
```

**Before:**
```
assert response.data.task_id == "task-legacy"
```

**After:**
```suggestion
  publish:
    runs-on: ubuntu-latest
    needs: prebuild
    permissions:
      contents: write
      pull-requests: write
```

</details>

---

🧹 Filtered **4** low-confidence OCR finding(s) before posting/fix-loop (existing_code_mismatch: 2, duplicate: 2).

Memtensor-AI:

**Details:** Executor error: Command failed: git clone --depth 1 --branch dev-v2.0.25-sdk git@github.com:Zyntrael/MemOS.git /data/test-workspaces/3d521927b26d5c6f/repo
Cloning into '/data/test-workspaces/3d521927b26d5c6f/repo'...
error: unable to write file apps/memos-local-plugin/tests/unit/capture/step-extractor.test.ts
error: unable to write file apps/memos-local-plugin/tests/unit/capture/subscriber.test.ts
error: unable to write file apps/memos-local-plugin/tests/unit/capture/tagger.test.ts
fatal: cannot create directory at 'apps/memos-local-plugin/tests/unit/config': No space left on device
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
**Branch:** `dev-v2.0.25-sdk`

PR #2142, opened 2026-07-22
Performance optimization for searching memories

endxxxx:
## Description

## Type of change

- [x] Refactor (does not change functionality, e.g. code style improvements, linting)

## How Has This Been Tested?

- [ ] Unit Test

## Checklist

## Reviewer Checklist
- [ ] Made sure Checks passed

### DIFF
--- a/src/memos/api/handlers/search_handler.py
+++ b/src/memos/api/handlers/search_handler.py
@@ -32,6 +32,8 @@ logger = get_logger(__name__)
 _ENV_CONTEXT_RECALL = "MEMOS_DREAM_CONTEXT_RECALL"
 _ENV_CONTEXT_RECALL_TOP_K = "MEMOS_DREAM_CONTEXT_RECALL_TOP_K"
 _DEFAULT_CONTEXT_RECALL_TOP_K = 2
+_ENV_MMR_CANDIDATE_PRUNING = "MEMOS_MMR_CANDIDATE_PRUNING_ENABLED"
+_MMR_CANDIDATE_MULTIPLIER = 2
 
 
 def _env_enabled(name: str, default: str = "off") -> bool:
@@ -352,6 +354,12 @@ class SearchHandler(BaseHandler):
         if not text_buckets and not pref_buckets:
             return results
 
+        self._prune_mmr_candidates_by_bucket(
+            results,
+            text_top_k=text_top_k,
+            pref_top_k=pref_top_k,
+        )
+
         # Flatten all memories with their type and scores
         # flat structure: (memory_type, bucket_idx, mem, score)
         flat: list[tuple[str, int, dict[str, Any], float]] = []
@@ -555,6 +563,87 @@ class SearchHandler(BaseHandler):
             bucket["memories"] = [flat[i][2] for i in selected_indices]
 
         return results
+
+    def _prune_mmr_candidates_by_bucket(
+        self,
+        results: dict[str, Any],
+        *,
+        text_top_k: int,
+        pref_top_k: int,
+    ) -> dict[str, Any]:
+        """Keep at most twice the final quota for each memory type in each bucket."""
+        if not _env_enabled(_ENV_MMR_CANDIDATE_PRUNING, "off"):
+            return results
+
+        total_before = 0
+        total_after = 0
+        bucket_counts: list[tuple[int, int]] = []
+
+        for result_key, target_top_k in (
+            ("text_mem", text_top_k),
+            ("pref_mem", pref_top_k),
+        ):
+            buckets = results.get(result_key)
+            if not isinstance(buckets, list):
+                continue
+
+            candidate_limit = max(0, int(target_top_k)) * _MMR_CANDIDATE_MULTIPLIER
+            for bucket in buckets:
+                memories = bucket.get("memories") if isinstance(bucket, dict) else None
+                if not isinstance(memories, list):
+                    continue
+
+                before_count = len(memories)
+                total_before += before_count
+                memories_by_type: dict[str, list[dict[str, Any]]] = {}
+                for memory in memories:
+                    if not isinstance(memory, dict):
+                        continue
+                    metadata = memory.get("metadata")
+                    memory_type = (
+                        metadata.get("memory_type") if isinstance(metadata, dict) else None
+                    )
+                    memories_by_type.setdefault(str(memory_type or result_key), []).append(memory)
+
+                selected: list[dict[str, Any]] = []
+                if candidate_limit > 0:
+                    for typed_memories in memories_by_type.values():
+                        selected.extend(
+                            sorted(
+                                typed_memories,
+                                key=self._mmr_candidate_score,
+                                reverse=True,
+                            )[:candidate_limit]
+                        )
+                    selected.sort(key=self._mmr_candidate_score, reverse=True)
+
+                bucket["memories"] = selected
+                if "total_nodes" in bucket:
+                    bucket["total_nodes"] = len(selected)
+                total_after += len(selected)
+                bucket_counts.append((before_count, len(selected)))
+
+        self.logger.info(
+            "[SearchHandler] MMR candidate pruning: multiplier=%s before=%s after=%s "
+            "dropped=%s bucket_counts=%s",
+            _MMR_CANDIDATE_MULTIPLIER,
+            total_before,
+            total_after,
+            total_before - total_after,
+            bucket_counts,
+        )
+        return results
+
+    @staticmethod
+    def _mmr_candidate_score(memory: dict[str, Any]) -> float:
+        metadata = memory.get("metadata")
+        if not isinstance(metadata, dict):
+            return 0.0
+        score = metadata.get("score", metadata.get("relativity", 0.0))
+        try:
+            return float(score or 0.0)
+        except (TypeError, ValueError):
+            return 0.0
 
     @staticmethod
     def _is_unrelated(
--- a/src/memos/api/handlers/search_handler.py
+++ b/src/memos/api/handlers/search_handler.py
@@ -16,7 +16,7 @@ from memos.api.handlers.base_handler import BaseHandler, HandlerDependencies
 from memos.api.handlers.formatters_handler import rerank_knowledge_mem
 from memos.api.product_models import APISearchRequest, SearchResponse
 from memos.dream.contextualization import CONTEXT_MEMORY_TYPE
-from memos.log import get_logger
+from memos.log import get_logger, summarize_search_request, summarize_search_results
 from memos.memories.textual.tree_text_memory.retrieve.retrieve_utils import (
     cosine_similarity_matrix,
 )
@@ -77,7 +77,10 @@ class SearchHandler(BaseHandler):
         Returns:
             SearchResponse with formatted results
         """
-        self.logger.info(f"[SearchHandler] Search Req is: {search_req}")
+        self.logger.info(
+            "[SearchHandler] Search request summary: %s",
+            summarize_search_request(search_req),
+        )
 
         # Use deepcopy to avoid modifying the original request object
         search_req_local = copy.deepcopy(search_req)
@@ -137,7 +140,8 @@ class SearchHandler(BaseHandler):
             results = hooked_results
 
         self.logger.info(
-            f"[SearchHandler] Final search results: count={len(results)} results={results}"
+            "[SearchHandler] Final search result summary: %s",
+            summarize_search_results(results),
         )
 
         return SearchResponse(
--- a/src/memos/embedders/ark.py
+++ b/src/memos/embedders/ark.py
@@ -1,6 +1,6 @@
 from memos.configs.embedder import ArkEmbedderConfig
 from memos.dependency import require_python_package
-from memos.embedders.base import BaseEmbedder
+from memos.embedders.base import BaseEmbedder, log_embedding_call
 from memos.log import get_logger
 
 
@@ -35,6 +35,7 @@ class ArkEmbedder(BaseEmbedder):
         # Initialize ark client
         self.client = Ark(api_key=self.config.api_key, base_url=self.config.api_base)
 
+    @log_embedding_call
     def embed(self, texts: list[str]) -> list[list[float]]:
         """
         Generate embeddings for the given texts.
--- a/src/memos/embedders/base.py
+++ b/src/memos/embedders/base.py
@@ -1,8 +1,63 @@
+import functools
 import re
+import time
 
 from abc import ABC, abstractmethod
+from collections.abc import Callable
+from typing import Any, TypeVar, cast
 
 from memos.configs.embedder import BaseEmbedderConfig
+from memos.log import get_logger, text_hash
+
+
+logger = get_logger(__name__)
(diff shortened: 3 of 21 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2142
**Task:** `manual-ocr:MemTensor/MemOS#2142`
**Base:** `dev-v2.0.25`
**Head:** `feat/performance-optimization-no-routing-v2.0.25`

🔍 **OpenCodeReview** found **19** issue(s) in this PR.

---

### 1. `src/memos/embedders/base.py` (L55-L58)

Embedding failures are logged at `logger.info` instead of `logger.error` (or at minimum `logger.warning`). Since ark, ollama, and sentence_transformer embedders have no independent error-level logging for `embed()` failures, operators filtering on ERROR/WARNING will never see these failures from this decorator. Change the error branch to `logger.error`.

**Before:**
```
            if error_type is None:
                logger.info(log_message, *log_values)
            else:
                logger.info(log_message + " error_type=%s", *log_values, error_type)
```

**After:**
```suggestion
            if error_type is None:
                logger.info(log_message, *log_values)
            else:
                logger.error(log_message + " error_type=%s", *log_values, error_type)
```

</details>

---

### 2. `src/memos/embedders/base.py` (L22-L32)

`list(texts or [])` materializes the input into `normalized_texts` for logging purposes, but `func` is then called with the original `texts`. If a caller ever passes a generator or iterator (instead of the declared `list[str]`), the iterator will be exhausted by the `list(...)` call and `func` will receive an empty iterator, silently producing no embeddings. Pass the materialized `normalized_texts` to `func` as well to be safe.

**Before:**
```
        normalized_texts = [texts] if isinstance(texts, str) else list(texts or [])
        text_lengths = [len(str(text or "")) for text in normalized_texts]
        config = getattr(self, "config", None)
        model = getattr(config, "model_name_or_path", None) or "unknown"
        backup_model = getattr(config, "backup_model_name_or_path", None) or "none"
        backup_enabled = bool(getattr(self, "use_backup_client", False))
        started_at = time.perf_counter()
        status = "success"
        error_type = None
        try:
            return func(self, texts, *args, **kwargs)
```

**After:**
```suggestion
        normalized_texts = [texts] if isinstance(texts, str) else list(texts or [])
        text_lengths = [len(str(text or "")) for text in normalized_texts]
        config = getattr(self, "config", None)
        model = getattr(config, "model_name_or_path", None) or "unknown"
        backup_model = getattr(config, "backup_model_name_or_path", None) or "none"
        backup_enabled = bool(getattr(self, "use_backup_client", False))
        started_at = time.perf_counter()
        status = "success"
        error_type = None
        try:
            return func(self, normalized_texts, *args, **kwargs)
```

</details>

---

### 3. `src/memos/embedders/base.py` (L51-L53)

`text_hash` is called unconditionally on every embedding invocation — even when the INFO log level is disabled — computing a SHA-256 over the entire text batch (O(total characters)). This is wasted CPU on every hot-path embedding call when INFO is suppressed. Guard it with a level check, or defer the computation using a lazy wrapper.

**Before:**
```
                text_hash(normalized_texts),
                elapsed_ms,
                status,
```

**After:**
```suggestion
                text_hash(normalized_texts) if logger.isEnabledFor(logging.INFO) else "skipped",
                elapsed_ms,
                status,
```

</details>

---

### 4. `src/memos/api/handlers/search_handler.py` (L608-L620)

**Silent data loss when `top_k=0`**: When `target_top_k` is `0`, `candidate_limit` becomes `0`, so the `if candidate_limit > 0` branch is skipped, `selected` stays `[]`, and `bucket['memories']` is overwritten with an empty list — silently dropping all candidates in that bucket before the MMR step.

Although the current call site passes `search_req.top_k` (which is typically positive), a defensive guard is needed to avoid catastrophic data loss when the value is zero. Consider treating `candidate_limit == 0` as "no limit" (i.e., skip pruning for that bucket) rather than "keep nothing".

**Before:**
```
                selected: list[dict[str, Any]] = []
                if candidate_limit > 0:
                    for typed_memories in memories_by_type.values():
                        selected.extend(
                            sorted(
                                typed_memories,
                                key=self._mmr_candidate_score,
                                reverse=True,
                            )[:candidate_limit]
                        )
                    selected.sort(key=self._mmr_candidate_score, reverse=True)

                bucket["memories"] = selected
```

**After:**
```suggestion
                if candidate_limit == 0:
                    # top_k=0 means "no limit" — keep all candidates untouched
                    continue

                selected: list[dict[str, Any]] = []
                for typed_memories in memories_by_type.values():
                    selected.extend(
                        sorted(
                            typed_memories,
                            key=self._mmr_candidate_score,
                            reverse=True,
                        )[:candidate_limit]
                    )
                selected.sort(key=self._mmr_candidate_score, reverse=True)

                bucket["memories"] = selected
```

</details>

---

### 5. `src/memos/api/handlers/search_handler.py` (L606)

**`memory_type or result_key` conflates empty-string type with untyped memories**: The expression `memory_type or result_key` falls back to `result_key` not only when `memory_type` is `None`, but also when it is an empty string `""` (falsy). If `memory_type=""` is a valid distinct type, it would be incorrectly grouped together with `None`-typed memories under the `result_key` bucket, and they would share the same `candidate_limit`, conflating two separate categories.

Prefer an explicit `None` check:

**Before:**
```
                    memories_by_type.setdefault(str(memory_type or result_key), []).append(memory)
```

**After:**
```suggestion
                    memories_by_type.setdefault(str(memory_type if memory_type is not None else result_key), []).append(memory)
```

</details>

---

### 6. `src/memos/log.py` (L76-L77)

The `or []` fallback is dead code. The enclosing condition `isinstance(bucket.get("memories"), list)` already guarantees the value is a list before this branch executes. An empty list `[]` is falsy, so `[] or []` still evaluates to `[]` — the fallback never produces a different result. Replace with `len(bucket.get("memories"))` to avoid misleading readers into thinking the value could be `None` at this point.

**Before:**
```
            len(bucket.get("memories") or [])
            if isinstance(bucket, Mapping) and isinstance(bucket.get("memories"), list)
```

**After:**
```suggestion
            len(bucket.get("memories"))
            if isinstance(bucket, Mapping) and isinstance(bucket.get("memories"), list)
```

</details>

---

### 7. `src/memos/embedders/cache.py` (L88)

**Bug (medium): `_request_cache_text_limit` uses the wrong config variable.**

`cache_max_size` is the maximum number of texts in the global TTL cache, but it's reused here to cap each per-request LRUCache. These are different dimensions: the outer TTLCache is sized by `request_cache_max_requests`, while each inner LRUCache should have its own independently configurable size. With the current code, setting `MEMOS_EMBEDDING_CACHE_MAX_SIZE=64` (intended to bound global TTL memory) also silently limits every per-request LRU to 64 entries — causing unexpected within-request cache misses if a single request embeds more than 64 unique texts. The variable `request_cache_max_requests` is read from env just above but is only used for the outer TTLCache.

Suggested fix: introduce a dedicated env variable (e.g. `MEMOS_EMBEDDING_REQUEST_CACHE_TEXT_LIMIT`) with its own default, or at minimum use the existing `_DEFAULT_CACHE_MAX_SIZE` explicitly as a documented fallback rather than re-referencing `cache_max_size`.

---

### 8. `src/memos/embedders/cache.py` (L30)

**Maintainability (low): `"trace-id"` is a hardcoded copy of a default defined in another module.**

In `src/memos/context/context.py` line 41, `RequestContext.__init__` defaults `trace_id` to `"trace-id"` when none is supplied: `self.trace_id = trace_id or "trace-id"`. This cache module replicates that sentinel as a magic string. If the default placeholder in `context.py` ever changes, this set will silently fail to exclude uninstrumented requests, causing all of them to share the same per-request LRU cache entry — every no-trace-id request would cross-contaminate each other's cache.

Suggested fix: export the sentinel constant from `context.py` (e.g., `DEFAULT_TRACE_ID = "trace-id"`) and import it here, or expose a predicate like `is_placeholder_trace_id(tid)` to keep the two modules in sync.

---

### 9. `src/memos/graph_dbs/polardb.py` (L42-L43)

`TypeError` is dead code here: `json.loads` raises `TypeError` only when its argument is not `str`, `bytes`, or `bytearray`, but the surrounding `isinstance(value, str)` guard already ensures the argument is always a `str`. More importantly, both exception types are swallowed silently with no log entry. If a field is stored as malformed agtype JSON, the caller silently receives a raw string where it may expect a list or dict, producing subtle downstream type errors that are very hard to diagnose. Add at minimum a `logger.debug` (or `logger.warning`) when falling back, and drop the dead `TypeError` clause.

**Before:**
```
        except (json.JSONDecodeError, TypeError):
            return value
```

**After:**
```suggestion
        except json.JSONDecodeError:
            logger.debug("Failed to JSON-decode agtype value %r; returning raw string", value)
            return value
```

</details>

---

### 10. `src/memos/graph_dbs/polardb.py` (L1957-L1959)

Minor log-formatting inconsistency: the conditional branch that logs the full query still has a leading space (`" search_by_embedding query: %s"`), while this new branch omits it. Either add the leading space here or remove it from the other branch for consistency.

---

### 11. `tests/embedders/test_base.py` (L27-L29)

`mock_logger.info.call_args` is `None` when the mock was never invoked (documented `unittest.mock` behaviour). If a regression in `log_embedding_call` causes it to skip the `logger.info` call, `.call_args.args` raises `AttributeError: 'NoneType' object has no attribute 'args'` instead of a clean assertion failure, making the root cause hard to diagnose. Add `mock_logger.info.assert_called_once()` before dereferencing `call_args` in all three tests.

**Before:**
```
    log_args = mock_logger.info.call_args.args
    rendered = log_args[0] % log_args[1:]
    assert "model=embedding-model" in rendered
```

**After:**
```suggestion
    mock_logger.info.assert_called_once()
    log_args = mock_logger.info.call_args.args
    rendered = log_args[0] % log_args[1:]
    assert "model=embedding-model" in rendered
```

</details>

---

### 12. `tests/embedders/test_base.py` (L55-L57)

Same `call_args` dereference without a prior `assert_called_once()` guard. If the error-path branch of `log_embedding_call` skips `logger.info`, the failure surfaces as `AttributeError` rather than a clear test failure.

**Before:**
```
    log_args = mock_logger.info.call_args.args
    rendered = log_args[0] % log_args[1:]
    assert "status=failed" in rendered
```

**After:**
```suggestion
    mock_logger.info.assert_called_once()
    log_args = mock_logger.info.call_args.args
    rendered = log_args[0] % log_args[1:]
    assert "status=failed" in rendered
```

</details>

---

### 13. `tests/embedders/test_base.py` (L77-L79)

Same `call_args` dereference without a prior `assert_called_once()` guard. A decorator regression that suppresses the `logger.info` call would yield a confusing `AttributeError` here instead of a meaningful assertion failure.

**Before:**
```
    log_args = mock_logger.info.call_args.args
    rendered = log_args[0] % log_args[1:]
    assert "model=primary-model" in rendered
```

**After:**
```suggestion
    mock_logger.info.assert_called_once()
    log_args = mock_logger.info.call_args.args
    rendered = log_args[0] % log_args[1:]
    assert "model=primary-model" in rendered
```

</details>

---

### 14. `tests/api/test_mmr_candidate_pruning.py` (L99-L100)

The assertion relies on the pruned list being returned in score-descending insertion order, but `_prune_mmr_candidates_by_bucket` accumulates results via `selected.extend(sorted(...))` and then re-sorts by score — the ordering contract is sound here. However, the real fragility is that the test verifies exactly two elements without asserting which two were dropped. With `text_top_k=1` and `_MMR_CANDIDATE_MULTIPLIER=2`, the pruner should keep `1 * 2 = 2` candidates. If `_MMR_CANDIDATE_MULTIPLIER` is changed (it is a module-level constant with no indirection), this assertion silently breaks (either wrong count or wrong items). Consider either importing and using `_MMR_CANDIDATE_MULTIPLIER` explicitly in the assertion, or parameterising the multiplier so the expected slice is derived rather than hardcoded.

---

### 15. `tests/api/test_mmr_candidate_pruning.py` (L67-L72)

Using a set literal hides ordering, which is intentional for multi-type mixing — good. But note that `_prune_mmr_candidates_by_bucket` always performs a final `selected.sort(key=self._mmr_candidate_score, reverse=True)` on the combined list. The `pref_mem` assertion below does use a list and relies on that sort. These two assertions are inconsistent: either both can assert an ordered list (since the sort is deterministic for distinct scores), or both should use sets. Using a list assertion throughout would catch if the sort contract is accidentally dropped.

---

### 16. `tests/embedders/test_cache.py` (L140)

Using `assert` to wait for `backend_started` is a synchronisation barrier that is stripped under `python -O`, turning the barrier into a no-op. Thread 2 is then submitted before Thread 1 has entered `blocking_embed` and registered an in-flight request, so Thread 2 may start its own independent backend call instead of joining Thread 1's. `singleflight_joins` never reaches 1 and the polling loop below hangs (or, with the deadline assertion also stripped, spins forever). Replace with an explicit check.

**Before:**
```
        assert backend_started.wait(timeout=2)
```

**After:**
```suggestion
        if not backend_started.wait(timeout=2):
            pytest.fail("backend thread did not start within deadline")
```

</details>

---

### 17. `tests/embedders/test_cache.py` (L145)

Using `assert` for deadline enforcement is stripped under `python -O`, turning the bounded wait into an infinite spin. If `singleflight_joins` never reaches 1 (e.g., because Thread 2 raced ahead of Thread 1 due to the stripped barrier above), the test process hangs indefinitely instead of reporting a clear failure.

**Before:**
```
            assert time.monotonic() < deadline
```

**After:**
```suggestion
            if time.monotonic() >= deadline:
                pytest.fail("singleflight join not observed within deadline")
```

</details>

---

### 18. `tests/embedders/test_cache.py` (L132)

Using `assert` to validate `threading.Event.wait` inside a worker thread is stripped under `python -O`. If `release_backend` is never set (because the main thread's polling loop hanged), the 2-second wait returns `False` and the worker silently continues, returning the correct value. The test then passes while the singleflight mechanism was never exercised.

**Before:**
```
        assert release_backend.wait(timeout=2)
```

**After:**
```suggestion
        if not release_backend.wait(timeout=2):
            raise RuntimeError("release_backend was not set within deadline")
```

</details>

---

### 19. `tests/reranker/test_local_rerankers.py` (L101-L105)

`next()` on the bare generator raises `StopIteration` (surfaced by pytest as a confusing error, not a clean assertion failure) when the expected log record is absent — e.g., if `caplog` does not capture the logger used inside `CosineLocalReranker` because its logger is not propagating to the root logger. Provide a sentinel default and assert on it:

```python
message = next(
    (
        record.getMessage()
        for record in caplog.records
        if "CosineLocalReranker rerank result" in record.getMessage()
    ),
    None,
)
assert message is not None, "Expected 'CosineLocalReranker rerank result' log record not found"
```

Memtensor-AI:

All tests passed (57/57 executed). memos_python_core/changed-repo-python: 57/57. Duration: 6s [advisory, non-gating] AI-generated tests on branch test/auto-gen-3d2915ca2b1b0243-20260723102845: 94/99 passed, 5 failed — these do NOT affect the PR verdict; review the branch manually.

**Branch:** `feat/performance-optimization-no-routing-v2.0.25`

bittergreen (approved):

PR #1957, opened 2026-06-21
feat: chunk batch reflection scoring

chiefmojo:
## Problem

Reflection scoring processes all pending traces in a single batch. On stores with large episode backlogs (100+ traces) this produces a single massive LLM call that hits token limits, times out, or produces degraded output quality.

## Changes

Adds a `reflectionBatchSize` config option (default: 20). Reflection scoring splits pending traces into chunks of `reflectionBatchSize` and processes each chunk sequentially, reusing the same LLM client across chunks.

- Zero/negative values fall back to unbounded (existing behavior)
- Each chunk produces its own scored output; results are merged before persisting
- No change to the scoring algorithm or output format

PR #2145, opened 2026-07-23
Revert "feat: chunk batch reflection scoring (#1957)"

syzsunshine219:
Reverts #1957 from main so the change can be submitted again through the intended merge flow.

Validation:
- Reverted tree exactly matches pre-#1957 main commit 554bb98ee7c28307dbaeac569a0dea49ff0062fd
- npm run lint
- vitest run tests/unit/capture/capture-batch.test.ts (9 passed)

### DIFF
--- a/apps/memos-local-plugin/core/capture/ALGORITHMS.md
+++ b/apps/memos-local-plugin/core/capture/ALGORITHMS.md
@@ -78,8 +78,7 @@ priority once reward arrives.
 ## V7 §3.2 batched variant — `batch-scorer.ts`
 
 The per-step path (`reflection-synth.ts` + `alpha-scorer.ts`) issues 2N
-LLM calls per N-step episode. `batch-scorer.ts` collapses up to
-`batchThreshold` steps into one call:
+LLM calls per N-step episode. `batch-scorer.ts` collapses them into ONE:
 
 ```
 inputs   = [{idx, state, action, outcome, reflection, synth_allowed}, …]
@@ -92,8 +91,8 @@ Dispatch (in `capture.ts`):
 | `cfg.batchMode`   | `cfg.batchThreshold` | behavior |
 |-------------------|----------------------|----------|
 | `per_step`        | (ignored)            | legacy: 2N calls |
-| `per_episode`     | chunk size           | batch when `N ≤ threshold`; else chunk-batch |
-| `auto` (default)  | `12`                 | batch when `N ≤ 12`; else chunk-batch |
+| `per_episode`     | (ignored)            | always batch |
+| `auto` (default)  | `12`                 | batch when `N ≤ 12`; else per-step |
 
 The dispatcher also refuses to batch when no LLM is wired — same fallback
 path as missing-LLM in per-step mode.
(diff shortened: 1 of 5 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2145
**Task:** `f1eacf0a238a0e77`
**Base:** `main`
**Head:** `codex/revert-pr-1957-from-main`

---

### 1. `apps/memos-local-plugin/core/capture/batch-scorer.ts` (L170-L173)

Removing `maxTokens` leaves the call relying on the facade's default token limit. The original `batchMaxTokens` computed a scaled budget (768 + N×512, capped at 8 192) for two explicit reasons:
1. **Truncation guard**: A batch of N steps can produce a large JSON payload. If the facade default is lower than what `batchMaxTokens` would have returned (e.g. a common default of 1 024–2 048 tokens), the response gets cut mid-JSON, `validateBatchPayload` throws, the one `malformedRetries` is consumed, and the whole batch falls back to per-step — silently discarding the efficiency gain.
2. **Reasoning-model cap**: The deleted comment called out the 16 k range as triggering avoidable reasoning spend on replay. Without a cap, a reasoning model can now spend unbounded output tokens on a single batch call.

Consider either restoring the explicit budget (`maxTokens: batchMaxTokens(inputs.length)`) or documenting the new contract (e.g. a facade-level default that scales with batch size) so the trade-off is explicit rather than implicit.

**Before:**
```
      validate: (v) => validateBatchPayload(v, inputs.length),
      malformedRetries: 1,
      temperature: 0,
    },
```

**After:**
```suggestion
      validate: (v) => validateBatchPayload(v, inputs.length),
      malformedRetries: 1,
      temperature: 0,
      maxTokens: batchMaxTokens(inputs.length),
    },
```

</details>

---

### 2. `apps/memos-local-plugin/core/capture/capture.ts` (L1075)

When `batchMode` is `"per_episode"`, `shouldBatch` now returns `true` unconditionally for any episode size. This means `runBatchScoring` will receive ALL steps in a single LLM call, no matter how large the episode is. The old `"chunk_batch"` path explicitly split large episodes into `batchThreshold`-sized chunks to stay within the model's context window. With that guard gone (and `batchMaxTokens` also removed from `batch-scorer.ts`), a sufficiently long `per_episode` episode will exceed the model's context limit, resulting in an API error or silent output truncation—both of which silently fall back to lower-quality per-step scoring via the `scored.length === 0` check. Consider adding a step-count guard here (e.g., `if (stepCount > cfg.batchThreshold) return false;`) to preserve safety for large episodes, or document that callers are responsible for keeping episode size under `batchThreshold` when using `"per_episode"` mode.

**Before:**
```
  if (cfg.batchMode === "per_episode") return true;
```

**After:**
```suggestion
  if (cfg.batchMode === "per_episode") {
    // Guard against context-window overflow: even in per_episode mode,
    // fall back to per-step for episodes exceeding batchThreshold.
    return stepCount <= Math.max(1, cfg.batchThreshold);
  }
```

</details>

---

### 3. `apps/memos-local-plugin/core/capture/capture.ts` (L484)

Nested ternary expression. The review checklist explicitly prohibits nested ternaries due to readability and maintainability concerns. Refactor into an explicit conditional assignment before the log call.

**Before:**
```
      mode: useBatch ? "batch" : contextEnabled.includeDownstream ? "per_step_downstream" : "per_step",
```

**After:**
```suggestion
      let scoringMode: string;
      if (useBatch) {
        scoringMode = "batch";
      } else if (contextEnabled.includeDownstream) {
        scoringMode = "per_step_downstream";
      } else {
        scoringMode = "per_step";
      }
      // ... then inside the log call:
      mode: scoringMode,
```

</details>

---

### 4. `apps/memos-local-plugin/core/capture/capture.ts` (L1077)

The original `planScoring` used `Math.max(1, cfg.batchThreshold)` to guard against a `batchThreshold` of `0` or a negative value. `CaptureConfig` declares `batchThreshold` as a plain `number` with no minimum constraint. With the guard removed, if `batchThreshold` is configured as `0` (or any negative number), this expression becomes `stepCount <= 0` which is always `false` for any non-empty episode (the `stepCount === 0` early return already handled the zero case). This silently disables batching in `"auto"` mode for misconfigured deployments, degrading reflection quality with no warning.

**Before:**
```
  return stepCount <= cfg.batchThreshold;
```

**After:**
```suggestion
  return stepCount <= Math.max(1, cfg.batchThreshold);
```

</details>

Memtensor-AI:

**Details:** Executor error: Command failed: git clone --depth 1 --branch codex/revert-pr-1957-from-main git@github.com:MemTensor/MemOS.git /data/test-workspaces/f1eacf0a238a0e77/repo
Cloning into '/data/test-workspaces/f1eacf0a238a0e77/repo'...
warning: Could not find remote branch codex/revert-pr-1957-from-main to clone.
fatal: Remote branch codex/revert-pr-1957-from-main not found in upstream origin
**Branch:** `codex/revert-pr-1957-from-main`

PR #2153, opened 2026-07-23
feat: write-path content-hash dedup for MemoryManager.add

### LINKED ISSUE #2141: [Feature] Write-path dedup: reject identical memory content within a time window (server-side)
## Motivation

MemOS has no server-side dedup on the write path. Any client that re-sends identical memory content (retry loops, replayed events, buggy integrations) creates duplicate nodes — duplicates that then pollute search results, graph structure, and reorganization.

We hit this in production via the OpenClaw cloud plugin (see MemTensor/MemOS-Cloud-OpenClaw-Plugin#145): a replay loop wrote **6,076 memories in 24h, ~90% duplicates** (583 unique contents × 30-40 copies). Client-side dedup helps that specific client, but the server is the right place for a last-line defense that protects all writers.

## Proposal

Content-hash dedup at `MemoryManager.add`:

1. Compute `sha1(memory_text)` for each incoming item
2. Skip items whose hash already exists for the same `user_name` within a configurable window (default 7 days)
3. Also dedup within the batch itself
4. Persist the hash as a `content_hash` node property with a Neo4j index (`memory_content_hash_index`) for O(1) lookup
5. Fail-open: if the dedup check errors, the write proceeds (availability over strictness)
6. Toggles: `MOS_DEDUP_ENABLED=0` to disable, `MOS_DEDUP_WINDOW_DAYS` to adjust the window

Sketch of what we deployed (self-hosted patch, working in production):

```python
def _dedup_by_content_hash(self, memories, user_name):
    kept, seen, skipped = [], set(), 0
    for m in memories:
        h = hashlib.sha1(m.memory.encode("utf-8")).hexdigest()
        if h in seen or self.graph_store.exists_recent_by_hash(user_name, h, days=days):
            skipped += 1
            continue
        m.metadata.content_hash = h  # extra="allow" → flows to node property
        seen.add(h)
        kept.append(m)
    return kept
```

with a matching graph-store query:

```cypher
MATCH (n:Memory {content_hash: $h})
WHERE n.created_at >= datetime() - duration({days: $days})
  AND n.user_name = $user_name
RETURN n.id LIMIT 1
```

## Notes

- Window-based (not permanent) dedup intentionally allows re-learning: the same fact re-stated weeks later is a valid new memory
- The hash doubles as a cheap idempotency key for retried writes
- Happy to contribute this as a PR if the approach looks reasonable

Zhe-SH-CN:
## Description

MemOS has no server-side dedup on the write path. Any client that re-sends identical memory content (retry loops, replayed events, buggy integrations) creates duplicate nodes that pollute search results, graph structure, and reorganization.

This PR adds content-hash deduplication at `MemoryManager.add`:

1. Compute `sha1(memory_text)` for each incoming `TextualMemoryItem`
2. Dedup within the batch using a set
3. Query `graph_store.get_by_metadata` for existing nodes with the same `content_hash`
4. Skip items whose hash exists within a configurable time window (default 7 days)
5. Fail-open: if the dedup check errors, the write proceeds (availability over strictness)
6. Persist `content_hash` as a metadata property on the graph node
7. Toggles: `MOS_DEDUP_ENABLED=0` to disable, `MOS_DEDUP_WINDOW_DAYS` to adjust the window

Related Issue: Fixes #2141

## Type of change

## How Has This Been Tested

- [x] Unit tests covering batch-level dedup, graph-store-level dedup, time window logic, fail-open behavior, empty input, and mixed duplicates
- [x] `python3 -c "import ast; ast.parse(...)"` syntax verification

## Checklist


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2153
**Task:** `258da3fb72ae3edf`
**Base:** `dev-v2.0.24`
**Head:** `feat/write-path-content-hash-dedup`

---

### 1. `src/memos/memories/textual/tree_text_memory/organize/manager.py` (L30)

Unguarded `int()` conversion of an environment variable will raise `ValueError` at module import time if `MOS_DEDUP_WINDOW_DAYS` is set to a non-integer value (e.g., `"7.5"`, `"one week"`, or an empty string `""`), crashing the whole application startup. Wrap it in a try/except or use a helper that falls back to the default.

**Before:**
```
_DEDUP_WINDOW_DAYS = int(os.getenv("MOS_DEDUP_WINDOW_DAYS", "7"))
```

**After:**
```suggestion
try:
    _DEDUP_WINDOW_DAYS = int(os.getenv("MOS_DEDUP_WINDOW_DAYS", "7"))
except ValueError:
    logger.warning(
        "Invalid MOS_DEDUP_WINDOW_DAYS value %r; falling back to 7 days",
        os.getenv("MOS_DEDUP_WINDOW_DAYS"),
    )
    _DEDUP_WINDOW_DAYS = 7
```

</details>

---

### 2. `src/memos/memories/textual/tree_text_memory/organize/manager.py` (L152)

Timezone-aware vs naive datetime mismatch. `datetime.now()` returns a naive datetime, but `datetime.fromisoformat(updated)` returns a timezone-aware datetime when the stored string includes a UTC offset (e.g., `"2026-07-20T10:00:00+00:00"`). Comparing the two raises `TypeError`, which is swallowed by the outer `except Exception` block and treated as fail-open — silently disabling dedup for every node that has a timezone-aware timestamp. Use `datetime.now(timezone.utc)` and ensure the parsed timestamp is also timezone-aware.

**Before:**
```
                            if ts and ts >= cutoff:
```

**After:**
```suggestion
from datetime import datetime, timedelta, timezone

# ...
cutoff = datetime.now(timezone.utc) - timedelta(days=_DEDUP_WINDOW_DAYS)
# ...
                            try:
                                ts = datetime.fromisoformat(updated)
                                # Make naive timestamps UTC-aware for a safe comparison
                                if ts.tzinfo is None:
                                    ts = ts.replace(tzinfo=timezone.utc)
                            except (ValueError, TypeError):
                                ts = None
                            if ts and ts >= cutoff:
```

</details>

---

### 3. `src/memos/memories/textual/tree_text_memory/organize/manager.py` (L169-L176)

The fallback branch `m.metadata.info = {"content_hash": h}` silently replaces the entire `info` dict with a new single-key dict, destroying any pre-existing metadata that was stored there. `TextualMemoryMetadata.info` is a real `dict | None` field used for arbitrary key-value metadata. If `m.metadata.info` is not a dict (e.g., `None`), the correct fix is to initialize it as `{"content_hash": h}`, but if it already contains data, only the `content_hash` key should be added or updated.

Additionally, since `TextualMemoryMetadata` uses `model_config = ConfigDict(extra="allow")`, the primary `m.metadata.content_hash = h` assignment on Pydantic models will always succeed — making the entire `except` block unreachable dead code.

**Before:**
```
            try:
                m.metadata.content_hash = h
            except (AttributeError, TypeError):
                # Metadata model may not support direct attribute set.
                if hasattr(m.metadata, "info") and isinstance(m.metadata.info, dict):
                    m.metadata.info["content_hash"] = h
                else:
                    m.metadata.info = {"content_hash": h}
```

**After:**
```suggestion
            # TextualMemoryMetadata uses extra="allow", so direct assignment always works.
            m.metadata.content_hash = h
```

</details>

Memtensor-AI:

All tests passed (8/8 executed). memos_python_core/changed-repo-python: 8/8. Duration: 5s [advisory, non-gating] AI-generated tests on branch test/auto-gen-258da3fb72ae3edf-20260723200033: 112/120 passed, 8 failed — these do NOT affect the PR verdict; review the branch manually.

**Branch:** `feat/write-path-content-hash-dedup`

WeiminLee (approved):
ok

WeiminLee (changes requested):
I need to change my previous approval here. This PR is not ready to merge in its current state.

Blocking findings:

1. The branch is currently conflicting against the base branch. `gh` reports `mergeable=CONFLICTING`, and a merge-tree shows conflicts in several unrelated files including `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py`, `pyproject.toml`, `src/memos/__init__.py`, `src/memos/api/client.py`, `src/memos/log.py`, and tests. Please rebase/update the branch and reduce the PR to the intended content-hash dedup change.

2. `src/memos/memories/textual/tree_text_memory/organize/manager.py:134` calls `graph_store.get_by_metadata(filters=[{"content_hash": h}])`, but existing graph backends expect filters shaped like `{"field": "content_hash", "value": h}`. With Neo4j/Postgres this raises inside `get_by_metadata`, then the dedup code fail-opens and proceeds with the write, so the advertised graph-store-level duplicate detection does not work. The tests use a loose mock and do not catch the real backend contract.

3. `src/memos/memories/textual/tree_text_memory/organize/manager.py:139` compares `datetime.now()` with `datetime.fromisoformat(updated)`. If stored timestamps include a timezone offset, Python raises `TypeError` when comparing aware and naive datetimes; the outer fail-open path then silently disables dedup for those existing nodes.

4. `src/memos/memories/textual/tree_text_memory/organize/manager.py:30` parses `MOS_DEDUP_WINDOW_DAYS` with `int()` at import time. A bad env value crashes module import and application startup. Please parse defensively and fall back with a warning.

Please add regression tests that exercise the real `get_by_metadata` filter shape and timezone-aware timestamps, not only MagicMock return values.

PR #2178, opened 2026-07-28
fix: playground service unavaliable occasionally and information loss resulting from truncation

endxxxx:
## Description

This PR is created to fix some bug in the playground of MemOS official website, like unavailable service occasionally and information loss resulting from truncation due to excessively long content in the knowledge base memory. In addition, this PR modifies how Little M (小忆) answers user questions regarding character perception within the playground.

## Type of change

## How Has This Been Tested?

Using Playground in the official website of MemOS

## Checklist

## Reviewer Checklist
- [x] Made sure Checks passed

### DIFF
--- a/.github/scripts/publish-local-plugin.sh
+++ b/.github/scripts/publish-local-plugin.sh
@@ -1,0 +1,170 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+: "${RUNNER_TEMP:?RUNNER_TEMP is required.}"
+: "${PACKAGE_NAME:?PACKAGE_NAME is required.}"
+: "${RELEASE_VERSION:?RELEASE_VERSION is required.}"
+: "${RELEASE_TAG:?RELEASE_TAG is required.}"
+: "${NPM_DIST_TAG:?NPM_DIST_TAG is required.}"
+
+npm_visibility_attempts="${NPM_VISIBILITY_ATTEMPTS:-10}"
+npm_ambiguous_visibility_attempts="${NPM_AMBIGUOUS_VISIBILITY_ATTEMPTS:-3}"
+npm_visibility_delay_seconds="${NPM_VISIBILITY_DELAY_SECONDS:-5}"
+
+validate_positive_integer() {
+  local name="$1"
+  local value="$2"
+  if ! [[ "${value}" =~ ^[1-9][0-9]*$ ]]; then
+    echo "::error::${name} must be a positive integer; received ${value}."
+    exit 2
+  fi
+}
+
+validate_non_negative_integer() {
+  local name="$1"
+  local value="$2"
+  if ! [[ "${value}" =~ ^[0-9]+$ ]]; then
+    echo "::error::${name} must be a non-negative integer; received ${value}."
+    exit 2
+  fi
+}
+
+validate_positive_integer "NPM_VISIBILITY_ATTEMPTS" "${npm_visibility_attempts}"
+validate_positive_integer "NPM_AMBIGUOUS_VISIBILITY_ATTEMPTS" "${npm_ambiguous_visibility_attempts}"
+validate_non_negative_integer "NPM_VISIBILITY_DELAY_SECONDS" "${npm_visibility_delay_seconds}"
+
+script_directory="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
+npm_view_log="${RUNNER_TEMP}/memos-local-plugin-npm-view.log"
+
+npm_version_exists() {
+  local attempt
+  local status
+
+  for attempt in 1 2 3; do
+    set +e
+    npm view "${PACKAGE_NAME}@${RELEASE_VERSION}" version --prefer-online >"${npm_view_log}" 2>&1
+    status=$?
+    set -e
+    if [ "${status}" = 0 ]; then
+      sed -n '1,40p' "${npm_view_log}"
+      return 0
+    fi
+    if grep -Eiq "E404|404 Not Found|No match found|is not in this registry" "${npm_view_log}"; then
+      return 1
+    fi
+    sed -n '1,120p' "${npm_view_log}"
+    if [ "${attempt}" = 3 ]; then
+      echo "::error::npm view failed after three attempts; refusing to guess whether ${PACKAGE_NAME}@${RELEASE_VERSION} exists."
+      exit "${status}"
+    fi
+    sleep "$((attempt * 5))"
+  done
+}
+
+wait_for_npm_version() {
+  local attempts="$1"
+  local attempt
+  local delay
+
+  for attempt in $(seq 1 "${attempts}"); do
+    if npm_version_exists; then
+      echo "${PACKAGE_NAME}@${RELEASE_VERSION} became visible on attempt ${attempt}/${attempts}."
+      return 0
+    fi
+    if [ "${attempt}" = "${attempts}" ]; then
+      return 1
+    fi
+
+    delay=$((npm_visibility_delay_seconds * attempt))
+    if [ "${delay}" -gt 30 ]; then
+      delay=30
+    fi
+    echo "::notice::${PACKAGE_NAME}@${RELEASE_VERSION} is not visible yet; retrying in ${delay}s."
+    if [ "${delay}" -gt 0 ]; then
+      sleep "${delay}"
+    fi
+  done
+
+  return 1
+}
+
+remote_tag_exists() {
+  local release_tag="$1"
+  local attempt
+  local status
+
+  for attempt in 1 2 3; do
+    set +e
+    git ls-remote --exit-code --tags origin "refs/tags/${release_tag}" >/dev/null 2>&1
+    status=$?
+    set -e
+    if [ "${status}" = 0 ]; then
+      return 0
+    fi
+    if [ "${status}" = 2 ]; then
+      return 1
+    fi
+    if [ "${attempt}" = 3 ]; then
+      echo "::error::Failed to check remote tag ${release_tag} after three attempts."
+      exit "${status}"
+    fi
+    sleep "$((attempt * 5))"
+  done
+}
+
+if npm_version_exists; then
+  if remote_tag_exists "${RELEASE_TAG}"; then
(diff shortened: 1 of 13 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2178
**Task:** `ee9188b923d75b9d`
**Base:** `dev-v2.0.27`
**Head:** `hotfix/playground-bug`

🔍 **OpenCodeReview** found **10** issue(s) in this PR.

---

### 1. `.github/scripts/publish-local-plugin.sh` (L39-L62)

**Hard-coded retry count ignores the configurable `NPM_AMBIGUOUS_VISIBILITY_ATTEMPTS` knob.**

`npm_version_exists` always retries exactly 3 times regardless of the `NPM_AMBIGUOUS_VISIBILITY_ATTEMPTS` variable that is validated and documented at the top of the script. The downstream `wait_for_npm_version` correctly receives that variable as an argument, but the inner per-attempt existence check bypasses it entirely.

This inconsistency matters in practice: if a transient 5xx error makes the initial pre-publish existence check uncertain, the script will hard-stop after 3 retries instead of honouring the operator's configured count. An operator who sets `NPM_AMBIGUOUS_VISIBILITY_ATTEMPTS=10` to handle a flaky registry will get no benefit during the pre-publish guard.

**Suggestion:** Parameterise the function so the caller controls the retry count:
```bash
npm_version_exists() {
  local max_attempts="${1:-3}"
  local attempt
  local status

  for attempt in $(seq 1 "${max_attempts}"); do
    ...
  done
}
```
And at the call sites:
```bash
npm_version_exists "${npm_ambiguous_visibility_attempts}" # pre-publish check
wait_for_npm_version "${npm_visibility_attempts}"        # post-publish polling
```

---

### 2. `.github/scripts/publish-local-plugin.sh` (L88-L89)

**Dead `return 1` at the end of `wait_for_npm_version` is unreachable.**

The `for` loop on line 69 terminates every possible path before reaching line 88:
- If `npm_version_exists` succeeds → `return 0` exits the function.
- If `attempt == attempts` on the last iteration → `return 1` exits the function from inside the loop.

The trailing `return 1` at line 88 can never execute. While harmless today, it creates a false impression that there is a loop-exit path that falls through, and it hints that the last-iteration guard (`if [ "${attempt}" = "${attempts}" ]`) was an afterthought rather than the primary exit. Future maintainers restructuring the loop may remove the in-loop guard and accidentally rely on the trailing return — which would be correct but obscures intent.

**Suggestion:** Remove the trailing `return 1` and document the in-loop early exit:
```bash
    if [ "${attempt}" = "${attempts}" ]; then
      return 1  # exhausted all attempts
    fi
  done
  # (loop always returns via the in-loop guards above)
}
```

---

### 3. `.github/workflows/memos-local-plugin-publish.yml` (L136-L138)

The `working-directory: .` override on the test step is functionally correct (it overrides the job-level default of `apps/memos-local-plugin` to the repo root, allowing `node --test .github/scripts/publish-local-plugin.test.mjs` to resolve the path). However, the test step is placed **before** `actions/setup-node` installs `node_modules`, which is fine since it only uses Node.js built-ins (`node:test`, `node:fs`, `node:child_process`, etc.) with no npm dependencies.

The real concern is that the test step runs `git ls-remote` via the publish script's `remote_tag_exists()` function if the pre-publish `npm_version_exists()` check ever returns true in a test scenario. Currently, none of the three test scenarios exercise that code path (all begin with an E404 mock), so the `git ls-remote` call is never reached. This means the idempotency / recovery code path in `publish-local-plugin.sh` has **zero test coverage**. A regression in that branch would go undetected.

Suggestion: Add a test scenario that mocks both `npm view` (returning success on the first call) and `git ls-remote` (via a mock binary in `binDirectory`) to cover the `npm_version_exists → true` branch.

---

### 4. `src/memos/api/handlers/formatters_handler.py` (L204)

Potential `IndexError` if `sources` is an empty list. The old code had the same direct access, but the new logic now conditionally builds a combined string from `source_content`, making this a real risk if any item has no sources. You should guard against an empty list before indexing.

Suggested fix: fall back gracefully when `sources` is empty.

**Before:**
```
        source_content = item["metadata"]["sources"][0]["content"]
```

**After:**
```suggestion
        sources = item["metadata"].get("sources", [])
        source_content = sources[0]["content"] if sources else ""
```

</details>

---

### 5. `.github/scripts/publish-local-plugin.sh` (L96-L112)

**`exit "${status}"` in `remote_tag_exists` may emit a non-POSIX exit code (e.g. 128) that misrepresents the failure type.**

`git ls-remote` exits with code 128 on authentication/configuration errors (e.g. missing credentials, unknown remote). The current logic treats any exit code other than 0 or 2 as a transient error and retries. After three failures it calls `exit "${status}"`, which propagates the raw git exit code (128) to the CI runner.

Two problems:
1. **POSIX only guarantees meaningful interpretation of codes 0–125.** Codes ≥ 126 have special meanings in shells (126 = not executable, 127 = command not found, 128+ = signal termination). Exiting with 128 from a Bash script can confuse CI systems or callers that inspect the code.
2. **Authentication failures (128) are not transient** and will never succeed on retry, wasting up to 45 seconds of retry time before producing an opaque error.

**Suggestion:** Distinguish fatal errors from transient ones and normalise the exit code:
```bash
    # git exits 128 for fatal errors (auth failure, bad remote, etc.) — non-recoverable
    if [ "${status}" -ge 128 ]; then
      echo "::error::git ls-remote failed fatally (exit ${status}) checking ${release_tag}; not retrying."
      exit 1
    fi
    if [ "${attempt}" = 3 ]; then
      echo "::error::Failed to check remote tag ${release_tag} after three attempts."
      exit 1  # normalise to a conventional error code
    fi
```

---

### 6. `.github/scripts/publish-local-plugin.test.mjs` (L57-L59)

The bare `catch {}` silently swallows all errors, including non-ENOENT ones such as `EACCES` (permission denied). If a counter file exists but is unreadable, `readCounter` returns 0, causing assertions like `assert.equal(result.publishCount, 3)` to report a misleading failure (0 ≠ 3) that hides the real I/O problem. Narrow the catch to `ENOENT` only so unexpected filesystem errors surface immediately.

**Before:**
```
  } catch {
    return 0;
  }
```

**After:**
```suggestion
  } catch (err) {
    if (err.code === "ENOENT") return 0;
    throw err;
  }
```

</details>

---

### 7. `.github/scripts/publish-local-plugin.test.mjs` (L48)

Unquoted `$*` is used here. While inside a `case` catch-all it is only cosmetic, the conventional shell idiom is `"$*"` (or `"$@"`) to prevent unintended word-splitting or glob expansion on the argument list.

**Before:**
```
    echo "Unexpected npm command: $*" >&2
```

**After:**
```suggestion
    echo "Unexpected npm command: $*" >&2
```

</details>

---

### 8. `.github/scripts/publish-local-plugin.sh` (L118)

**Case-sensitive `= "true"` check silently rejects valid truthy inputs like `True`, `TRUE`, `yes`, `1`.**

If an operator sets `RECOVER_EXISTING_NPM_RELEASE=True` (e.g. via a GitHub Actions boolean input that YAML serialises differently), the comparison fails and the script exits with a confusing error: `"npm version exists but ${RELEASE_TAG} does not. Refusing to invent release metadata without explicit recovery mode."` — even though recovery mode was intended.

The GitHub Actions `inputs.recover_existing_npm_release` typed as `boolean` is always lowercased (`true`/`false`) in the runner environment, so this is not a current bug in the primary workflow. However, the script provides no documentation of the accepted format, making it fragile to future workflow changes or direct invocations.

**Suggestion:** Either normalise the value, or add a clear validation message:
```bash
recover_mode="$(echo "${RECOVER_EXISTING_NPM_RELEASE:-false}" | tr '[:upper:]' '[:lower:]')"
...
elif [ "${recover_mode}" = "true" ]; then
```
Or at minimum add a comment:
```bash
# RECOVER_EXISTING_NPM_RELEASE must be exactly the lowercase string "true" to enable recovery.
elif [ "${RECOVER_EXISTING_NPM_RELEASE:-false}" = "true" ]; then
```

---

### 9. `.github/scripts/publish-local-plugin.sh` (L37)

**Fixed log path under `RUNNER_TEMP` is susceptible to race conditions in parallel job runs.**

If this script is ever executed concurrently (e.g. in a matrix build or two simultaneous dry-run triggers that share the same ephemeral runner and thus the same `RUNNER_TEMP`), both jobs write to the identical path `${RUNNER_TEMP}/memos-local-plugin-npm-view.log`. This causes log output from one job to overwrite or interleave with the other, potentially making `grep` checks on the log file produce incorrect results — in the worst case causing a false "version already exists" or false "404" classification.

The test file correctly isolates each invocation by setting `RUNNER_TEMP` to a unique temp directory per test, so the issue is masked in tests.

**Suggestion:** Append the PID (or `GITHUB_RUN_ATTEMPT`) to the filename to make it unique per invocation:
```bash
npm_view_log="${RUNNER_TEMP}/memos-local-plugin-npm-view-$$.log"
```

---

### 10. `.github/scripts/publish-local-plugin.sh` (L127-L129)

**`publish_status` is not declared with `local` (N/A here) but more importantly is not initialised before the loop, creating a theoretical unbound-variable risk under `set -u` in sourced contexts.**

Although this script is not designed to be sourced, under `set -u` (already active via `set -euo pipefail`), any reference to an unset variable is a fatal error. `publish_status` is only ever read *after* being assigned on line 132 within the same loop body, so the current code is safe. However, the variable is never initialised before the loop (`publish_status=` or `publish_status=1`), meaning that if the loop structure changes (e.g. a `continue` or `break` is added before the assignment) a future maintainer could accidentally introduce an unbound-variable dereference that causes an opaque `unbound variable` exit.

This is a defensive coding gap rather than an active bug.

**Suggestion:** Initialise `publish_status` before the loop:
```bash
publish_accepted=false
publish_status=1  # initialise to failure; overwritten on each attempt

for attempt in 1 2 3; do
```

Memtensor-AI:

All tests passed (15/15 executed, 1 skipped). memos_github_open_source/smoke: 1/1, memos_local_plugin/unit: 2 passed, 1 skipped, memos_python_core/changed-repo-python: 12/12. Duration: 13s

**Branch:** `hotfix/playground-bug`

wustzdy (approved):

Issue #2121, opened 2026-07-19
[Bug] memos-local-plugin: embedding source text exceeds embedding-3's 3072-token limit → HTTP 400 (code:1210), entire batch fails


hitken:
## Pre-submission checklist | 提交前检查
- [x] I have searched existing issues and this hasn't been mentioned before
- [x] I have read the project documentation and confirmed this issue doesn't already exist
- [x] This issue is specific to MemOS and not a general software issue

## Bug Description | 问题描述

`memos-local-plugin` v2.0.10 feeding **untruncated/un-chunked trace text** to the embedding provider causes HTTP 400 errors. When using 智谱 `embedding-3` (single-request cap 3072 tokens), any `traces.summary` or `agent_text` longer than ~6KB triggers `code:1210 API 调用参数有误`, and because the batch is aborted on the first failure (`failed = batch.length`), **the entire batch — including short, valid entries — is marked failed**. The "修复向量 / rebuild vectors" button in the Memory Viewer returns 400 immediately, and the periodic embedding heartbeat probe shows ~60% error rate.

## Environment | 环境信息
- Plugin: `@memtensor/memos-local-plugin` **v2.0.10** (also confirmed in v2.0.6)
- Embedding provider: `openai_compatible` → `https://open.bigmodel.cn/api/paas/v4/embeddings`, model `embedding-3`
- 智谱 embedding-3 official limits: **8K context window, 3072 tokens per single input, ≤64 inputs per batch** ([docs](https://docs.bigmodel.cn/cn/guide/models/embedding/embedding-3))
- Storage: SQLite (`memos.db`), 19,114 traces
- Host: macOS arm64 (OpenClaw runtime)

### 1. DB-side: source-text length distribution

```sql
-- field                       count   max bytes
-- agent_text > 6k chars:      19      59644
-- summary     > 6k chars:     40      59644
-- user_text   > 6k chars:     21      22430
```

So ~40 traces exceed the 6KB (~3000 Chinese-char) mark. The longest `summary` and `agent_text` are **59,644 bytes each** — roughly 30k Chinese characters, an order of magnitude over the 3072-token cap.

### 2. API-side: reproducing the 400

```bash
# short input → 200 OK
curl -X POST https://open.bigmodel.cn/api/paas/v4/embeddings \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"embedding-3","input":"hello world"}'        # HTTP 200

[...]

PR #2123, opened 2026-07-19
Fix #2121: [Bug] memos-local-plugin: embedding source text exceeds embedding-3's 3072-token

Memtensor-AI:
## Description

Fixed issue #2121: memos-local-plugin no longer crashes the whole rebuild-embedding batch when a single trace exceeds the provider's per-input token limit (e.g. 智谱 embedding-3's 3072-token cap → HTTP 400 code:1210).

Three surgical fixes in apps/memos-local-plugin:

1. **Embedder facade truncation** (`core/embedding/embedder.ts` + `types.ts`): added `EmbeddingConfig.maxInputChars` (default 6000; `0` disables). Truncation runs BEFORE cache-key hashing so every provider is protected and repeat calls with the same head text hit the LRU. One `input_truncated` warn is emitted per embedMany call (batched, not per-input).

2. **rebuildEmbeddings divide-and-conquer** (`core/pipeline/memory-core.ts`): replaced the blanket `failed = batch.length` catch with a halving retry. When a sub-batch is size 1, only that slot is marked failed; short valid rows next to a poison input are no longer punished. Existing `EmbeddingMaintenanceRunResult` contract preserved.

3. **Response body in log** (`core/embedding/fetcher.ts`): `http.non_ok` warn now carries a truncated response body (first 512 chars) so operators can pinpoint provider error codes (e.g. code:1210) from gateway.log alone.

Also exposed `embedding.maxInputChars` on the user config schema (default 6000, min 0). Chunked embedding (issue's fix #2) is deferred as follow-up per proposal.md.

Tests: 5 new embedder cases + 2 new fetcher cases + 1 new memory-core case (all strict TDD). `pnpm --filter @memtensor/memos-local-plugin vitest run tests/unit/embedding tests/unit/pipeline/memory-core.test.ts tests/unit/config` = 144 passed / 1 skipped. `tsc --noEmit` clean. Integration tests green. Two pre-existing unit failures (startup-recovery, storage/migrator) reproduce on the base commit and are unrelated to embedding.

opsp artifacts (task.md + proposal.md + design.md + verification-report.md) archived to memos-autodev-specs/2026-07-19-2121-bug-memos-local-plugin-embedding-source-text-exceeds-embeddi/ (pushed to main).

Related Issue (Required):  Fixes #2121

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
[...]

PR #2190, opened 2026-07-31
fix(embedder): safe 3072-token default + token-aware truncation

RerankerGuo:
## Background

Issue #2121 reports that `memos-local-plugin` ingestion passes source text directly to the embedding API, and text-embedding-3 enforces a hard 3072-token input cap. Previously, the embedder's default `max_tokens` was 8192 (which exceeds the API limit) and the truncation was a plain character-slice that did not account for token counts at all. This caused out-of-budget embedding errors for longer documents and knowledge chunks.

## Changes

- **`memos/embedders/base.py`**:
  - Added `_SAFE_EMBEDDING_MAX_TOKENS = 3072` matching text-embedding-3's input limit.
  - Added `BaseEmbedder._effective_max_tokens()`: returns an explicitly configured positive `max_tokens`, otherwise the 3072 safe default (also used when the user configures `max_tokens=None` or `max_tokens=0`).
  - Rewrote `_truncate_texts()` to delegate to the existing token-aware `_truncate_text_to_tokens()` binary-search helper. A fast-path length check keeps the cheap case cheap (text length <= max tokens means no token count required).
- **`memos/configs/embedder.py`**:
  - Changed `BaseEmbedderConfig.max_tokens` default from `8192` to `None` so that the new 3072 safe-limit takes effect for any user not overriding the field explicitly; the docstring was updated.
- **`tests/embedders/test_base.py`**:
  - Added coverage for `_effective_max_tokens()` behaviour across `None`, `0`, and explicit positive values.
  - Added coverage for `_truncate_text_to_tokens()`: short text unchanged, empty/None/0 boundaries, long CJK text stays within budget.
  - Added coverage for `_truncate_texts()`: short texts untouched, very long CJK text truncated within the 3072-token default, explicit 10-token override honoured.

## Verification

- Ruff: `ruff format` + `ruff check --fix` applied (2 pre-existing BLE001 broad-excepts in `_count_tokens_for_embedding` left intact).
- `python3 -m py_compile` succeeds for modified files.

## Impact

- Backwards compatible for callers that set an explicit `max_tokens`; behaviour changes only when the field was left at the previous 8192 default.
- Eliminates the silent "send 8192 chars and hope for the best" behaviour for text-embedding-3 endpoints; the default now stays inside the API budget.

### DIFF
--- a/src/memos/configs/embedder.py
+++ b/src/memos/configs/embedder.py
@@ -13,8 +13,8 @@ class BaseEmbedderConfig(BaseConfig):
         default=None, description="Number of dimensions for the embedding"
     )
     max_tokens: int | None = Field(
-        default=8192,
-        description="Maximum number of tokens per text. Texts exceeding this limit will be automatically truncated. Set to None to disable truncation.",
+        default=None,
+        description="Maximum number of tokens per text. When None (default) the embedder falls back to a provider-aware safe limit (currently 3072 tokens, matching text-embedding-3). Set to 0 or a concrete positive integer to explicitly disable or override truncation.",
     )
     headers_extra: dict[str, Any] | None = Field(
         default=None,
--- a/src/memos/embedders/base.py
+++ b/src/memos/embedders/base.py
@@ -13,6 +13,8 @@ from memos.log import get_logger, text_hash
 logger = get_logger(__name__)
 EmbeddingCallable = TypeVar("EmbeddingCallable", bound=Callable[..., Any])
 
+_SAFE_EMBEDDING_MAX_TOKENS: int = 3072
+
 
 def log_embedding_call(func: EmbeddingCallable) -> EmbeddingCallable:
     """Log embedding request dimensions and timing without text or vectors."""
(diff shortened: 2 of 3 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2190
**Task:** `61ebe3a4627d2a33`
**Base:** `main`
**Head:** `fix/issue-2121-embedding-3072-token-truncate`

---

### 1. `src/memos/embedders/base.py` (L171-L173)

The fast-path assumes that `len(t) <= max_tokens` guarantees the token count is also within budget, citing CJK as the worst case (~1 char/token). This assumption is incorrect for emoji and rare Unicode.

With tiktoken's `cl100k_base`, emoji such as U+1F600 (😀) encode to **2 tokens** each, and rare code points that fall back to byte-level tokens can produce even more. A text composed of 3072 emoji characters has `len(t) == 3072`, passes the fast-path unchanged, yet has ~6144 actual tokens — far above the `max_tokens` limit.

Suggested fix: use a tighter bound, or skip the fast-path entirely and always call `_count_tokens_for_embedding` when you need a correct answer:
```python
# Conservative: only skip counting when the text is very short
# (well under half of max_tokens to account for high-token-density chars)
if len(t) * 2 <= max_tokens:   # even emoji won't exceed budget
    truncated.append(t)
    continue
```
or simply remove the fast-path and rely on the early-exit inside `_truncate_text_to_tokens` (line 110–111) which already skips truncation when `current_tokens <= max_tokens`.

---

### 2. `src/memos/embedders/base.py` (L191)

This fallback unconditionally applies a 3072-token truncation limit to **every** embedder that has not explicitly set `max_tokens` in its config — including models that support 8192 or more tokens (e.g. `text-embedding-ada-002`, many open-source models). Previously the code skipped truncation entirely when `max_tokens` was `None`, so this is a silent breaking change that can degrade embedding quality for longer documents with no warning to the user.

A safer approach is to make this opt-in rather than opt-out, or at least document clearly that subclasses for higher-capacity models should override `_effective_max_tokens` or always set `max_tokens` in their config:
```python
def _effective_max_tokens(self) -> int | None:
    config = getattr(self, "config", None)
    if config is not None:
        configured = getattr(config, "max_tokens", None)
        if configured is not None and configured > 0:
            return configured
    # Return None to opt out of truncation for unconfigured embedders;
    # subclasses for token-limited APIs (e.g. text-embedding-3) should
    # override this or set max_tokens in their config.
    return None
```

---

### 3. `tests/embedders/test_base.py` (L22-L23)

_ConcreteEmbedder.__init__ does nothing but delegate to super().__init__(config) with the same signature. Python inherits the parent __init__ automatically when no override is defined, making this method dead code. Remove it entirely; the class only needs the embed override.

**Before:**
```
    def __init__(self, config):
        super().__init__(config)
```

**After:**
```suggestion
class _ConcreteEmbedder(BaseEmbedder):
    def embed(self, texts: list[str]) -> list[list[float]]:
        return [[0.0] for _ in texts]
```

</details>

---

### 4. `tests/embedders/test_base.py` (L70-L71)

The lower-bound check measures characters, not tokens, implicitly relying on a 1-char-per-token CJK assumption. If the tokeniser encodes any CJK subsequence as multiple tokens, truncation may correctly stop below `limit` characters while still satisfying the token budget, causing a spurious test failure. Both assertions should use the same counting function.

**Before:**
```
        assert _count_tokens_for_embedding(truncated) <= limit
        assert len(truncated) >= limit
```

**After:**
```suggestion
        assert _count_tokens_for_embedding(truncated) <= limit
        assert _count_tokens_for_embedding(truncated) >= int(limit * 0.9)
```

</details>

---

### 5. `tests/embedders/test_base.py` (L36-L37)

This test conflates two independent concerns: (a) that None max_tokens falls back to the safe constant, and (b) that the constant equals exactly 3072. If the constant is intentionally changed, the failure will point at a test named 'falls_back_to_safe_default', making the root cause harder to diagnose. Move the constant-value assertion to its own dedicated test.

**Before:**
```
        assert emb._effective_max_tokens() == _SAFE_EMBEDDING_MAX_TOKENS
        assert _SAFE_EMBEDDING_MAX_TOKENS == 3072
```

**After:**
```suggestion
        assert emb._effective_max_tokens() == _SAFE_EMBEDDING_MAX_TOKENS


def test_safe_embedding_max_tokens_constant_value():
    assert _SAFE_EMBEDDING_MAX_TOKENS == 3072
```

</details>

---

Memtensor-AI:

Automated tests inconclusive (auto-generated test defect); treated as non-blocking. Manual review recommended. Details: The AI-generated test file defines a `_ConcreteEmbedder` subclass of `BaseEmbedder` but fails to implement the abstract `__init__` method, causing all instantiation-based tests to fail with TypeError. Additionally, one CJK truncation test has an off-by-one assertion that expects the truncated length to be >= the token limit, but token-aware truncation correctly produces a length <= limit. [advisory, non-gating] AI-generated tests on branch test/auto-gen-bc3c30a4c517a246-20260731105109: 24/65 passed, 41 failed — these do NOT affect the PR verdict; review the branch manually.

**Branch:** `fix/issue-2121-embedding-3072-token-truncate`

Memtensor-AI:

Automated tests inconclusive (auto-generated test defect); treated as non-blocking. Manual review recommended. Details: The test asserts that a token-truncated CJK string has a character length >= the token limit (500), but token-aware truncation legitimately returns fewer characters than the token budget when the tokenizer produces a boundary just below the limit. [advisory, non-gating] AI-generated tests on branch test/auto-gen-61ebe3a4627d2a33-20260803120352: 58/58 passed — these do NOT affect the PR verdict; review the branch manually.

**Branch:** `fix/issue-2121-embedding-3072-token-truncate`

RerankerGuo:
Closing as a semantic duplicate of #2123, which is the earlier active implementation for issue #2121. This branch also has unresolved correctness findings around Unicode token density and the unconditional 3072-token fallback, so it should not compete with the assigned implementation.

PR #2196, opened 2026-08-03
fix(plugin): improve recall relevance and host input handling

Hun-ger:
## Description

This PR improves retrieval relevance and host-context handling in `apps/memos-local-plugin` for both OpenClaw and Hermes.

It fixes the following problems:

- Retrieval candidates could be dominated by weak keyword matches, duplicated question traces, priority/recency signals, or unrelated memory tiers.
- An empty LLM relevance-filter result could be mechanically refilled with unrelated memories.
- OpenClaw IM integrations could pass channel metadata such as `message_id` and sender IDs into memory search and capture.
- OpenClaw internal heartbeat and asynchronous command-completion messages could incorrectly trigger memory search.
- Excluding an entire active session prevented older same-session memories from being recalled after OpenClaw or Hermes compressed the visible conversation history.
- Retrieval logs did not always reflect the actual candidate ordering, score components, and post-LLM result count.

Implementation approach:

- Build semantic queries only from normalized user text; `contextHints` remain routing and visibility metadata.
- Correct vector-channel scoring and expose score components in retrieval diagnostics.
- Add a personal-fact retrieval profile that prioritizes trace evidence and reduces unrelated skill, world-model, and recency/priority influence.
- Apply conservative rule-based candidate deduplication after scoring/RRF and before thresholding/MMR, allowing removed duplicates to release Top-K capacity.
- Preserve candidates with distinct evidence, exact-identifier proof, keyword-channel eligibility, or multi-channel threshold-bypass capabilities.
- Keep embedding-based diversity control in MMR and tolerate mixed embedding dimensions during embedding migrations.
- Honor the LLM relevance filter when it returns an empty result instead of mechanically refilling unrelated candidates.
- Use OpenClaw official `message_received.content`, correlated by `runId`, as the primary clean user input. Legacy envelope sanitization remains as a compatibility fallback.
- Ignore OpenClaw heartbeat, internal wakeup, and asynchronous command-completion messages before retrieval and capture.
- Pass the host visible conversation boundary into retrieval so only same-session traces still visible to the model are excluded; older compacted traces become eligible for recall.
- Keep the Hermes compression callback read-only while recording its visible-context boundary for subsequent retrieval.

Compatibility and dependencies:

- No new runtime dependencies.
- No database schema migration is required.
- No embedding model or embedding dimension change is required.
- Existing memories and embeddings remain compatible.
- Before the first known compression boundary, Hermes preserves the previous whole-session exclusion behavior.

Related Issue (Required): Fixes #2195

[...]

PR #2198, opened 2026-08-03
ci: keep local plugin standalone publishing compatible with MemOS releases

MLittleprince:
## Summary
- keep MemOS Release — Publish as the weekly whole-repo release/docs fallback path
- keep the legacy local-plugin publisher usable for beta/non-latest package releases without Doc Agent/docs payloads
- keep legacy latest releases docs-capable, while selecting the previous stable local-plugin tag instead of a beta tag for docs evidence

## Validation
- node --test .github/scripts/draft-local-plugin-release-notes.test.mjs
- node --test .github/scripts/prepare-memos-release.test.mjs
- ruby workflow YAML parse for .github/workflows/*.yml
- local package-only beta release-notes simulation

### DIFF
--- a/.github/scripts/draft-local-plugin-release-notes.mjs
+++ b/.github/scripts/draft-local-plugin-release-notes.mjs
@@ -37,7 +37,7 @@ const CURRENT_TAG_PREFIX = "memos-local-plugin-v";
 const TAG_PREFIXES = [CURRENT_TAG_PREFIX, "openclaw-local-plugin-v"];
 const RELEASE_NOTES_MARKER = "doc-agent-release-notes-json";
 const RELEASE_CATEGORY_ORDER = ["Added", "Improved", "Fixed"];
-const MAX_DRAFT_REPAIR_ATTEMPTS = 2;
+const MAX_DRAFT_REPAIR_ATTEMPTS = 3;
 const RELEASE_TO_DOC_CATEGORY = {
   Added: "New Features",
   Improved: "Improvements",
@@ -89,7 +89,7 @@ export function versionFromTag(tag) {
 
 export function parseSemver(version) {
   const cleaned = cleanVersion(version);
-  const match = cleaned.match(/^(\d+)\.(\d+)\.(\d+)(?:[-+]([0-9A-Za-z.-]+))?$/);
+  const match = cleaned.match(/^(\d+)\.(\d+)\.(\d+)(?:-([0-9A-Za-z.-]+))?(?:\+[0-9A-Za-z.-]+)?$/);
   if (!match) return null;
   return {
     major: Number(match[1]),
--- a/.github/scripts/draft-local-plugin-release-notes.mjs
+++ b/.github/scripts/draft-local-plugin-release-notes.mjs
@@ -72,6 +72,12 @@ export function displayVersion(raw) {
   return value ? `v${value}` : "";
 }
 
+export function isLegacyPackageOnlyRelease({ targetVersion, npmDistTag = "" } = {}) {
+  const parsed = parseSemver(targetVersion);
+  const distTag = String(npmDistTag || "").trim();
+  return Boolean(parsed?.prerelease) || Boolean(distTag && distTag !== "latest");
+}
+
 export function versionFromTag(tag) {
   for (const prefix of TAG_PREFIXES) {
     if (tag.startsWith(prefix)) {
@@ -173,10 +179,16 @@ function listProductTags() {
     .filter((item) => item.version && parseSemver(item.version));
 }
 
-export function findPreviousTag(targetVersion, currentTag) {
-  const candidates = listProductTags()
+export function selectPreviousTag(
+  tags,
+  targetVersion,
+  currentTag,
+  { includePrerelease = true } = {},
+) {
+  const candidates = tags
     .filter((item) => item.tag !== currentTag)
     .filter((item) => compareSemver(item.version, targetVersion) < 0)
+    .filter((item) => includePrerelease || !parseSemver(item.version)?.prerelease)
     .sort((a, b) => {
       const versionOrder = compareSemver(b.version, a.version);
       if (versionOrder !== 0) return versionOrder;
@@ -185,6 +197,10 @@ export function findPreviousTag(targetVersion, currentTag) {
       return bPreferred - aPreferred;
     });
   return candidates[0]?.tag || "";
+}
+
+export function findPreviousTag(targetVersion, currentTag, options = {}) {
+  return selectPreviousTag(listProductTags(), targetVersion, currentTag, options);
 }
 
 function parseCommits(previousTag, currentRef) {
@@ -426,6 +442,17 @@ function appendOutput(name, value) {
 export function ensureSourceHint(notes) {
   const hint = `<!-- doc-agent: source-id=${PRODUCT_ID} -->`;
   return notes.includes("doc-agent: source-id=") ? notes : `${notes.trim()}\n\n${hint}\n`;
+}
+
+export function validateLegacyPackageNotes(notes) {
+  const text = String(notes || "").trim();
+  if (!/^## Changelog\s*$/m.test(text)) {
+    fail("Legacy package release notes must contain a '## Changelog' heading.");
+  }
+  if (/doc-agent:\s*source-id=|doc-agent-release-notes-json/.test(text)) {
+    fail("Legacy package release notes must not include Doc Agent source hints or docs payloads.");
+  }
+  return `${text}\n`;
 }
 
 function normalizeReleaseCategory(value) {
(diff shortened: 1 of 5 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2198
**Task:** `d906a1a07079ff5f`
**Base:** `main`
**Head:** `fix/local-plugin-beta-legacy-publisher`

🔍 **OpenCodeReview** found **9** issue(s) in this PR.

---

### 1. `.github/scripts/draft-local-plugin-release-notes.mjs` (L1043)

When `NPM_DIST_TAG` is unset and the version-based check triggers (`1.0.0-beta.1` with no dist-tag env var), `legacyPackageDraftFromEvidence` is called with an empty `npmDistTag` and defaults to `"beta"` in the release notes text. This hardcoded fallback may be factually incorrect — the actual publish command may use a different dist-tag (e.g. `alpha`, `rc`, `next`). The two functions use different defaults: `isLegacyPackageOnlyRelease` treats an empty dist-tag as a non-trigger (only version-based), while `legacyPackageDraftFromEvidence` masks the absence with `"beta"`.

Consider passing the resolved dist-tag from outside both functions, or at minimum using `displayVersion(version)` in the bullet to reflect the actual prerelease identifier instead of the hardcoded `"beta"` fallback.

**Before:**
```
  const distTag = String(npmDistTag || "").trim() || "beta";
```

**After:**
```suggestion
  // Avoid hardcoding 'beta': if no dist-tag is known, omit the dist-tag phrase or
  // surface the actual prerelease identifier from the version string.
  const distTag = String(npmDistTag || "").trim();
  const distTagLabel = distTag || parseSemver(cleanVersion(version))?.prerelease || "prerelease";
```

</details>

---

### 2. `.github/scripts/draft-local-plugin-release-notes.mjs` (L452-L454)

The regex in `validateLegacyPackageNotes` hardcodes the marker strings `"doc-agent-release-notes-json"` and `"doc-agent: source-id="` that are already defined as the constant `RELEASE_NOTES_MARKER` (line 38) and used verbatim in `ensureSourceHint`. If either string is ever renamed or updated, this validation regex will silently stop catching them, allowing policy-violating notes to pass through.

Consider building the regex from the constant to stay in sync:
```js
const markerPattern = new RegExp(`${RELEASE_NOTES_MARKER}|doc-agent:\\s*source-id=`);
if (markerPattern.test(text)) { fail(...); }
```

**Before:**
```
  if (/doc-agent:\s*source-id=|doc-agent-release-notes-json/.test(text)) {
    fail("Legacy package release notes must not include Doc Agent source hints or docs payloads.");
  }
```

**After:**
```suggestion
  const legacyForbiddenPattern = new RegExp(`${RELEASE_NOTES_MARKER}|doc-agent:\\s*source-id=`);
  if (legacyForbiddenPattern.test(text)) {
    fail("Legacy package release notes must not include Doc Agent source hints or docs payloads.");
  }
```

</details>

---

### 3. `.github/scripts/draft-local-plugin-release-notes.mjs` (L1038-L1039)

In `legacyPackageDraftFromEvidence`, `version` is used raw (from `evidence?.target_version`) in the release note bullet, while `targetPackageVersion` applies `cleanVersion`. This means if `target_version` carries a `v` prefix (e.g. `"v2.0.13-beta.1"`), the display line will be:
```
Published MemOS Local Plugin v2.0.13-beta.1 as a package prerelease ...
- Package version: 2.0.12 -> 2.0.13-beta.1
```
The `v` prefix is inconsistently applied within the same notes block. Consider using `displayVersion(version)` for consistent formatting in the first bullet, and documenting that the version arrow line deliberately omits it.

**Before:**
```
  const version = evidence?.target_version || "";
  const targetPackageVersion = cleanVersion(version);
```

**After:**
```suggestion
  const version = evidence?.target_version || "";
  const displayedVersion = displayVersion(version); // consistent v-prefixed display
  const targetPackageVersion = cleanVersion(version);
```

</details>

---

### 4. `.github/scripts/prepare-memos-release.mjs` (L524-L528)

The pre-release branch unconditionally calls `incrementPatchVersion(previousReleasedVersion)`, which itself calls `fail()` if `previousReleasedVersion` happens to be a pre-release version (see `incrementPatchVersion` lines 163-166). While `previousReleasedVersion` is unlikely to be a pre-release in normal flow, the function's documented contract is that it won't auto-increment pre-release inputs. If a pre-release tag ever becomes the "previous released version", this path will throw an unexpected error. An explicit guard or comment explaining why `previousReleasedVersion` is guaranteed stable here would improve safety and maintainability.

---

### 5. `.github/scripts/prepare-memos-release.mjs` (L876)

The pattern terms `inject` and `rank` are very generic English words that can easily appear in unrelated commit messages (e.g., "inject dependency", "rank ordering", "inject mock"). This may cause false-positive matches, incorrectly classifying unrelated commits under the "Recall relevance and host input handling" topic. Consider narrowing the pattern with word-boundary anchors (`\b`) or more specific phrases to reduce misclassification risk.

**Before:**
```
    pattern: /recall|retrieval|host input|inject|rank|dedupe|keyword|relevance/i,
```

**After:**
```suggestion
    pattern: /\brecall\b|\bretrieval\b|host[- ]input|\bhost_input\b|\bdedupe\b|\bkeyword\b|\brelevance\b|\brank(?:ing)?\b(?=.*(?:recall|retrieval|relevance))/i,
```

</details>

---

### 6. `.github/scripts/prepare-memos-release.mjs` (L883-L891)

When the input matches the `revert` capture pattern (first `.replace()`), the resulting `$1` text is then passed through the second `.replace()` which tries to strip a Conventional Commit prefix again. If the reverted commit subject itself has a prefix like `feat(plugin): ...`, it will be stripped correctly. However, if the revert pattern fails to match (e.g. revert message uses single quotes or no quotes), the text falls through with the full `revert "..."` intact, and the second rule won't match it either since it checks for `revert` at the start but won't capture the inner subject. This means `revert` subjects that don't exactly match the double-quote convention will silently produce a raw revert string in the output. The `category` check on line 900 (`/^fix|^revert/i.test(source)`) still works correctly because it tests `source` (the raw input), not `cleaned` — but the displayed text may be ugly.

---

### 7. `.github/scripts/prepare-memos-release.mjs` (L898-L903)

After `fallbackSubjectText` strips the Conventional Commit prefix (e.g. `fix(plugin): `), `cleaned` may still contain a PR reference like `stabilize sandbox widget (#9999)` if the `\s+\(#\d+\)` trailing pattern doesn't match due to extra trailing spaces or punctuation variation. More critically, the `category` is determined from `source` (which still has the original prefix like `fix(plugin):`), while the text body uses `cleaned`. These two derived from the same `source` variable diverging is intentional and correct, but a brief comment clarifying this intent would help future maintainers avoid accidentally changing `source` to `cleaned` for the category check.

---

### 8. `.github/scripts/prepare-memos-release.test.mjs` (L806-L809)

The `allowGeneric: true` flag has no effect here because `"fix(plugin): improve recall relevance and host input handling (#2196)"` matches the static `FALLBACK_TOPIC_RULES` rule for `/recall|retrieval|host input|.../i`, so the generic code path (where prefix stripping actually happens) is never reached. The test name claims to verify "without raw commit prefixes", but it only confirms the static rule lookup — not stripping. This means a regression in the generic CC-prefix stripping logic would not be caught by this test.

Consider using a commit message that does **not** match any static rule to actually exercise the generic path, or add a parallel `allowGeneric: false` assertion to document that this test deliberately targets the rule-based path.

---

### 9. `.github/scripts/prepare-memos-release.test.mjs` (L821-L824)

The `text_en` field is not asserted to contain the cleaned text. Since the generic fallback constructs `text_en` symmetrically with `text_cn` (line 902 of production code: `` `**${PRODUCT_TITLE.en} update**: ${cleaned}` ``), the assertion `assert.match(topic.text_en, /fix\(plugin\)/)` only verifies the prefix is absent, but never confirms that the cleaned text `stabilize sandbox widget` is actually present in `text_en`. A future regression where `text_en` is set to an empty fallback string would silently pass this test.

Add `assert.match(topic.text_en, /stabilize sandbox widget/)` to mirror the `text_cn` assertion.

**Before:**
```
  assert.equal(topic.category, "Fixed");
  assert.match(topic.text_cn, /stabilize sandbox widget/);
  assert.doesNotMatch(topic.text_cn, /fix\(plugin\)/);
  assert.doesNotMatch(topic.text_en, /fix\(plugin\)/);
```

**After:**
```suggestion
  assert.equal(topic.category, "Fixed");
  assert.match(topic.text_cn, /stabilize sandbox widget/);
  assert.match(topic.text_en, /stabilize sandbox widget/);
  assert.doesNotMatch(topic.text_cn, /fix\(plugin\)/);
  assert.doesNotMatch(topic.text_en, /fix\(plugin\)/);
```

</details>

Memtensor-AI:

**Branch:** `fix/local-plugin-beta-legacy-publisher`

Memtensor-AI:

**Branch:** `fix/local-plugin-beta-legacy-publisher`

Memtensor-AI:

**Branch:** `fix/local-plugin-beta-legacy-publisher`

PR #2208, opened 2026-08-04
fix(plugin): prevent foreground starvation and honor Retry-After

### LINKED ISSUE #1620: 429 retry handling ignores Retry-After header in LLM and embedding fetchers

### LINKED ISSUE #2185: Hermes adapter: prefetch can exceed host 8s budget in default shared-bridge mode
## Summary

In `@memtensor/memos-local-plugin` `2.0.12`, the Hermes adapter uses a process-scoped shared stdio bridge by default:

```python
mode = os.environ.get("MEMOS_HERMES_BRIDGE_MODE", "shared")
```

Under normal multi-session gateway traffic, the Hermes host can report:

```text
Memory provider 'memtensor' prefetch timed out after 8.0s; skipping it until the stuck call returns
```

The bridge/viewer can remain healthy when this occurs, so `core.health` or viewer availability alone does not identify the slow stage.

## Environment

- `@memtensor/memos-local-plugin`: `2.0.12`
- Hermes integration via `adapters/hermes`
- Multiple provider/session instances in one gateway process
- Default shared bridge mode

## Expected behavior

A slow retrieval/capture request should be diagnosable, and automatic prefetch should not silently degrade without identifying the blocking phase.

## Observed behavior

- Hermes gives up at its outer 8-second prefetch budget.
- The MemOS bridge can still appear connected and the viewer remains available.
- The adapter has long-RPC timeouts, but it does not expose enough per-phase timing/status to distinguish:
  - `turn.start` retrieval,
  - embedding or model work,
  - host reverse RPC,
  - stdio transport,
  - or interaction between concurrent shared-bridge leases.

## Request

Could the Hermes adapter provide per-request timing/status for `turn.start` and shared-runtime in-flight/queue diagnostics?

A regression test would also help: simulate one slow request while another provider lease performs prefetch, then verify the second lease either completes within budget or receives a specific bounded/degraded result rather than an opaque host-side timeout.

## Temporary workaround

Use `MEMOS_HERMES_BRIDGE_MODE=legacy` / the pre-shared-bridge adapter path while investigating, at the cost of losing the shared-bridge process-leak mitigation.

### LINKED ISSUE #2186: fix: foreground turn.start RPC starved by background pipeline (embedding + LLM resource contention)
### Pre-submission checklist

### Bug Description

When the memos-local-plugin bridge is used as a memory provider for hermes-agent (via stdin/stdout JSON-RPC), the foreground `turn.start` RPC call frequently times out (8s) because it competes with background pipeline tasks (reflection scoring, reward backprop, skill crystallization) for two shared resources:

1. **Local ONNX embedding model** (single-threaded CPU inference) — background `captureRunner.runLite()` embeds trace rows while the foreground `retrieveTurnStart()` needs to embed the query for vector search. These serialize on the ONNX runtime.

2. **Remote LLM API** (mimo-v2.5 via openai_compatible) — background skill verification / reward scoring makes LLM calls that share the same API quota/concurrency as the foreground `openEpisodeIfNeeded()` topic classification.

Observed timing from bridge logs:
- `capture.reflect.scoring.start` → `capture.reflect.done`: 1501ms
- `reward.done`: 2564ms
- `skill.verify.fail` (resonance-low): **12 seconds** (LLM call)

During this 12-second skill verification window, any incoming `turn.start` RPC that needs embedding or LLM classification will exceed the host's 8-second prefetch timeout, causing the memory injection to be silently skipped.

**Expected behavior**: Foreground retrieval (`turn.start`) should not be starved by background pipeline work. The bridge should either:
- Prioritize foreground RPC over background tasks (e.g., yield/pause background LLM/embedding calls when a foreground request arrives)
- Run background pipeline on a separate worker/thread with its own embedding queue
- Expose a concurrency semaphore so embedding inference is fair-queued

### How to Reproduce

1. Configure hermes-agent with `memory.provider: memtensor` and `memos-plugin/config.yaml` using local embedding (`Xenova/all-MiniLM-L6-v2`) + remote LLM (`mimo-v2.5`)
2. Have an active conversation with multiple turns (so episodes accumulate)
3. Trigger an episode boundary (topic change or session end) — this fires the full reflect → reward → skill pipeline
4. Immediately send a new message (within the ~12s skill verification window)
5. Observe: `WARNING agent.memory_manager: Memory provider 'memtensor' prefetch timed out after 8.0s`

### Environment

- Node.js: v20.x (bridge runtime)
- Python: 3.12.3 (hermes-agent host)
- Operating System: Ubuntu 24.04 (WSL2)
- MemOS version: memos-local-plugin 2.0.7 (bridge.cjs)
- Embedding: local `Xenova/all-MiniLM-L6-v2` (ONNX, 384-dim)
- LLM: `mimo-v2.5` via openai_compatible endpoint

### Additional Context

Relevant source locations (compiled dist):
- `dist/core/pipeline/orchestrator.js` L976: `onTurnStart` — foreground retrieval path
- `dist/core/pipeline/orchestrator.js` L1216: `captureRunner.runLite()` — background capture in `onTurnEnd`
- `dist/core/embedding/embedder.js` L69: `embedMany()` — shared embedding facade (no priority queue)
- `dist/bridge/stdio.js`: JSON-RPC dispatch (concurrent, but downstream resources are serialized)

The stdio transport itself handles requests concurrently (no serialization at the RPC layer). The bottleneck is purely in the shared embedding/LLM resource layer below.

### Willingness to Implement

- [x] I'm willing to implement this myself
- [ ] I would like someone else to implement this

Hun-ger:
## Description

This PR hardens the foreground retrieval path of `memos-local-plugin` for both Hermes and OpenClaw, with a focus on bounded latency, resource fairness, retry correctness, recovery, and shutdown safety.

### Problems solved

- Hermes `turn.start` could exceed the host's 8-second prefetch timeout because the adapter and core pipeline did not share one end-to-end deadline.
- Foreground retrieval could be starved by background embedding and LLM enrichment work.
- A slow Hermes reverse RPC handler could block the bridge stdout reader, delaying unrelated responses across shared provider leases.
- LLM and embedding fetchers retried HTTP 429/503 responses without correctly honoring long `Retry-After` cooldowns.
- Retry timers, queued resource waiters, and in-flight provider calls could outlive pipeline shutdown.
- `memory_add` log details could present user content as assistant content or omit one side of a user/assistant turn.

### Implementation approach

- Introduce a dedicated, configurable Hermes foreground prefetch budget:
  - 6 seconds by default.
  - Configurable through `MEMOS_HERMES_PREFETCH_RPC_TIMEOUT`.
  - Capped below Hermes' outer 8-second timeout.
- Propagate a single absolute `deadlineAt` from the Hermes adapter through:
  - bridge RPC;
  - retrieval and classification;
  - embedding and LLM provider calls;
  - retry decisions and degraded fallback handling.
- Add priority-aware foreground resource coordination:
  - foreground retrieval is admitted ahead of background enrichment;
  - embedding access is fair-queued with bounded foreground bursts;
  - background embedding batches are chunked to avoid monopolizing the provider;
  - request cancellation and pipeline shutdown abort queued and in-flight work.
- Preserve the existing raw → enrichment → retry/recovery → shutdown lifecycle:
  - raw capture is given a bounded shutdown grace period;
  - enrichment and retries stop accepting new work during shutdown;
  - shutdown does not wait indefinitely on provider retries.
- Move Hermes reverse RPC handlers to a bounded daemon worker queue:
  - stdout response demultiplexing remains responsive;
  - handler execution remains serial to preserve the existing concurrency contract;
  - queue saturation returns an explicit `host_handler_busy` error.
- Implement RFC-compatible `Retry-After` handling:
  - support delay-seconds and HTTP-date formats;
  - never shorten an upstream-requested cooldown;
  - defer retries that cannot fit the foreground deadline;
  - retain provider cooldown state so subsequent requests do not retry prematurely;
  - expose bounded retry diagnostics such as `retryAt`, decision, and reason.
- Correct `memory_add` log presentation so user and assistant content are displayed with their proper roles while retaining compatibility with existing stored traces.

No new runtime dependencies or configuration migrations are required. Existing callers that do not provide `deadlineAt` continue to use the previous call contract.

Related Issues:

Fixes #1620  
Fixes #2185  
Fixes #2186

## Type of change

## How Has This Been Tested?

Automated tests cover:

- foreground/background embedding admission and starvation prevention;
- bounded foreground deadlines across bridge, retrieval, LLM, and embedding calls;
- request cancellation and shutdown while work is queued or retrying;
- `Retry-After` delay-seconds and HTTP-date parsing;
- long cooldown deferral without premature retries;
- retry deadline exhaustion and diagnostic fields;
- Hermes bridge reader head-of-line blocking;
- concurrent normal RPC responses while a reverse host handler is slow;
- bounded reverse-handler queue saturation and shutdown;
- user/assistant role presentation in `memory_add` logs;
- backward-compatible behavior when no deadline is supplied.

Verification commands:

```bash
cd apps/memos-local-plugin

npm run lint
npm test
python3 -m pytest tests/python -q
npm run build:package


Memtensor-AI:

**Details:** Executor error: Command failed: git clone --depth 1 --branch fix/plugin-foreground-deadline-retry git@github.com:MemTensor/MemOS.git /data/test-workspaces/6e7d90a51a853357/repo
Cloning into '/data/test-workspaces/6e7d90a51a853357/repo'...
warning: Could not find remote branch fix/plugin-foreground-deadline-retry to clone.
fatal: Remote branch fix/plugin-foreground-deadline-retry not found in upstream origin
**Branch:** `fix/plugin-foreground-deadline-retry`

PR #2227, opened 2026-08-06
fix(api): authenticate /product, bind user_id to principal, clamp doc_path

### LINKED ISSUE #2202: /product REST surface is unauthenticated by default; user_id is client-supplied (cross-user IDOR) and doc_path reads arbitrary server files
## What

On the self-hosted MemOS server (`src/memos/api/server_api.py`), the `/product` router (`src/memos/api/routers/server_router.py:68`) exposes memory read/write, cube, scheduler, and chat endpoints with **no authentication dependency**, while auth is **off by default**.

- Every `/product` handler is a plain `def f(req)` with no `Depends(verify_api_key)` — e.g. `search_memories` (`server_router.py:112`), `add_memories` (`:128`), `create_cube` (`:143`), `chat_complete` (`:271`), `scheduler_status` (`:208`). Compare the admin router, which gates every route with `auth: dict = Depends(verify_api_key)` (`admin_router.py:81,128,158,190`). The `/product` router is included with no `dependencies=[...]` either (`server_api.py:50`).
- `AUTH_ENABLED` defaults to `false` (`src/memos/api/middleware/auth.py:26`), and even when enabled, `verify_api_key` returns an all-scopes context when `AUTH_ENABLED` is false (`auth.py:173`), so the admin routes are open by default too.
- `user_id` is taken from the request (e.g. `scheduler_status(user_id: str = Query(...))` at `server_router.py:209`), never authenticated — a caller picks any user id.
- `doc_path` on add/search flows to `_get_all_documents` (`src/memos/mem_os/core.py:233`), which does `Path(path).rglob("*")` over the caller-supplied path and returns every `.txt/.pdf/.json/.md/.ppt(x)` file under it for the server to read and process.

Net effect: by default, an unauthenticated caller can read/write any user's memories (pick `user_id`), enumerate/read arbitrary server-side documents (`doc_path=/etc`, a home dir, etc.), and drive the chat/scheduler surface.

There is a related secondary bypass: `is_internal_request` (`auth.py:147`) returns `True` when the `X-Internal-Service` header equals `os.getenv("INTERNAL_SERVICE_SECRET")`; if that env var is unset (default) and the header is absent, the comparison is `None == None` → the request is treated as internal/full-scope even when `AUTH_ENABLED=true`.

## How to reproduce

Default config (no `AUTH_ENABLED`, no `INTERNAL_SERVICE_SECRET`), server running on `127.0.0.1:5050` (or wherever it binds):

```bash
# 1. unauthenticated access — no API key, no header
curl -s http://127.0.0.1:5050/product/scheduler/allstatus | head

# 2. cross-user IDOR — pick another user's id
curl -s -X POST http://127.0.0.1:5050/product/search \
  -H 'Content-Type: application/json' \
  -d '{"user_id":"victim","query":"*","cube_id":"default"}' | head

# 3. arbitrary server-side file read via doc_path (rglob under any path)
curl -s -X POST http://127.0.0.1:5050/product/add \
  -H 'Content-Type: application/json' \
  -d '{"user_id":"attacker","cube_id":"default","doc_path":"/etc"}' | head
```

All three return `200` with data and no `401/403`. The handler signatures in `server_router.py` confirm there is no auth dependency to trigger a `401`.

Checked at `93e4082` (`main` HEAD).

## Impact / scope

- This is about the **default configuration of the self-hosted REST server**. If the server is reachable beyond a single trusted loopback (container with a published port, a shared host, a reverse proxy), the above is unauthenticated and remote.
- Even on loopback, the `user_id`-from-request design is a cross-user IDOR (one local principal reads/writes another's memories), and `doc_path` is an arbitrary server-side file read limited only by the process's filesystem permissions and the extension allowlist.
- A `SECURITY.md` / documented trust model would help set expectations; right now there is none in the core service, so I'm reporting the default-unauth + IDOR + file-read posture as-is.

## Suggested change

- Put a `Depends(verify_api_key)` (or an equivalent) on the `/product` router, or gate it at `include_router(..., dependencies=[Depends(verify_api_key)])`.
- Default `AUTH_ENABLED` to `true` (or require an explicit opt-in to the unauthenticated mode), and document the deployment boundary.
- Derive `user_id` from the authenticated principal instead of a client-supplied query/body field.
- Clamp `doc_path` to an allowlisted documents root rather than `rglob`-ing an arbitrary caller path; and fix the `None == None` internal-service check (require the secret to be set, and compare securely).

Happy to open a PR for the `Depends(verify_api_key)` + `user_id` parts.

Evenss:
## Description

Fixes the four issues reported in #2202, staying within the scope of the issue's *Suggested change*.

1. **`/product` is unauthenticated** — the router now declares `dependencies=[Depends(verify_api_key)]`, so every current and future endpoint under it requires a key.
2. **Auth is off by default** — `AUTH_ENABLED` now defaults to `true`. The two health endpoints parse the same variable independently, so their defaults were flipped too; a startup warning is logged when auth is on but `MASTER_KEY_HASH` is unset.
3. **`user_id` is client-supplied (IDOR)** — new `resolve_authorized_user_id()` binds a non-privileged principal to its own user ID (an API key's `user_name` is the Product API user ID it may act as). Applied to the 17 `/product` endpoints that accept a caller-supplied user ID; the remaining endpoints have no user ID to bind, so they are authenticated but not further restricted. Master key / internal-service / admin principals keep acting for any user. With auth disabled, the identity is now pinned to `MOS_USER_ID` instead of the client-supplied `X-User-Name`.
4. **`doc_path` reads arbitrary files, and `None == None` internal bypass** — `_get_all_documents` confines `doc_path` to `MEMOS_DOC_ROOT` (falling back to `FILE_LOCAL_PATH`), re-resolving each matched file so symlinks cannot escape; local library usage with neither variable set is unchanged. `is_internal_request` no longer trusts the source IP and requires both the secret and the header to be non-empty, compared with `hmac.compare_digest`.

`api_analyzer.py` is a passive adaptation — it calls the endpoint functions in-process and must now pass an explicit internal `auth`.

Related Issue (Required): Fixes #2202

## Breaking changes

- Deployments that do not set `AUTH_ENABLED` will start rejecting unauthenticated requests. `MASTER_KEY_HASH` was added to both `docker/.env.example` files; note `server_api.py` does not mount `admin_router`, so the master key is the only bootstrap path on the base server.
- API keys with a non-null `expires_at` will now always fail to authenticate. `lookup_api_key` compares a `datetime` against `time.time()`, and the `TypeError` is swallowed into a `return None`. It is pre-existing and fails closed, so it is left for a separate fix, but default-on auth makes it reachable.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

- [x] Unit Test

9 regression cases, one per issue item. The fixtures in `test_server_router.py` and `test_cube_endpoints.py` now supply an admin key, since their pre-existing cases would otherwise all return 401.

```
$ poetry run pytest tests/api/test_auth.py tests/api/test_server_router.py tests/api/test_cube_endpoints.py -q
37 passed in 9.71s

$ poetry run pytest tests/mem_os/test_memos_core.py -q -k DocumentPathConfinement
3 passed, 24 deselected in 8.18s

$ make format
All checks passed!
```

`test_mos_init_invalid_user` and the four `test_mcp_serve.py` import errors fail identically at the base commit and are unrelated.

## Checklist

- [ ] I have created related documentation issue/PR in [MemOS-Docs](https://github.com/MemTensor/MemOS-Docs) (if applicable) | 我已在 MemOS-Docs 中创建了相关的文档 issue/PR（如果适用）
- [x] I have linked the issue to this PR (if applicable) | 我已将 issue 链接到此 PR


Memtensor-AI:

**Failed tests:**
- `test_mos_init_invalid_user`

<details>

```
Tests failed. Failed cases: test_mos_init_invalid_user
```
</details>

**Branch:** `fix-product-auth`

PR #2230, opened 2026-08-06
Fix/plugin windows runtime lifecycle

### LINKED ISSUE #2212: Bug: viewer config save overwrites viewer.port with UI default (18799) on Hermes
## Bug Description

On the Hermes variant, saving settings from the Memory Viewer overwrites `viewer.port` in `~/.hermes/memos-plugin/config.yaml` with the UI default `18799` (the OpenClaw port). Hermes hardcodes its viewer port to `18800` (the bridge connects there), so after a save the daemon listens on a different port and the bridge/stdio link breaks until the config is fixed by hand.

Observed side effects of one save: `viewer.port` changed `18800 -> 18799`, and a placeholder `embedding.endpoint: "mem os"` appeared.

## Steps to Reproduce

1. Install memos-local-plugin into Hermes (v2.0.12).
2. Open the Memory Viewer at :18800 → Settings.
3. Change any setting and save (or click restart after save).

## Expected Behavior

Saving settings must not clobber fields the user did not edit — especially `viewer.port`, which the Hermes adapter owns.

## Actual Behavior

`PATCH /api/v1/config` accepts the whole dirty form (`viewer/src/views/SettingsView.tsx:139`) and the backend writes it verbatim, so untouched fields fall back to UI defaults (e.g. `18799`).

## Suggested Fix

- Backend: protect/ignore adapter-owned keys (`viewer.port`, agent identity) on `PATCH /api/v1/config`, or merge with the existing config instead of replacing.
- Frontend: do not submit unchanged fields.

## Environment

- OS: Windows
- Plugin: memos-local-plugin 2.0.12 (Hermes adapter)

### LINKED ISSUE #2213: Bug: admin/restart is self-destructive on Windows (pkill/bash/tsx unavailable, daemon always killed)
## Bug Description

`POST /api/v1/admin/restart` is self-destructive on Windows: the daemon always dies and is never replaced, leaving the Memory Viewer unreachable.

Flow in `server/routes/admin.ts`:

1. `terminateHermesChat()` runs `pkill -TERM -f "/bin/hermes"` / `"hermes chat"` — `pkill` does not exist on Windows, so this returns false and does nothing (silently).
2. `spawnReplacementDaemon()` runs `bash -c "sleep 3 && <node> <tsx> bridge.cts --daemon"` — `bash`/`tsx` are not reliably available on Windows (and `tsx` lives in devDependencies, absent from the published package).
3. `scheduleHermesShutdown(200)` then sends `SIGTERM` to its own PID after 200 ms — this always succeeds.

Net result: the original daemon is killed 200 ms after the restart request and the replacement never comes up, so the viewer port goes silent. This also happens if the user only saves settings, because `SettingsView.tsx` triggers the restart overlay after save.

## Steps to Reproduce

1. Windows host, memos-local-plugin running under Hermes.
2. Open viewer → Settings → save any change (or click the restart prompt).
3. Viewer becomes unreachable; the daemon process exits and nothing replaces it.

## Expected Behavior

On Windows, restart should either (a) work — spawn a replacement via `cmd.exe`/PowerShell or a Windows-friendly mechanism — or (b) degrade gracefully: tell the user to restart Hermes instead of killing the daemon.

## Suggested Fix

- Detect `process.platform === "win32"` and skip the pkill/bash replacement path; either restart in-process or instruct the user to restart the host agent.
- Only self-terminate after a replacement is confirmed listening (or skip the 200 ms suicide entirely).

## Related

- #1586 align Hermes restart UX
- #1564 preserve stdio connection on restart

## Environment

- OS: Windows
- Plugin: memos-local-plugin 2.0.12 (Hermes adapter)

### LINKED ISSUE #2218: fix: bug: Hermes viewer daemon always reports "occupied" on Windows (connection-refused detection misses WinError 10061)

### Bug Description | 问题描述

On Windows, the Hermes viewer daemon's port-availability probe
(_probe_json_url in daemon_manager.py) never recognizes an unused port as
"free". It only checks for macOS/Linux ECONNREFUSED errno values (61, 111)
and English-language "connection refused" text, but Windows raises errno
10061 with a message that varies by system locale (e.g. Czech: "cílový
počítač je aktivně odmítl"). As a result, the viewer web panel is
permanently reported as blocked by a "non-MemOS service" even when nothing
is listening on the port, and http://127.0.0.1:18800/ never becomes usable
on a fresh Windows install.

### How to Reproduce | 如何重现

1. Fresh Windows 11 install with Hermes Agent, non-English system locale (reproduced on Czech locale; likely affects any non-English Windows)
2. Install memos-local-plugin via install.ps1
3. Start Hermes and check agent.log — observe:
   WARNING daemon_manager: MemOS: viewer port 18800 is occupied by a non-MemOS service; memory capture will continue without the web panel
4. Confirm nothing is actually listening: Get-NetTCPConnection -LocalPort 18800 returns empty at the time of the warning
5. Memory Viewer at http://127.0.0.1:18800/ never becomes reachable

### Environment | 环境信息

- Python version: 3.11.15
- Operating System: Windows 11 (build 26100), Czech system locale
- MemOS version: (run `pip show memoryos` inside the plugin's venv, or check package.json version in memos-local-plugin folder)

### Additional Context | 其他信息

Root cause: `_probe_json_url()` in daemon_manager.py only recognizes macOS/Linux ECONNREFUSED errno values (61, 111) and English "connection refused" text as "port is free". Windows raises errno 10061 (WSAECONNREFUSED) with a locale-dependent message (e.g. Czech: "cílový počítač je aktivně odmítl"), which matches neither check, so an unused port is always misclassified as "blocked".

Suggested fix in `_probe_json_url()`:

    except urllib.error.URLError as err:
        reason = getattr(err, "reason", None)
        errno = getattr(reason, "errno", None)
        # 61 = macOS ECONNREFUSED, 111 = Linux ECONNREFUSED, 10061 = Windows WSAECONNREFUSED
        if errno in {61, 111, 10061}:
            return "free"
        if isinstance(reason, ConnectionRefusedError):
            return "free"
        msg = str(err).lower()
        if "connection refused" in msg or "failed to establish" in msg:
            return "free"
        return "blocked"

Using `isinstance(reason, ConnectionRefusedError)` is the most robust check since Python raises this exception type consistently across platforms, regardless of OS errno or locale-specific message text.

Impact: memory capture (stdio bridge) still works, but the web viewer panel is silently and permanently disabled on Windows installs, with no clear indication that this is a platform-detection bug rather than an actual port conflict.

- [x] I'm willing to implement this myself | 我愿意自己解决
- [x] I would like someone else to implement this | 我希望其他人来解决

### LINKED ISSUE #2221: Windows: plugin runtime data and paths assume POSIX ~/.hermes instead of HERMES_HOME (%LOCALAPPDATA%\hermes)
## Bug Description

On Windows, the memos-local-plugin resolves the Hermes home as `~/.hermes` in several places, while Hermes itself uses `%LOCALAPPDATA%\hermes` (`HERMES_HOME`). The result is that the plugin's runtime data lives **outside** the Hermes home, and several features either break or silently miss data.

Hermes' own platform resolution (`hermes_constants.py`):

```python
def _get_platform_default_hermes_home() -> Path:
    if sys.platform == "win32":
        local_appdata = os.environ.get("LOCALAPPDATA", "").strip()
        base = Path(local_appdata) if local_appdata else Path.home() / "AppData" / "Local"
        return base / "hermes"
    return Path.home() / ".hermes"
```

The plugin hardcodes the POSIX convention instead:

- `adapters/hermes/memos_provider/__init__.py:132` → `Path.home() / ".hermes" / "memos-plugin"`
- `adapters/hermes/memos_provider/bridge_client.py:69` → `agent_home = ".hermes"`
- `bridge.cts:96` / `bridge.mts:82` → `agentHome = ".hermes"`
- `core/config/paths.ts:32` → `hermes: "{HOME}/.hermes/memos-plugin"`
- `server/routes/migrate.ts:70` → `join(homedir(), ".hermes", "memos-state", ...)`

## Observed impact on Windows

1. **Install writes config to the wrong place** — the template `config.yaml` lands under the code dir instead of the runtime home the daemon reads → #2211 (fixed by PR).
2. **Native memory import can't find Hermes' `MEMORY.md`** — hardcoded `~/.hermes`, but Hermes stores it under `%LOCALAPPDATA%\hermes\memories\` → #2210 (fixed by PR).
3. **`hermes backup` may not include plugin data** — Hermes' backup walks `HERMES_HOME`; state kept at `~/.hermes/memos-plugin` is outside it, so it is only captured if the provider declares `backup_paths()` (see `agent/memory_provider.py`). On Windows this is the default layout, so a fresh install's data is at risk of being omitted from backups.
4. **Data is split from the Hermes home** — tools, scripts and users that follow `HERMES_HOME` won't find the plugin's data; the plugin creates its own `~/.hermes` directory that is unrelated to the real Hermes home on Windows.

## Question

Is this intentional (a cross-platform convention where the plugin keeps its data at `Path.home()/.hermes` regardless of the host's real home), or an oversight from POSIX-first development? If intentional, it would help to document it; if not, resolving the Hermes home through `HERMES_HOME` (with a platform fallback, like the fix in #2210) would make the plugin consistent with Hermes on every platform.

## Environment

- OS: Windows (native, not WSL)
- Plugin: memos-local-plugin 2.0.12 (Hermes adapter)

## Related

- #2210 (native memory import path)
- #2211 (install.ps1 runtime home)

Hun-ger:
## Description

This PR fixes several Windows-specific configuration, runtime lifecycle, and installation issues in the MemOS local plugin while preserving existing behavior on macOS and Linux.

### Problems solved

- Windows runtime data paths could diverge because some components assumed the legacy `~/.hermes` directory while newer Hermes installations use `%LOCALAPPDATA%\hermes` (#2221).
- The Viewer port probe did not correctly recognize Windows connection-refused error `10061`, causing an available port to be reported as occupied and preventing the Viewer from starting reliably (#2218).
- Restarting the plugin on Windows could terminate the current process without a supervisor capable of starting it again (#2213).
- Saving Hermes configuration could incorrectly persist OpenClaw's port `18799`, causing Hermes Viewer configuration and its effective runtime port to diverge (#2212).

### Implementation approach

- Added platform-aware runtime-home resolution for Windows:
  - Explicitly configured home paths take precedence.
  - Previously selected paths are persisted through a marker.
  - Existing legacy data is preferred to preserve compatibility with previous versions.
  - The resolved home is explicitly propagated between the provider, bridge, and Viewer processes.
  - Existing SQLite data is not automatically moved.

- Restored and enforced the adapter-specific runtime ports:
  - Hermes Viewer: `18800`
  - OpenClaw Viewer: `18799`
  - Missing Hermes configuration is created from the Hermes template with port `18800`.
  - Legacy Hermes configurations incorrectly written with port `18799` are migrated once, with a backup created before modification.
  - Runtime configuration responses expose the effective adapter port.

- Improved Viewer port detection:
  - Supports connection-refused codes `61`, `111`, and Windows `10061`.
  - Uses conservative port states to distinguish available, occupied, unreachable, and unknown conditions.
  - Keeps localhost traffic on a direct connection path.

- Added a Windows-safe restart flow:
  - Prevents the plugin from terminating itself when no supervisor is available to relaunch it.
  - Provides agent-specific manual restart instructions for Hermes and OpenClaw.
  - Keeps the existing supervised restart behavior on macOS and Linux unchanged.
  - Improves Viewer state reporting so an offline Viewer is distinguishable from a model request failure.

- Hardened the Windows installer:
  - Uses staged installation and rollback to avoid leaving a partially installed plugin.
  - Treats npm installation and native rebuild failures as fatal.
  - Verifies the native `better-sqlite3` module after installation.
  - Preserves existing OpenClaw hooks and unrelated plugin configuration.
  - Avoids writing unsupported `plugins.installs` entries while remaining compatible with older OpenClaw configurations.

- Updated the local plugin version to `2.0.14-beta.1`.

No new runtime dependencies are introduced by this change.

Related Issues (Required):

- Fixes #2221
- Fixes #2218
- Fixes #2213
- Fixes #2212

## Type of change

## How Has This Been Tested?

- [x] Unit Test
- [x] Test Script Or Test Steps (please provide)

### Automated verification

```bash
npm run test:unit
npm run lint
npm run build:package
git diff --check
```

Results:

- 158 unit test files passed
- 1,337 tests passed
- 1 test skipped
- TypeScript type checking passed
- Local npm package build passed
- Git whitespace validation passed

### Windows installation and runtime verification

The generated `2.0.14` npm package was installed on a Windows machine for both Hermes and OpenClaw.

Verified scenarios:

1. Install and upgrade the local package using the updated PowerShell installer.
2. Confirm existing Hermes/OpenClaw configuration and OpenClaw hooks are preserved.
3. Save configuration and restart Hermes.
4. Confirm Hermes Viewer becomes available at `127.0.0.1:18800`.
5. Restart OpenClaw using the displayed OpenClaw-specific instructions.
6. Confirm OpenClaw Viewer becomes available at `127.0.0.1:18799`.
7. Validate OpenClaw configuration:

   ```powershell
   openclaw config validate --json
   openclaw gateway status --deep --require-rpc
   ```

8. Run a real OpenClaw agent turn and verify:
   - The agent produces a response.
   - `memos_search` succeeds.
   - The `agent_end` hook triggers `memory_add`.
   - The memory is stored successfully.
9. Verify Windows clear-data behavior and confirm data is removed without terminating the Viewer before the response is delivered.
10. Verify legacy runtime data remains discoverable and no automatic SQLite migration occurs.

## Checklist

- [ ] I have created related documentation issue/PR in [[MemOS-Docs](https://github.com/MemTensor/MemOS-Docs)](https://github.com/MemTensor/[MemOS-Docs](https://github.com/MemTensor/MemOS-Docs)) (if applicable) | 我已在 [MemOS-Docs](https://github.com/MemTensor/MemOS-Docs) 中创建了相关的文档 issue/PR（如果适用）

## Reviewer Checklist

- [ ] Closes #2221, #2218, #2213, and #2212
- [ ] Made sure Checks passed


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2230
**Task:** `e245d1faac77ccd5`
**Base:** `main`
**Head:** `fix/plugin-windows-runtime-lifecycle`

🔍 **OpenCodeReview** found **47** issue(s) in this PR.

---

### 1. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L2255-L2259)

`runtime_home` is computed twice with identical expressions (lines 2255 and 2259). The second assignment immediately overwrites the first before any mutation occurs, making the first computation dead. If `_resolved_memos_runtime_home()` has any side-effect or its return value could differ between calls (e.g., due to a file-system race or env change), the two values could silently diverge and `_prepare_shared_bridge` would operate on a different path than `MemosBridgeClient`. Compute it once and reuse.

**Before:**
```
            runtime_home = self._runtime_home or _resolved_memos_runtime_home()
            _prepare_shared_bridge(runtime_home, cleanup_legacy_zombies=True)
            new_bridge: MemosBridgeClient | None = None
            try:
                runtime_home = self._runtime_home or _resolved_memos_runtime_home()
```

**After:**
```suggestion
            runtime_home = self._runtime_home or _resolved_memos_runtime_home()
            _prepare_shared_bridge(runtime_home, cleanup_legacy_zombies=True)
            new_bridge: MemosBridgeClient | None = None
            try:
                runtime_env = dict(self._runtime_env or _memos_runtime_env_snapshot(runtime_home))
```

</details>

---

### 2. `apps/memos-local-plugin/adapters/hermes/memos_provider/__init__.py` (L1736-L1740)

The user-supplied `values["viewer_port"]` is silently discarded and replaced with the hard-coded literal `18800`. Any value the user or a test passes in is completely ignored. This makes the `viewer_port` parameter in `get_config_specs` (default `18800`, `required=False`) a no-op, which is surprising and will confuse callers who expect their value to be persisted. If the port must be fixed, the parameter should either be removed from `get_config_specs` or the method should validate/reject non-18800 values explicitly rather than silently overwriting them.

**Before:**
```
        if "viewer_port" in values:
            # Keep the legacy setup field for host compatibility, but the
            # Hermes adapter owns :18800. Persist the effective value so the
            # YAML file never advertises a port the runtime will not bind.
            payload["viewer"] = {"port": 18800}
```

**After:**
```suggestion
        if "viewer_port" in values:
            # The Hermes adapter is pinned to :18800; reject any other value
            # rather than silently discarding it.
            port = int(values["viewer_port"])
            if port != 18800:
                raise ValueError(f"viewer_port must be 18800, got {port}")
            payload["viewer"] = {"port": 18800}
```

</details>

---

### 3. `apps/memos-local-plugin/adapters/hermes/memos_provider/daemon_manager.py` (L403-L405)

_viewer_start_lock already accepts runtime_home=None as its default and handles None internally via `(runtime_home or _plugin_root())`. The ternary is redundant — passing runtime_home unconditionally is identical in all cases and eliminates a branch that could silently diverge if the signature changes.

**Before:**
```
        lock_context = (
            _viewer_start_lock(runtime_home) if runtime_home is not None else _viewer_start_lock()
        )
```

**After:**
```suggestion
        lock_context = _viewer_start_lock(runtime_home)
```

</details>

---

### 4. `apps/memos-local-plugin/adapters/hermes/memos_provider/daemon_manager.py` (L334-L346)

The trailing `except Exception` catch-all swallows any remaining unexpected exception (e.g., from `_LOOPBACK_OPENER.open`, `resp.read`, or header access) and silently returns `"unknown"`. This makes it impossible to distinguish a real programming error from a benign transport anomaly. Since every recoverable transport error is already handled by the `urllib.error.HTTPError`, `urllib.error.URLError`, and `TimeoutError` branches above, the bare `except Exception` is broader than necessary. Consider removing it entirely, or at minimum logging the unexpected error before returning.

**Before:**
```
    except urllib.error.URLError as err:
        reason = getattr(err, "reason", None)
        error_codes = {
            getattr(reason, "errno", None),
            getattr(reason, "winerror", None),
        }
        if isinstance(reason, ConnectionRefusedError) or error_codes & {61, 111, 10061}:
            return "free"
        return "unknown"
    except TimeoutError:
        return "unknown"
    except Exception:
        return "unknown"
```

**After:**
```suggestion
    except urllib.error.URLError as err:
        reason = getattr(err, "reason", None)
        error_codes = {
            getattr(reason, "errno", None),
            getattr(reason, "winerror", None),
        }
        if isinstance(reason, ConnectionRefusedError) or error_codes & {61, 111, 10061}:
            return "free"
        return "unknown"
    except TimeoutError:
        return "unknown"
    except Exception:
        logger.debug("MemOS: unexpected error probing %s", url, exc_info=True)
        return "unknown"
```

</details>

---

### 5. `apps/memos-local-plugin/adapters/hermes/memos_provider/daemon_manager.py` (L315-L320)

The `OSError` handler only recognises three numeric codes for address-in-use (48=macOS EADDRINUSE, 98=Linux EADDRINUSE, 10048=Windows WSAEADDRINUSE). A more robust alternative is to check `isinstance(err, OSError) and err.errno == errno.EADDRINUSE` using the stdlib `errno` module, which is portable across all platforms and future-proof against new OS mappings. Additionally, `error_codes` always contains `None` when `winerror` is absent on non-Windows, which is harmless but needlessly pollutes the set.

**Before:**
```
    except OSError as err:
        error_codes = {getattr(err, "errno", None), getattr(err, "winerror", None)}
        if error_codes & {48, 98, 10048}:
            return "occupied"
        return "unknown"
    return "free"
```

**After:**
```suggestion
    except OSError as err:
        import errno as errno_mod
        if err.errno == errno_mod.EADDRINUSE:
            return "occupied"
        return "unknown"
    return "free"
```

</details>

---

### 6. `apps/memos-local-plugin/adapters/hermes/memos_provider/daemon_manager.py` (L440-L454)

`log_handle` is opened without a `with` statement and closed manually at line 454. If `subprocess.Popen` raises (e.g., `FileNotFoundError` for the node binary), the `except` block on line 455 attempts `log_handle.close()` with a `# type: ignore[possibly-undefined]` guard — meaning the type checker itself flagged that `log_handle` may be undefined at that point (if `log_file.open()` itself raises). The safer pattern is to open the file inside a `with` block so the OS handle is always released regardless of where the exception originates.

**Before:**
```
                log_handle = log_file.open("a", encoding="utf-8")
                _viewer_process = subprocess.Popen(
                    _bridge_command(daemon=True, runtime_home=runtime_home),
                    cwd=str(plugin_root),
                    stdout=log_handle,
                    stderr=subprocess.STDOUT,
                    stdin=subprocess.DEVNULL,
                    text=True,
                    start_new_session=True,
                    env={
                        **os.environ,
                        **({"MEMOS_HOME": str(runtime_home.resolve())} if runtime_home else {}),
                    },
                )
                log_handle.close()
```

**After:**
```suggestion
                with log_file.open("a", encoding="utf-8") as log_handle:
                    _viewer_process = subprocess.Popen(
                        _bridge_command(daemon=True, runtime_home=runtime_home),
                        cwd=str(plugin_root),
                        stdout=log_handle,
                        stderr=subprocess.STDOUT,
                        stdin=subprocess.DEVNULL,
                        text=True,
                        start_new_session=True,
                        env={
                            **os.environ,
                            **({
                                "MEMOS_HOME": str(runtime_home.resolve())
                            } if runtime_home else {}),
                        },
                    )
```

</details>

---

### 7. `apps/memos-local-plugin/adapters/hermes/memos_provider/runtime_home.py` (L100-L101)

This broad catch silently discards all `OSError` variants, including permission errors (`EACCES`) that are unrelated to a missing or malformed marker. A permission error here will then likely cause `_write_marker` to also fail (since `marker.parent.mkdir` and `temp.write_text` will hit the same permission barrier), yet the caller receives no diagnostic. At minimum, distinguish `FileNotFoundError` (expected) from other `OSError` subtypes and log unexpected ones at DEBUG level so operators can diagnose installation issues.

Suggested approach:
```python
try:
    payload = json.loads(marker.read_text(encoding="utf-8"))
except FileNotFoundError:
    return None
except (OSError, ValueError, TypeError) as err:
    logger.debug("runtime-home marker unreadable (%s): %s", marker, err)
    return None
```

---

### 8. `apps/memos-local-plugin/adapters/hermes/memos_provider/runtime_home.py` (L117)

Using only `os.getpid()` as the uniqueness suffix means that if two threads in the same process both reach `_write_marker` concurrently (e.g., parallel bridge initialisation calls), they produce the identical temp-file path. One thread's `write_text` will silently truncate the other's in-progress write before `replace()` is called, which could leave a corrupt or empty JSON file in the marker position.

Add a thread-unique component (e.g., `threading.get_ident()`) to the suffix:
```python
import threading
temp = marker.with_name(f"{marker.name}.{os.getpid()}.{threading.get_ident()}.tmp")
```
This is not a concern if callers are always serialised by an external lock, but no such lock is visible at the call sites in `__init__.py` and `bridge_client.py`.

---

### 9. `apps/memos-local-plugin/adapters/hermes/memos_provider/runtime_home.py` (L136-L139)

`OSError` covers both `FileNotFoundError` (the `skills` directory simply does not exist — expected) and `PermissionError` (the directory exists but cannot be listed). In the `PermissionError` case, returning `False` silently treats a locked legacy home as having no meaningful data, which causes `select_windows_runtime_home` to select the canonical install root and potentially direct the application away from a home that does contain user data.

Consider narrowing the catch to `FileNotFoundError` for the expected case and either re-raising or logging `PermissionError`:
```python
try:
    return next(skills.iterdir(), None) is not None
except FileNotFoundError:
    return False
except OSError:
    logger.debug("cannot list skills dir %s; assuming no data", skills)
    return False
```

---

### 10. `apps/memos-local-plugin/core/config/defaults.ts` (L14-L17)

The function silently returns `undefined` when `agent` is a known string but not present in `FIXED_VIEWER_PORTS`. That case is currently unreachable (callers pass `agent` values read from the CLI, such as `"hermes"` or `"openclaw"`), but the type signature accepts any string, so future callers could pass an unrecognised agent and get `undefined` without any warning. Consider returning `undefined` only for the *expected* unknown-agent case and throwing (or at least logging) for truly unexpected values, or document the silent-miss behaviour explicitly in the JSDoc so callers are not surprised.

Additionally — and more critically — the semantics expressed by this function (and the JSDoc comment "own these well-known ports **even for legacy YAML files**") means that the returned port unconditionally overwrites whatever the user has configured in their `viewer.port` YAML field (see `resolveConfig` in `index.ts`). A user who intentionally changes `viewer.port` for a `hermes` or `openclaw` agent will have that setting silently discarded on every config load. If overriding user config is truly the intended contract, the JSDoc here (and the schema documentation) should call this out explicitly; if not, the override in `resolveConfig` should be applied only when the port has not been explicitly set by the user (i.e., when it still equals the default).

---

### 11. `apps/memos-local-plugin/core/config/index.ts` (L80-L83)

The `effectiveViewerPort` override unconditionally replaces `merged.viewer.port` for any known agent, even if the user has explicitly set a different port in their `config.yaml`. This silently disregards the user's configuration without any feedback.

Consider either:
1. Only applying the override when the port in `merged` matches the default schema value (18799), so a deliberate user-set value is respected.
2. Or, if the port truly must be locked for agent operation, emit a warning into the `warnings` array when the override changes an explicitly configured value, so the user is at least informed.

Suggested approach (emit a warning on override):
```ts
const viewerPort = effectiveViewerPort(agent);
if (viewerPort !== undefined && isPlainObject(merged.viewer)) {
  if (merged.viewer.port !== undefined && merged.viewer.port !== viewerPort) {
    warnings?.push(
      `viewer.port is locked to ${viewerPort} for agent '${agent}'; ignoring configured value ${merged.viewer.port}`
    );
  }
  merged.viewer.port = viewerPort;
}
```

**Before:**
```
  const viewerPort = effectiveViewerPort(agent);
  if (viewerPort !== undefined && isPlainObject(merged.viewer)) {
    merged.viewer.port = viewerPort;
  }
```

**After:**
```suggestion
  const viewerPort = effectiveViewerPort(agent);
  if (viewerPort !== undefined && isPlainObject(merged.viewer)) {
    if (merged.viewer.port !== undefined && merged.viewer.port !== viewerPort) {
      warnings?.push(
        `viewer.port is locked to ${viewerPort} for agent '${agent}'; ignoring configured value ${merged.viewer.port}`
      );
    }
    merged.viewer.port = viewerPort;
  }
```

</details>

---

### 12. `apps/memos-local-plugin/core/config/migrations.ts` (L91)

Temp file naming collision: `process.pid` + `Date.now()` is not unique when `atomicWrite` is called twice in rapid succession within the same process (which happens in `migrateHermesViewerPort` — once for `home.configFile`, once for `markerFile`). On a fast system both calls can land in the same millisecond, producing the same temp path for two different target files. The second `writeFile` silently overwrites the first temp file before the first `rename` completes, potentially writing the marker file's JSON content into `config.yaml` or vice-versa.

The existing `writer.ts` uses the same pattern but names the temp file after its specific target (`.config.${pid}.${ts}.tmp`), making cross-target collision impossible. Apply the same fix here by including the target basename in the temp name, or use a cryptographically random suffix.

Suggested fix:
```ts
import { randomBytes } from "node:crypto";
// ...
const tempFile = join(dirname(target), `.${basename(target)}.${process.pid}.${randomBytes(4).toString("hex")}.tmp`);
```

**Before:**
```
  const tempFile = join(dirname(target), `.${process.pid}.${Date.now()}.tmp`);
```

**After:**
```suggestion
  const tempFile = join(dirname(target), `.${basename(target)}.${process.pid}.${randomBytes(4).toString("hex")}.tmp`);
```

</details>

---

### 13. `apps/memos-local-plugin/core/config/migrations.ts` (L76-L79)

Missing delay after stale-lock removal: when the lock directory is stale and gets removed (line 74 `fs.rm`), the loop immediately `continue`s with no sleep. If another process races to recreate the lock directory in the same instant, this becomes a tight busy-loop for the remainder of the 100 attempts, wasting CPU. Adding a small sleep before the retry (consistent with the 25 ms sleep elsewhere in the loop) prevents this.

**Before:**
```
        if (Date.now() - stat.mtimeMs > 60_000) {
          await fs.rm(lockDir, { recursive: true, force: true });
          continue;
        }
```

**After:**
```suggestion
        if (Date.now() - stat.mtimeMs > 60_000) {
          await fs.rm(lockDir, { recursive: true, force: true });
          await new Promise((resolve) => setTimeout(resolve, 25));
          continue;
        }
```

</details>

---

### 14. `apps/memos-local-plugin/core/config/migrations.ts` (L66-L71)

Retry budget misaligned with stale-lock timeout: `acquireMigrationLock` retries 100 times with a 25 ms sleep each iteration, yielding a maximum wait of ~2.5 seconds before throwing `config_write_failed`. However the stale-lock threshold is 60 seconds, meaning a legitimate slow migrator (e.g., on slow I/O) holding the lock for longer than 2.5 s will cause callers to fail with a misleading error. Either increase the retry count/sleep to cover at least the stale-lock window, or derive them from a single named constant to keep them in sync.

---

### 15. `apps/memos-local-plugin/core/config/migrations.ts` (L93-L98)

On Windows, `fs.rename` over an existing target file throws `EPERM` instead of atomically replacing it (unlike POSIX). `atomicWrite` does not handle this case, so writing the marker file (which may already exist from a previous attempt) or overwriting `home.configFile` will fail with an unhandled `EPERM` on Windows. The installer (`install.ps1`) confirms Windows is a supported platform. Consider catching `EPERM`/`EXDEV` on Windows and falling back to a copy-then-delete, or using a library that handles cross-platform atomic rename.

---

### 16. `apps/memos-local-plugin/core/config/paths.ts` (L95-L96)

If `writeRuntimeHomeMarker` throws (e.g., disk full, permission denied on `installRoot`), the exception propagates out of `selectWindowsHermesRuntimeHome` and then out of `resolveHome`, crashing the startup even though a perfectly valid `selection` was already computed. Since persisting the marker is a best-effort optimisation, write failures should be caught and logged as a warning rather than treated as fatal.

**Before:**
```
  if (options.persist !== false) writeRuntimeHomeMarker(markerFile, selection);
  return selection;
```

**After:**
```suggestion
  if (options.persist !== false) {
    try {
      writeRuntimeHomeMarker(markerFile, selection);
    } catch {
      // Non-fatal: marker write failed (e.g. disk full, permission denied).
      // The correct home was still selected; a later startup will retry.
    }
  }
  return selection;
```

</details>

---

### 17. `apps/memos-local-plugin/core/config/paths.ts` (L77-L82)

This exception is thrown inside `resolveHome`, which is called early during startup (e.g. in `bootstrapMemoryCoreFull` and `loadConfigForAgent`). Neither call site wraps it in a try/catch, so on a Windows machine that has both databases the process crashes with an unhandled exception and the full filesystem paths (`legacyHome`, `installRoot`) are exposed in the error message. Consider either (a) returning the conflict as a structured result so callers can surface it cleanly, or (b) at minimum redacting the paths from the message and documenting that callers must catch.

**Before:**
```
  if (legacyDb && canonicalDb) {
    throw new Error(
      "both Windows Hermes runtime homes contain a database; " +
      `set MEMOS_HOME explicitly (${legacyHome} or ${installRoot})`,
    );
  }
```

**After:**
```suggestion
  if (legacyDb && canonicalDb) {
    throw new Error(
      "both Windows Hermes runtime homes contain a memos.db; " +
      "set MEMOS_HOME explicitly to choose one",
    );
  }
```

</details>

---

### 18. `apps/memos-local-plugin/core/config/paths.ts` (L124)

The temp-file name is composed only of PID and millisecond timestamp. On Windows, PIDs are recycled aggressively and `Date.now()` resolution can be as coarse as 15 ms, so two concurrent startups (e.g. two rapid restarts) can produce the same temp-file name. One process will overwrite the other's in-flight write before `renameSync`, resulting in a truncated or empty marker file. Use `crypto.randomUUID()` (available in Node 14.17+, no extra import needed) for a collision-proof suffix.

**Before:**
```
  const tempFile = `${markerFile}.${process.pid}.${Date.now()}.tmp`;
```

**After:**
```suggestion
  const { randomUUID } = await import("node:crypto");
  const tempFile = `${markerFile}.${randomUUID()}.tmp`;
```

</details>

---

### 19. `apps/memos-local-plugin/core/config/paths.ts` (L138-L142)

`readdirSync` can throw for reasons other than the directory not existing — most notably `EACCES` (permission denied) and `ENOTDIR`. Silently returning `false` in all of these cases causes `hasMeaningfulRuntimeData` to incorrectly report an unreadable `skills/` directory as absent, which can steer the runtime to the wrong home without any diagnostic. Only `ENOENT` should be suppressed.

**Before:**
```
  try {
    return readdirSync(join(root, "skills")).length > 0;
  } catch {
    return false;
  }
```

**After:**
```suggestion
  try {
    return readdirSync(join(root, "skills")).length > 0;
  } catch (err) {
    if ((err as NodeJS.ErrnoException).code === "ENOENT") return false;
    throw err;
  }
```

</details>

---

### 20. `apps/memos-local-plugin/core/config/paths.ts` (L113-L116)

`pathResolve(value.path)` is called after only an `isAbsolute` check. On Windows, `isAbsolute` returns `true` for UNC paths (`\\server\share\...`), so a crafted marker file can redirect the runtime home to an attacker-controlled network share, enabling data exfiltration or NTLM hash capture. Validate that the resolved path starts with one of the two expected roots (`legacyHome` or `installRoot`) before accepting it.

Note: `readRuntimeHomeMarker` does not currently have access to those root values. Consider adding them as parameters, or performing the allowlist check in `selectWindowsHermesRuntimeHome` after `readRuntimeHomeMarker` returns.

---

### 21. `apps/memos-local-plugin/core/config/paths.ts` (L50-L54)

The `persist` field is optional but defaults to `true` (write side-effect) when omitted. This is not documented in the interface or the JSDoc of `selectWindowsHermesRuntimeHome`, which is misleading: callers who pass only `{ legacyHome, installRoot }` expecting a pure selection will unknowingly trigger a file write. Document the default explicitly.

**Before:**
```
interface WindowsHermesHomeOptions {
  legacyHome: string;
  installRoot: string;
  persist?: boolean;
}
```

**After:**
```suggestion
interface WindowsHermesHomeOptions {
  legacyHome: string;
  installRoot: string;
  /**
   * When `true` (the default), the resolved home is persisted to the marker
   * file so future startups skip re-detection. Pass `false` for a dry-run
   * selection that causes no side effects.
   */
  persist?: boolean;
}
```

</details>

---

### 22. `apps/memos-local-plugin/core/config/writer.ts` (L66-L72)

When the patch explicitly contains `viewer.port`, `applyPatch` has already applied the user-supplied value. This block then silently overwrites it with the fixed Hermes port, discarding the user's intent without any warning or error. If the goal is that Hermes must always use port 18800, the appropriate approach is to validate (and reject with an error) rather than silently mutate. If silent override is intentional, at least log a warning so the user understands why their setting was ignored.

**Before:**
```
  if (
    agent === "hermes" &&
    isPlainObject(patch.viewer) &&
    Object.hasOwn(patch.viewer, "port")
  ) {
    doc.setIn(["viewer", "port"], effectiveViewerPort("hermes"));
  }
```

**After:**
```suggestion
  if (
    agent === "hermes" &&
    isPlainObject(patch.viewer) &&
    Object.hasOwn(patch.viewer, "port") &&
    patch.viewer.port !== effectiveViewerPort("hermes")
  ) {
    throw new MemosError(
      "config_invalid",
      `The viewer port for the Hermes agent is fixed to ${effectiveViewerPort("hermes")} and cannot be changed.`,
      { source: home.configFile },
    );
  }
```

</details>

---

### 23. `apps/memos-local-plugin/core/config/writer.ts` (L71)

The string literal `"hermes"` is hardcoded here, but the enclosing `if` block already guarantees `agent === "hermes"`. Passing the variable `agent` instead is more consistent and avoids a subtle divergence if `agent` is later refined (e.g. a sub-variant string).

**Before:**
```
    doc.setIn(["viewer", "port"], effectiveViewerPort("hermes"));
```

**After:**
```suggestion
    doc.setIn(["viewer", "port"], effectiveViewerPort(agent));
```

</details>

---

### 24. `apps/memos-local-plugin/core/config/writer.ts` (L44)

If `migrateHermesViewerPort` throws (e.g., lock timeout, backup conflict, or fs error), the exception propagates directly from `patchConfig` with no additional context, making it hard for callers to distinguish a migration failure from a normal config write failure. Consider wrapping the call so the error surface is clear, or at minimum ensuring the error message identifies it as a migration issue.

**Before:**
```
  if (agent === "hermes") await migrateHermesViewerPort(home);
```

**After:**
```suggestion
  if (agent === "hermes") {
    try {
      await migrateHermesViewerPort(home);
    } catch (err) {
      throw new MemosError(
        "config_write_failed",
        `Hermes port migration failed, cannot proceed with config write: ${(err as Error).message}`,
        { source: home.configFile, cause: (err as Error).message },
      );
    }
  }
```

</details>

---

### 25. `apps/memos-local-plugin/core/pipeline/memory-core.ts` (L2142-L2146)

The conflict branch (`existing.turnKey === turnKey` but `existing.userText !== turn.userText`) sets `shouldWriteApiLog = true` and leaves `apiLogClaim = null`, then writes a log row — but never updates the map entry. The map still holds the original `userText`. Every subsequent call with the same turnKey and the conflicting userText will re-enter this branch, evaluate `existing.userText !== turn.userText` as `true` again, and write another log row. The deduplication mechanism is permanently broken for this turnKey after a conflict: the very duplicate rows this change intends to prevent will be created on each retry.

If the conflict is truly rejected upstream (as the comment states), the write at `shouldWriteApiLog = true` is also superfluous. If it can slip through, the map entry must be updated to the new userText after a successful write so that identical retries are suppressed:
```ts
// After a successful write in the conflict case, update the map:
if (shouldWriteApiLog && apiLogWritten && existing) {
  existing.userText = turn.userText;
}
```
Alternatively, set `apiLogClaim` in the conflict branch as well so the standard cleanup path handles it consistently.

**Before:**
```
      if (existing?.turnKey === turnKey) {
        // The orchestrator will reject a reused key with different text. Keep
        // that genuine conflict observable while suppressing exact replays.
        shouldWriteApiLog = existing.userText !== turn.userText;
      } else {
```

**After:**
```suggestion
      if (existing?.turnKey === turnKey) {
        // The orchestrator will reject a reused key with different text. Keep
        // that genuine conflict observable while suppressing exact replays.
        shouldWriteApiLog = existing.userText !== turn.userText;
        // If the conflict case is allowed through and written, update the map
        // so that an identical retry of the conflicting call is suppressed.
        if (shouldWriteApiLog) {
          existing.userText = turn.userText;
        }
      } else {
```

</details>

---

### 26. `apps/memos-local-plugin/install.ps1` (L167-L174)

If `Prepare-StagedPackage` throws, `$StagedPrefix` is never assigned — PowerShell leaves the variable undefined (or retains a stale value from a previous call in the same session). The catch block then evaluates `Test-Path $StagedPrefix` against an empty/null string, which PowerShell resolves as `Test-Path ""` — this either throws or returns false. More critically, on some PowerShell versions a null/empty path passed to `Remove-Item` can resolve to the current working directory.

Initialize `$StagedPrefix = $null` before calling `Prepare-StagedPackage` and add a null-guard in the catch:
```powershell
if ($StagedPrefix -and (Test-Path $StagedPrefix)) {
    Remove-Item -Recurse -Force $StagedPrefix -ErrorAction SilentlyContinue
}
```

**Before:**
```
    $StagedPrefix = Prepare-StagedPackage
    $BackupDir = "$Prefix.memos-backup-$([guid]::NewGuid().ToString('N'))"
    $HadExisting = Test-Path $Prefix
    $LiveMovedToBackup = $false
    $StagedMovedLive = $false
    $DeploySucceeded = $false

    try {
```

**After:**
```suggestion
    $StagedPrefix = $null
    try {
        $StagedPrefix = Prepare-StagedPackage
    } catch {
        throw
    }
    $BackupDir = "$Prefix.memos-backup-$([guid]::NewGuid().ToString('N'))"
    $HadExisting = Test-Path $Prefix
    $LiveMovedToBackup = $false
    $StagedMovedLive = $false
    $DeploySucceeded = $false

    try {
```

</details>

---

### 27. `apps/memos-local-plugin/install.ps1` (L208-L216)

The null-path risk for `$StagedPrefix` mentioned above also applies here. If `Prepare-StagedPackage` throws before assigning `$StagedPrefix`, the catch-block `Test-Path $StagedPrefix` receives a null/empty string.

Add a null guard: `if ($StagedPrefix -and (Test-Path $StagedPrefix))`.

**Before:**
```
        if ($StagedMovedLive -and (Test-Path $Prefix)) {
            Remove-Item -Recurse -Force $Prefix -ErrorAction SilentlyContinue
        }
        if ($LiveMovedToBackup -and (Test-Path $BackupDir)) {
            Move-Item -Path $BackupDir -Destination $Prefix -Force
        }
        if (Test-Path $StagedPrefix) {
            Remove-Item -Recurse -Force $StagedPrefix -ErrorAction SilentlyContinue
        }
```

**After:**
```suggestion
        if ($StagedMovedLive -and (Test-Path $Prefix)) {
            Remove-Item -Recurse -Force $Prefix -ErrorAction SilentlyContinue
        }
        if ($LiveMovedToBackup -and (Test-Path $BackupDir)) {
            Move-Item -Path $BackupDir -Destination $Prefix -Force
        }
        if ($StagedPrefix -and (Test-Path $StagedPrefix)) {
            Remove-Item -Recurse -Force $StagedPrefix -ErrorAction SilentlyContinue
        }
```

</details>

---

### 28. `apps/memos-local-plugin/install.ps1` (L724-L730)

Passing `-ArgumentList` as a single pre-escaped string is unreliable on Windows. When `$HomeDir` or `$BridgeEntry` contains spaces, parentheses, or ampersands — common in user-profile paths like `C:\Users\John Smith\...` — the Windows process creation API may mis-tokenize the argument string, causing the daemon to fail silently.

Pass arguments as an array instead, which is how `Start-Process` is designed to handle them:

**Before:**
```
        if ($BridgeEntry.EndsWith(".cjs")) {
            $DaemonArgs = "`"$BridgeEntry`" --agent=hermes --daemon --home=`"$HomeDir`""
            Start-Process -FilePath $NodeBin -ArgumentList $DaemonArgs -WindowStyle Hidden -RedirectStandardOutput $DaemonLog -RedirectStandardError $DaemonLogErr
        } else {
            $DaemonArgs = "`"$TsxBin`" `"$BridgeEntry`" --agent=hermes --daemon --home=`"$HomeDir`""
            Start-Process -FilePath $NodeBin -ArgumentList $DaemonArgs -WindowStyle Hidden -RedirectStandardOutput $DaemonLog -RedirectStandardError $DaemonLogErr
        }
```

**After:**
```suggestion
        if ($BridgeEntry.EndsWith(".cjs")) {
            $DaemonArgs = @($BridgeEntry, "--agent=hermes", "--daemon", "--home=$HomeDir")
            Start-Process -FilePath $NodeBin -ArgumentList $DaemonArgs -WindowStyle Hidden -RedirectStandardOutput $DaemonLog -RedirectStandardError $DaemonLogErr
        } else {
            $DaemonArgs = @($TsxBin, $BridgeEntry, "--agent=hermes", "--daemon", "--home=$HomeDir")
            Start-Process -FilePath $NodeBin -ArgumentList $DaemonArgs -WindowStyle Hidden -RedirectStandardOutput $DaemonLog -RedirectStandardError $DaemonLogErr
        }
```

</details>

---

### 29. `apps/memos-local-plugin/install.ps1` (L580-L587)

The guard only replaces the `hooks` object when it is absent, non-object, or an array. It does **not** check whether `allowConversationAccess` is already explicitly set to `false` by host/admin policy. The unconditional assignment on the last line silently overrides a deliberate `false`, escalating plugin privileges without user awareness.

Preserve an explicit `false` if one is already present:

**Before:**
```
    if (
      !config.plugins.entries[pluginId].hooks ||
      typeof config.plugins.entries[pluginId].hooks !== 'object' ||
      Array.isArray(config.plugins.entries[pluginId].hooks)
    ) {
      config.plugins.entries[pluginId].hooks = {};
    }
    config.plugins.entries[pluginId].hooks.allowConversationAccess = true;
```

**After:**
```suggestion
    if (
      !config.plugins.entries[pluginId].hooks ||
      typeof config.plugins.entries[pluginId].hooks !== 'object' ||
      Array.isArray(config.plugins.entries[pluginId].hooks)
    ) {
      config.plugins.entries[pluginId].hooks = {};
    }
    if (config.plugins.entries[pluginId].hooks.allowConversationAccess !== false) {
      config.plugins.entries[pluginId].hooks.allowConversationAccess = true;
    }
```

</details>

---

### 30. `apps/memos-local-plugin/install.ps1` (L410)

`$PID` is the current process ID, which is not unique enough for a temp-file suffix. If a previous installer run crashed leaving a stale `.tmp` file with the same PID (after OS PID reuse), the `Move-Item` could overwrite it without issue, but `[IO.File]::WriteAllText` failing mid-write would leave the orphan with no cleanup path. Every other temp path in this script uses `[guid]::NewGuid()` for uniqueness — this should be consistent.

**Before:**
```
    $TempFile = "$MarkerFile.$PID.tmp"
```

**After:**
```suggestion
    $TempFile = "$MarkerFile.$([guid]::NewGuid().ToString('N')).tmp"
```

</details>

---

### 31. `apps/memos-local-plugin/install.ps1` (L211-L213)

When rollback is triggered on a failed deploy, the backup directory is restored via `Move-Item` (correct), but there is no message telling the operator where the backup lives if the restore itself fails, or in other failure paths where `$BackupDir` is left on disk. Adding a `Write-Warn` after the catch would help users locate orphaned backup directories.

**Before:**
```
        if ($LiveMovedToBackup -and (Test-Path $BackupDir)) {
            Move-Item -Path $BackupDir -Destination $Prefix -Force
        }
```

**After:**
```suggestion
        if ($LiveMovedToBackup -and (Test-Path $BackupDir)) {
            Move-Item -Path $BackupDir -Destination $Prefix -Force
        } elseif ($LiveMovedToBackup) {
            Write-Warn "Rollback: backup directory not found at $BackupDir"
        }
        # Inform the operator if the backup is left on disk
        if (-not $StagedMovedLive -and (Test-Path $BackupDir)) {
            Write-Warn "Backup left at: $BackupDir"
        }
```

</details>

---

### 32. `apps/memos-local-plugin/server/routes/admin.ts` (L51-L64)

When `core.shutdown()` throws, `scheduleWindowsShutdownAfterResponse` is called immediately, scheduling a process exit even though cleanup has not been attempted and the response signals `ok: false, cleared: false`. On Windows there is no supervisor to restart the process, so the caller's next retry will find the server gone. The comment in the file header explicitly states "keeps the responding process alive so the route cannot self-destruct before a replacement exists", which contradicts this behaviour.

Consider omitting the shutdown schedule in the error path and only exiting after a successful clear:

**Before:**
```
      try {
        await deps.core.shutdown();
      } catch (err) {
        scheduleWindowsShutdownAfterResponse(ctx.res, options);
        return {
          ok: false,
          cleared: false,
          restarting: false,
          manualRestartRequired: true,
          platform,
          error: `Memory core did not shut down cleanly: ${errorMessage(err)}`,
          message: manualClearRestartMessage(agent, false),
        };
      }
```

**After:**
```suggestion
      try {
        await deps.core.shutdown();
      } catch (err) {
        // Do NOT schedule shutdown — leave the process alive so the caller can retry.
        return {
          ok: false,
          cleared: false,
          restarting: false,
          manualRestartRequired: true,
          platform,
          error: `Memory core did not shut down cleanly: ${errorMessage(err)}`,
          message: manualClearRestartMessage(agent, false),
        };
      }
```

</details>

---

### 33. `apps/memos-local-plugin/server/routes/admin.ts` (L173-L178)

`fs.access` checks whether the calling process has permission to access the path — it does not reliably indicate non-existence. On Windows, a file that was successfully `unlink`ed but is still held open by another handle will make `access` throw `EACCES` (not `ENOENT`), causing the catch block to suppress the error and fall through without pushing to `failures`. Conversely, on some configurations `access` may succeed for a deleted-but-still-open file, falsely marking it as a failure.

Use `fs.stat` and check for `ENOENT` to determine absence:

**Before:**
```
    try {
      await fs.access(target);
      failures.push(target);
    } catch {
      /* absent as required */
    }
```

**After:**
```suggestion
    try {
      await fs.stat(target);
      // File still exists after unlink — held open by another process.
      failures.push(target);
    } catch (statErr) {
      if ((statErr as NodeJS.ErrnoException).code !== "ENOENT") {
        // Unexpected error; treat as failure to be safe.
        failures.push(target);
      }
      /* ENOENT → absent as required */
    }
```

</details>

---

### 34. `apps/memos-local-plugin/server/routes/admin.ts` (L146-L147)

`process.pid` is embedded in the user-facing instruction string. On Windows there is no supervisor to guarantee process identity is stable: if the process is replaced between the time the response is sent and the user reads the message, the PID could belong to a completely different process. Killing an arbitrary recycled PID could terminate an unrelated application.

Additionally, exposing the internal PID in an unauthenticated (or lightly authenticated) API response is an information disclosure risk in networked deployments.

Consider instructing the user to stop the named process instead, which is unambiguous and PID-independent:

**Before:**
```
            `Configuration saved. Close Hermes, run Stop-Process -Id ${process.pid} ` +
            "in PowerShell to stop Memory Viewer, then start Hermes again.",
```

**After:**
```suggestion
            "Configuration saved. Close Hermes, then in PowerShell run " +
            "Get-Process -Name 'node' | Where-Object { $_.MainWindowTitle -eq 'Memory Viewer' } | Stop-Process " +
            "(or close the Memory Viewer window directly), then start Hermes again.",
```

</details>

---

### 35. `apps/memos-local-plugin/server/routes/admin.ts` (L120-L122)

The expression `options.lifecycle?.platform ?? process.platform` is resolved independently inside the `openclaw` block and again inside the `hermes` block of `POST /api/v1/admin/restart` (lines 121 and 138). The same expression is also used at the top of the `clear-data` handler. If the platform-detection logic ever changes it must be updated in all three places. Hoist the resolution to a single variable at the top of each route handler, as was already done in the `clear-data` handler:

**Before:**
```
    if (agent === "openclaw") {
      const platform = options.lifecycle?.platform ?? process.platform;
      if (platform === "win32") {
```

**After:**
```suggestion
  routes.set("POST /api/v1/admin/restart", async (_ctx) => {
    const agent = options.agent ?? "unknown";
    const platform = options.lifecycle?.platform ?? process.platform;
    if (agent === "openclaw") {
      if (platform === "win32") {
```

</details>

---

### 36. `apps/memos-local-plugin/server/routes/admin.ts` (L217-L219)

Once `finish` fires and the shutdown is scheduled, the `close` and `error` listeners remain attached to the `ServerResponse` for the entire 300 ms delay period. While minor per-request, these dangling listeners can cause confusion if the response stream emits additional events after finishing. Use a named handler and clean up the other listeners once one fires:

**Before:**
```
  res.once("finish", () => schedule(300));
  res.once("close", () => schedule(res.writableFinished ? 300 : 1_000));
  res.once("error", () => schedule(1_000));
```

**After:**
```suggestion
  const onFinish = (): void => { cleanup(); schedule(300); };
  const onClose = (): void => { cleanup(); schedule(res.writableFinished ? 300 : 1_000); };
  const onError = (): void => { cleanup(); schedule(1_000); };
  const cleanup = (): void => {
    res.off("finish", onFinish);
    res.off("close", onClose);
    res.off("error", onError);
  };
  res.once("finish", onFinish);
  res.once("close", onClose);
  res.once("error", onError);
```

</details>

---

### 37. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L1464-L1487)

`platform_name` is iterated but never used in the test body — the `subTest` label is the only reference. All three loop iterations run identical patches against identical state, so the loop provides no additional coverage and is misleading. The `_LOOPBACK_OPENER` (a `ProxyHandler({})` opener) is process-level state shared across all platforms; the only way to test platform-specific proxy bypass would be to swap out the opener or mock the OS-level proxy resolver. As written, the three subtests are fully redundant.

Either remove the loop and run the assertion once, or add genuine platform-differentiation (e.g., mocking `urllib.request.getproxies`) so each subtest exercises something different.

---

### 38. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L1509-L1527)

The `HTTPError` object is constructed before the `try` block. If the constructor raises an exception (e.g., due to an unexpected `HTTPError` API change), `error` would be unbound when the `finally` block executes `error.close()`, converting the original exception into a `NameError` and masking the real failure.

Move the construction inside the `try` block, or use `contextlib.closing`:
```python
try:
    error = urllib.error.HTTPError(...)
    with patch.object(...):
        ...
finally:
    error.close()
```

**Before:**
```
        error = urllib.error.HTTPError(
            "http://127.0.0.1:18800/test",
            401,
            "Unauthorized",
            {},
            None,
        )
        try:
            with patch.object(
                daemon_manager_mod._LOOPBACK_OPENER,
                "open",
                side_effect=error,
            ):
                self.assertEqual(
                    daemon_manager_mod._probe_json_url("http://127.0.0.1:18800/test"),
                    "occupied",
                )
        finally:
            error.close()
```

**After:**
```suggestion
        try:
            error = urllib.error.HTTPError(
                "http://127.0.0.1:18800/test",
                401,
                "Unauthorized",
                {},
                None,
            )
            with patch.object(
                daemon_manager_mod._LOOPBACK_OPENER,
                "open",
                side_effect=error,
            ):
                self.assertEqual(
                    daemon_manager_mod._probe_json_url("http://127.0.0.1:18800/test"),
                    "occupied",
                )
        finally:
            error.close()
```

</details>

---

### 39. `apps/memos-local-plugin/tests/python/test_bridge_client.py` (L1570-L1572)

The `tempfile.TemporaryDirectory()` context manager is opened but its value is never used or needed — `_probe_loopback_port(0)` only attempts a socket bind, not any filesystem operation. This creates a misleading impression that the test depends on a temporary directory. Remove the `with` wrapper.

**Before:**
```
    def test_bind_probe_confirms_a_free_loopback_port(self) -> None:
        with tempfile.TemporaryDirectory():
            self.assertEqual(daemon_manager_mod._probe_loopback_port(0), "free")
```

**After:**
```suggestion
    def test_bind_probe_confirms_a_free_loopback_port(self) -> None:
        self.assertEqual(daemon_manager_mod._probe_loopback_port(0), "free")
```

</details>

---

### 40. `apps/memos-local-plugin/viewer/src/components/RestartOverlay.tsx` (L73)

`AgentType` is locally redeclared here as `type AgentType = 'openclaw' | 'hermes'`, but `restart.ts` already exports an identical type as `RestartAgent`. This duplication means the two definitions can silently diverge if a new agent variant is added to the store. Import and reuse the already-exported type instead.

Suggested fix:
```ts
import {
  restartState,
  dismissRestartBanner,
  resolveRestartAgent,
  type RestartPhase,
  type RestartAgent,  // add this
} from "../stores/restart";

// Remove the local AgentType alias entirely
type AgentType = RestartAgent;  // or just use RestartAgent directly
```

---

### 41. `apps/memos-local-plugin/viewer/src/components/RestartOverlay.tsx` (L123-L132)

`isTerminalPhase()` uses a hardcoded string array with `.includes()` to classify terminal phases, with no compile-time linkage to the `RestartPhase` union. If a new terminal phase is added to `RestartPhase` in the future, this function will silently return `false` for it — causing the UI to show a spinner and no dismiss button with no type error to warn the developer.

Consider using an explicit record or satisfies pattern to enforce exhaustiveness:
```ts
const TERMINAL_PHASES = new Set<RestartPhase>([
  "restartFailed",
  "manualRestartRequired",
  "manualClearRestartRequired",
  "clearFailed",
  "clearResultUnknown",
  "manualCloseRequired",
]);

function isTerminalPhase(phase: RestartPhase): boolean {
  return TERMINAL_PHASES.has(phase);
}
```
This still doesn't enforce exhaustiveness at compile time, but at minimum avoids re-creating the array on every call. For full exhaustiveness, restructure as a switch with an explicit return per branch matching `overlayMessage()`.

---

### 42. `apps/memos-local-plugin/viewer/src/components/RestartOverlay.tsx` (L110-L117)

Multiple `overlayHint()` cases use `as any` to construct dynamic i18n keys (e.g., `` `restart.clearFailedHint.${agentType}` as any ``). While `agentType` is always `"openclaw" | "hermes"` at runtime, the `as any` casts bypass TypeScript's key validation entirely across five call sites simultaneously. The i18n translation files confirm all these keys exist, but any future key rename or agent addition will go undetected.

A safer pattern uses explicit key mapping to preserve type safety:
```ts
case "clearFailed":
  return agentType === "openclaw"
    ? t("restart.clearFailedHint.openclaw")
    : t("restart.clearFailedHint.hermes");
```
This is more verbose but keeps the type checker engaged for each key.

---

### 43. `apps/memos-local-plugin/viewer/src/stores/i18n.ts` (L1103-L1104)

This string is missing a trailing `。` (Chinese full stop), which is inconsistent with every other newly added ZH sentence in this block (e.g., `restart.manual`, `restart.clearing`, `restart.manualHint.hermes`, `restart.clearComplete`, etc. all end with `。`). The same issue exists on the updated `restart.failedHint.openclaw` line in the ZH locale below.

**Before:**
```
  "restart.manualHint.openclaw":
    "请在 PowerShell 中依次执行：openclaw gateway stop；openclaw gateway start",
```

**After:**
```suggestion
  "restart.manualHint.openclaw":
    "请在 PowerShell 中依次执行：openclaw gateway stop；openclaw gateway start。",
```

</details>

---

### 44. `apps/memos-local-plugin/viewer/src/stores/i18n.ts` (L1117-L1118)

Same missing trailing `。` as `restart.manualHint.openclaw` above. The EN counterpart ends with a period (`...openclaw gateway start.`), and all surrounding ZH strings are terminated with `。`.

**Before:**
```
  "restart.failedHint.openclaw":
    "请在 PowerShell 中依次执行：openclaw gateway stop；openclaw gateway start",
```

**After:**
```suggestion
  "restart.failedHint.openclaw":
    "请在 PowerShell 中依次执行：openclaw gateway stop；openclaw gateway start。",
```

</details>

---

### 45. `apps/memos-local-plugin/viewer/src/stores/restart.ts` (L171-L172)

The guard condition is inverted. The intent (per the `beginClearData` contract) is: if the phase is already `"clearing"`, the agent was already locked by `beginClearData()` — skip re-locking. But `!== "clearing"` means it calls `lockRestartAgent()` on every path *except* the `beginClearData` path, which is the opposite. Any caller that correctly calls `beginClearData()` first will have their locked agent silently overwritten here with the current health value, potentially returning a wrong agent if health is already offline at the point `triggerCleared` runs.

Suggestion: flip the condition.
```ts
if (restartState.value.phase !== "clearing") lockRestartAgent();
// should be:
if (restartState.value.phase === "clearing") { /* already locked */ } else { lockRestartAgent(); }
// i.e.:
if (restartState.value.phase !== "clearing") lockRestartAgent();
```
Wait — re-reading: the guard as written calls `lockRestartAgent()` when phase is NOT `"clearing"`. That means when `beginClearData()` was called (phase IS `"clearing"`), the lock is *skipped* — which is actually the correct intent. The issue is the opposite: when a direct caller skips `beginClearData()`, this line will lock from health at an arbitrary point. This is an undocumented assumption that `beginClearData()` must always precede `triggerCleared()`. The contract should be enforced explicitly or documented.

Recommendation: add a guard that throws or logs a warning if `lockedRestartAgent` is null and phase is not `"clearing"`, rather than silently falling back to a potentially stale health read.

**Before:**
```
  if (restartState.value.phase !== "clearing") lockRestartAgent();
  restartState.value = { phase: "restarting" };
```

**After:**
```suggestion
  if (restartState.value.phase !== "clearing") {
    // Direct call without beginClearData(): lock now, but health may be stale.
    // Prefer always calling beginClearData() before triggerCleared().
    lockRestartAgent();
  }
  // else: agent was already locked by beginClearData()
```

</details>

---

### 46. `apps/memos-local-plugin/viewer/src/stores/restart.ts` (L177-L184)

The `!response.ok` check fires before `response.manualRestartRequired` is checked. Looking at `admin.ts`, a Windows partial-failure path returns `{ ok: false, manualRestartRequired: true, ... }`. Because `!response.ok` is checked first, such a response will land on `"clearFailed"` instead of `"manualClearRestartRequired"`, showing the wrong UI state and losing the manual restart instruction. The `manualRestartRequired` check should come before the generic `!ok` check.

**Before:**
```
  if (response && !response.ok) {
    restartState.value = { phase: "clearFailed" };
    return;
  }
  if (response?.manualRestartRequired) {
    restartState.value = { phase: "manualClearRestartRequired" };
    return;
  }
```

**After:**
```suggestion
  if (response?.manualCloseRequired) {
    restartState.value = { phase: "manualCloseRequired" };
    return;
  }
  if (response?.manualRestartRequired) {
    restartState.value = { phase: "manualClearRestartRequired" };
    return;
  }
  if (response && !response.ok) {
    restartState.value = { phase: "clearFailed" };
    return;
  }
```

</details>

---

### 47. `apps/memos-local-plugin/viewer/src/stores/restart.ts` (L145-L151)

For the `openclaw` path, the POST is wrapped in a try/catch that silently swallows errors (to handle the case where the server exits before responding). However, the Windows OpenClaw path returns `manualRestartRequired: true` synchronously *before* any shutdown — the server deliberately stays alive to send the response. If a network-level error occurs on that path, the catch discards the response and the function falls through to `pollHealthUntilUp(60)` instead of showing the manual restart instruction, leaving Windows users polling for 150 s before seeing a `"restartFailed"` state. Consider differentiating a connection error (server went down) from an HTTP error (server is still up but request failed).

Memtensor-AI:

All tests passed (69/69 executed). memos_local_plugin/changed-repo-python: 69/69. Duration: 2s [advisory, non-gating] AI-generated tests on branch test/auto-gen-e245d1faac77ccd5-20260806224217: 62/67 passed, 5 failed — these do NOT affect the PR verdict; review the branch manually.

**Branch:** `fix/plugin-windows-runtime-lifecycle`

PR #2234, opened 2026-08-10
fix(llm): preserve all vLLM stream chunks

JiataiWang:
## Description

`VLLMLLM.generate_stream()` handled `delta.content` after the stream loop, so it only saw the final delta. Responses split across several content chunks were truncated, and an empty or choices-free stream could raise `UnboundLocalError` because `delta` was never assigned.

This moves content handling into the loop and closes an open `<think>` block when a reasoning-only stream ends. It keeps the existing `remove_think_prefix` behavior unchanged and adds focused regression coverage for multi-chunk content, empty/choices-free streams, reasoning-only streams, and output with think tags disabled.

No dependencies are added. I couldn't find an existing issue or open PR for this path.

Related Issue (Required): N/A

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [x] Unit Test
- [x] Test Script Or Test Steps

Before the fix, the new regression file produced 2 failures and 2 errors. On the pushed commit:

- `uv run --frozen --with 'pytest==8.3.5' pytest tests/llms/test_vllm.py -q` — 4 passed
- `uv run --frozen --with 'pytest==8.3.5' pytest tests/llms/ --ignore=tests/llms/test_hf.py -q` — 18 passed (4 existing warnings)
- `uv run --frozen --with 'poetry>=2,<3' --with 'ruff==0.11.13' make format` — Ruff passed; 617 files unchanged
- `git diff --check origin/dev-v2.0.29...HEAD` — passed

The complete `tests/llms/` collection was not available locally because `test_hf.py` imports the optional `torch` dependency.

## Checklist

- [x] I have performed a self-review of my own code | 我已自行检查了自己的代码
- [x] I have commented my code in hard-to-understand areas (no hard-to-understand code added) | 我已在难以理解的地方对代码进行了注释
- [x] I have added tests that prove my fix is effective or that my feature works | 我已添加测试以证明我的修复有效或功能正常
- [x] I have created related documentation issue/PR in MemOS-Docs (not applicable; no documentation behavior changed) | 我已在 MemOS-Docs 中创建了相关的文档 issue/PR（如适用）
- [x] I have linked the issue to this PR (not applicable; no existing issue found) | 我已将 issue 链接到此 PR（如适用）
- [x] I have mentioned the person who will review this PR | 我已提及将审查此 PR 的人

@WeiminLee when you have a moment, could you take a look?

## Reviewer Checklist

[...]

PR #1959, opened 2026-06-21
feat(l3): dedicated l3Llm config slot for abstraction pass

chiefmojo:
## Problem

L3 abstraction runs on the main `llm`. On cheap models (e.g. `gemini-2.5-flash-lite`) this over-extracts and truncates the JSON output, producing `'constraints'/'inference' must be an array` errors. The slot is quality-sensitive and runs off the turn-response path — a slower but more capable model here improves world-model quality with no impact on latency.

## Changes

Adds an `l3Llm` top-level config slot using the same `SkillEvolverSchema` shape as `skillEvolver`. When blank (the default), behavior is identical to today — `deps.l3Llm ?? deps.llm` falls through to the main LLM. When set, the L3 clustering → abstraction subscriber uses the dedicated client.

```yaml
l3Llm:
  provider: openai_compatible
  endpoint: https://openrouter.ai/api/v1
  model: anthropic/claude-sonnet-4-5
  apiKey: sk-or-...
```

- `defaults.ts` — blank default (inherits main llm)
- `schema.ts` — `l3Llm: SkillEvolverSchema` entry
- `memory-core.ts` — bootstrap client (mirrors `reflectLlm` path; errors log, don't crash)
- `deps.ts` / `types.ts` — `PipelineDeps.l3Llm` and `PipelineHandle.l3Llm` fields
- `orchestrator.ts` — threads the field through to the handle
- `orchestrator.test.ts` — two tests: dedicated client threads through, null when unconfigured

## Config example

Operators running a budget model as their main `llm` can now point L3 at a stronger model:

```yaml
llm:
  provider: openai_compatible
  model: gemini-2.5-flash-lite  # cheap, fast, good for scoring

l3Llm:
  provider: openai_compatible
  model: anthropic/claude-sonnet-4-5  # slow ok — L3 is async
  apiKey: sk-or-...
```

PR #2235, opened 2026-08-11
fix(config): resolve masked apiKey from env on load

### LINKED ISSUE #2245: bug: masked apiKey written to config.yaml is never resolved on daemon restart — LLM auth fails and bridge restart-loops
## Pre-submission checklist

- [x] I have searched existing issues and PRs — no open issue covers this; PR #2235 (open) is the fix
- [x] I have a minimal reproduction (below)
- [x] I have verified the root cause in the code (below)

## Bug Description

The bridge persists `config.yaml` with API keys masked to `__memos_secret__` (`maskSecrets()` in `core/pipeline/memory-core.ts`) and strips empty secrets from patches (`stripEmptySecrets()`). Nothing ever re-reads the real value back: when the daemon restarts, `loadConfig()` parses the mask as the literal API key, every LLM call fails auth, and the bridge restart-loops with `lastOkAt: null` while skill crystallize stays stuck (`skill.crystallize.failed ... timed out after 120000 ms` spam).

## How to Reproduce

1. Configure an `llm.apiKey` via the viewer/UI (which masks it to `__memos_secret__` on disk).
2. Restart the bridge daemon.
3. Observe: `lastOkAt: null`, LLM auth failures on every call, `skill.crystallize.failed` errors, restart loop.

## Root Cause

`resolveConfig()` (the single choke point for both disk-loaded and in-memory patched configs) never resolves masked/placeholder secret values back from the environment. Read-side resolution is missing entirely — only the write-side masking exists.

## Environment

- **Bridge:** `@memtensor/memos-local-plugin` 2.0.12-beta.1 (source matches `apps/memos-local-plugin`)
- **Trigger:** migration to an openai-compatible provider (opencode-go, `deepseek-v4-flash`). Config written via the viewer masks the key; restart loops; `curl` with the env key works while the bridge with the masked key fails.
- **Platform:** Debian 12 LXC (Hermes CT100), node 22, systemd `memos-bridge.service`

## Additional Context

Observed live 2026-08-11 with 290 candidate skills backlogged and `skill.crystallize.failed ... openai_compatible timed out after 120000 ms` in the journal.

## Willingness to Implement

Fix is in open PR #2235 (`fix(config): resolve masked apiKey from env on load`) — read-side only, on-disk masking preserved. Includes 6 unit tests; 54/54 pass.

kiwipaulrob:
## Summary

The bridge persists `config.yaml` with API keys masked to `__memos_secret__` (`maskSecrets()` in `core/pipeline/memory-core.ts`) and strips empty secrets from patches (`stripEmptySecrets()`). But nothing ever re-reads the real value back: when the daemon restarts, `loadConfig()` parses the mask as the literal API key, every LLM call fails auth, and the bridge restart-loops with `lastOkAt: null` while skill crystallize stays stuck. This was observed live on 2026-08-11 with 290 candidate skills backlogged and `skill.crystallize.failed ... openai_compatible timed out after 120000 ms` spam in the journal.

This PR makes `resolveConfig()` (the single choke point for both disk-loaded and in-memory patched configs) resolve masked/placeholder secret values from the environment, **read-side only** — the on-disk write stays masked, so the security posture of `maskSecrets()` is preserved.

## Change

`apps/memos-local-plugin/core/config/index.ts` — `resolveConfig(raw)` now walks `SECRET_FIELD_PATHS` before `pruneUnknown`/`deepMerge`:

- `${ENV_VAR}` references in any secret field resolve from `process.env[ENV_VAR]`
- `__memos_secret__` / empty-string `apiKey` fields fall back to `LLM_API_KEY`, `EMBEDDING_API_KEY`, then `OPENCODE_GO_API_KEY` / `OPENCODE_ZEN_API_KEY`
- Hub tokens (`hub.teamToken`, `hub.userToken`) have no env convention and are left untouched unless the user writes an explicit `${VAR}`
- Real (non-placeholder) values pass through unchanged

## Tests

New `tests/unit/config/resolve-secret-env.test.ts` (6 tests):

1. `${ENV_VAR}` expansion
2. `__memos_secret__` mask resolution
3. empty-string fallback
4. all `apiKey` `SECRET_FIELD_PATHS` leaves resolve (hub tokens stay masked)
5. real values untouched
6. placeholder stays when no env var is set

```
Test Files  4 passed (4)
     Tests  54 passed (54)   # incl. full existing config suite (load/paths/writer)
```

`tsc -p tsconfig.json --noEmit` passes clean.

## Related

- **Fixes #2245** — the tracked issue for this bug (masked apiKey never resolved on restart).

- **MemOS #1959** `feat(l3): dedicated l3Llm config slot for abstraction pass` — added the `l3Llm.apiKey` secret path this PR also covers; without the env fallback, L3 abstraction fails with `l3.abstraction: 'inference' must be an array` when the key is masked.
- **MemOS #2234** `fix(llm): preserve all vLLM stream chunks` — same class of LLM-call robustness: this PR removes the auth-failure half of the `timed out after 120000 ms` burst seen on `capture.reflect.scoring` (16-step) runs that overload single-slot local models.
- **Hermes #56569 / #56581 / #58055** — the compound-channel-ID email bug follows the same "masked write without read path → silent failure" pattern; listed as the cross-project archetype.
- **Hermes #44370** `MemOS skillInjectionMode full but memos_skill_get never called` — downstream symptom class where MemOS config/LLM wiring fails silently; this PR removes one root cause.

## Environment

- **Bridge:** `@memtensor/memos-local-plugin` 2.0.12-beta.1 (source matches `apps/memos-local-plugin`)
- **Trigger:** migration to the opencode-go provider (`https://opencode.ai/zen/go/v1`, `deepseek-v4-flash`). Config written via the UI/viewer masks the key, restart loops, `curl` with the env key works (`GO_DEEPSEEK_OK`), bridge with masked key fails.
- **Platform:** Debian 12 LXC (Hermes CT100), node 22, systemd `memos-bridge.service`

## Type of change

## How Has This Been Tested?

- [x] Unit Test (`vitest run tests/unit/config` — 54 passed)
- [x] Test Script Or Test Steps (live bridge: after patching dist the same way, `lastOkAt` populates and crystallize drains; verified 2026-08-11 on Hermes CT100)

## Checklist

- [x] My code follows the style guidelines of this project (no new deps, existing patterns)
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (PR body + code comments; doc note pending maintainer preference)
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally

### DIFF
--- a/apps/memos-local-plugin/core/config/defaults-pr.ts
+++ b/apps/memos-local-plugin/core/config/defaults-pr.ts
@@ -1,0 +1,371 @@
+/**
+ * The default config tree. Mirrors `schema.ts` exactly. When merging YAML,
+ * we deep-merge over this tree so users only need to specify what they want
+ * to change.
+ */
+
+import type { ResolvedConfig } from "./schema.js";
+
+const FIXED_VIEWER_PORTS: Readonly<Record<string, number>> = Object.freeze({
+  openclaw: 18799,
+  hermes: 18800,
+});
+
+/** Runtime adapters own these well-known ports even for legacy YAML files. */
+export function effectiveViewerPort(agent?: string): number | undefined {
+  return agent ? FIXED_VIEWER_PORTS[agent] : undefined;
+}
+
+export const DEFAULT_CONFIG: ResolvedConfig = {
+  version: 1,
+  viewer: {
+    // Per-agent default lives in `templates/config.<agent>.yaml`:
+    //   - openclaw → 18799
+    //   - hermes   → 18800
+    // The fallback here only matters when neither config file exists
+    // (early bootstrap, tests, etc.).
+    port: 18799,
+    bindHost: "127.0.0.1",
+    openOnFirstTurn: false,
+  },
+  bridge: {
+    port: 18911,
+    mode: "stdio",
+  },
+  embedding: {
+    provider: "local",
+    endpoint: "",
+    model: "Xenova/all-MiniLM-L6-v2",
+    apiKey: "",
+    providerIgnore: [],
+    providerOrder: [],
+    openRouter: false,
+    cache: {
+      enabled: true,
+      maxItems: 20_000,
+    },
+  },
+  llm: {
+    provider: "",
+    endpoint: "",
+    model: "",
+    temperature: 0,
+    fallbackToHost: true,
+    apiKey: "",
+    timeoutMs: 45_000,
+    maxRetries: 3,
+    providerIgnore: [],
+    providerOrder: [],
+    openRouter: false,
+    maxTokens: 1024,
+    headers: {},
+  },
+  l3Llm: {
+    // Empty by default — falls back to the shared `llm` settings.
+    // Operators set this when they want a stronger model for L3
+    // abstraction. L3 runs off the turn-response path, so a slower
+    // but more reliable model improves world-model quality without
+    // affecting companion latency.
+    provider: "",
+    endpoint: "",
+    model: "",
+    apiKey: "",
+    temperature: 0,
+    timeoutMs: 60_000,
+    providerIgnore: [],
+    providerOrder: [],
+    openRouter: false,
+    maxTokens: 1024,
+  },
+  skillEvolver: {
+    // Empty by default — falls back to the shared `llm` settings.
+    // Operators set this when they want a stronger model (e.g.
+    // claude-sonnet / gpt-5-thinking) for skill crystallisation.
+    provider: "",
+    endpoint: "",
+    model: "",
+    apiKey: "",
+    temperature: 0,
+    timeoutMs: 60_000,
+    providerIgnore: [],
+    providerOrder: [],
+    openRouter: false,
+    maxTokens: 1024,
+  },
+  storage: {
+    ftsTokenizer: "trigram",
+  },
+  algorithm: {
+    lightweightMemory: {
+      enabled: true,
+    },
+    capture: {
+      maxTextChars: 4_000,
+      maxToolOutputChars: 2_000,
+      embedTraces: true,
+      alphaScoring: true,
+      // OpenClaw's tool messages don't include explicit "reflection"
+      // blocks; without synthesis the alpha scorer sees an empty
+      // reflection and forces α = 0 (see `core/capture/alpha-scorer.ts`
+      // line 97). That makes reflection-weighted backprop degenerate
+      // into pure γ-discount and produces flat V distributions —
+      // L2 association + skill crystallization both starve. Enable
+      // synth by default so even turns without explicit reflections
+      // still contribute useful α values.
+      synthReflections: true,
+      llmConcurrency: 4,
+      // Bound topic-end reflect work so dirty startup recovery cannot replay
+      // a huge historical episode into thousands of paid LLM calls.
+      maxReflectLlmCalls: 128,
+      // Recovered episodes are reconstructed from persisted traces; replay
+      // orphans are usually matching drift, so do not insert duplicate rows.
+      maxRecoveryOrphanInserts: 0,
+      // V7 §3.2 batched variant. With "auto" we issue a single LLM call
+      // per episode for both reflection synth and α scoring as long as
+      // the episode is short enough — this collapses 2N per-step calls
+      // (N synth + N α) into 1 batched call. Long episodes (>12 steps)
+      // automatically fall back to the per-step path so the prompt
+      // never overflows the model's context window. R_human + backprop
+      // remain task-end events handled by `core/reward`, unchanged.
+      batchMode: "auto",
+      batchThreshold: 12,
+      // `reflectionContextMode` controls which extra prompt context blocks
+      // topic-end reflection receives:
+      //   - "none": no TASK CONTEXT and no DOWNSTREAM STEP PREVIEW
+      //   - "task": inject TASK CONTEXT only
+      //   - "downstream": inject DOWNSTREAM STEP PREVIEW only
+      //   - "task_downstream": inject both blocks
+      // `longEpisodeReflectMode` controls the fallback used when an episode is
+      // too long for batch scoring:
+      //   - "per_step_parallel": keep the current parallel per-step path. Each
+      //     step is reflected independently, using only the context blocks
+      //     enabled by `reflectionContextMode` that are available without
+      //     downstream preview.
+      //   - "per_step_downstream": still run per-step work in parallel, but
+      //     prebuild a bounded DOWNSTREAM STEP PREVIEW for each step (step+1
+      //     through step+N, capped by `downstreamStepCount`) and inject it when
+      //     `reflectionContextMode` includes "downstream".
+      reflectionContextMode: "task_downstream",
+      longEpisodeReflectMode: "per_step_downstream",
+      downstreamStepCount: 3,
+      taskContextMaxChars: 800,
+      downstreamContextMaxChars: 1_200,
+      downstreamPerStepMaxChars: 400,
+      synthOutcomeMaxChars: 600,
+    },
+    reward: {
+      gamma: 0.9,
+      tauSoftmax: 0.5,
+      decayHalfLifeDays: 30,
+      llmScoring: true,
+      implicitThreshold: 0.2,
+      // 10 minutes was too long for interactive chat — users moved on
+      // to the next task before reward ever fired, so no R_human was
+      // ever computed and V stayed 0 for every trace. 30 s gives the
+      // user a short window to reply ("thanks", "no, try again") that
+      // the scorer picks up as explicit feedback; when nothing
+      // arrives, the implicit fallback fires promptly so downstream
+      // L2/L3/Skill stages aren't starved of signal.
+      feedbackWindowSec: 30,
+      summaryMaxChars: 2_000,
+      llmConcurrency: 2,
+      // Default lowered 2→1 to support single-shot CLI patterns
+      // (`hermes chat -q "..."`, `openclaw run --once`). With the old
+      // floor every CLI single-query episode was abandoned with
+      // "对话轮次不足", starving reward → L2 → Skill of any signal.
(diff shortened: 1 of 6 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2235
**Task:** `746e1ab7fb4fd29d`
**Base:** `main`
**Head:** `fix/config-apikey-env-fallback`

---

### 1. `apps/memos-local-plugin/core/config/index.ts` (L173-L178)

When `hub.teamToken` or `hub.userToken` is masked as `__memos_secret__` (by `maskSecrets()`), this branch silently `continue`s with no warning. The literal string `"__memos_secret__"` then passes through into the resolved config and is sent as a credential to the hub, causing silent auth failures that are very hard to diagnose.

A warning should be emitted when a masked non-`apiKey` secret field is detected but cannot be auto-resolved, so users get actionable feedback instead of an opaque auth error.

**Before:**
```
    } else if (val === "__memos_secret__" || val === "") {
      // Masked/empty API key — infer the env var from the field path.
      // Only apiKey fields have a convention (OPENAI_API_KEY, etc.);
      // hub tokens (teamToken/userToken) have no env convention, so
      // they must be set explicitly via ${VAR} or the UI.
      if (leaf !== "apiKey") continue;
```

**After:**
```suggestion
    } else if (val === "__memos_secret__" || val === "") {
      // Masked/empty API key — infer the env var from the field path.
      // Only apiKey fields have a convention (OPENAI_API_KEY, etc.);
      // hub tokens (teamToken/userToken) have no env convention, so
      // they must be set explicitly via ${VAR} or the UI.
      if (leaf !== "apiKey") {
        if (val === "__memos_secret__") {
          warnings?.push(
            `config: '${dotted}' is masked but has no auto-resolution rule — ` +
              `set it explicitly via \${VAR} syntax or through the UI`
          );
        }
        continue;
      }
```

</details>

---

### 2. `apps/memos-local-plugin/core/config/index.ts` (L187-L192)

The `OPENCODE_GO_API_KEY` / `OPENCODE_ZEN_API_KEY` fallbacks apply to every LLM-class `apiKey` field (`llm.apiKey`, `l3Llm.apiKey`, `skillEvolver.apiKey`). These fields belong to different, independently configured providers. If a user has `OPENCODE_GO_API_KEY` set for one provider, that key will be silently injected into all other masked LLM apiKey fields — including those belonging to unrelated third-party providers (e.g. Anthropic, Gemini). This constitutes cross-provider key leakage: the key is sent to an endpoint it was never intended for.

The fallback should be tied to a specific provider identity (e.g. checking `llm.provider`), or the generic cross-provider fallback should be removed entirely in favour of requiring explicit `${VAR}` references.

---

### 3. `apps/memos-local-plugin/core/config/index.ts` (L179)

The `isEmbedding` check relies on the parent key at position `keys.length - 2` being literally `"embedding"`. This is a positional/string-name heuristic that is fragile relative to schema evolution. For example, if a future schema nests embedding config more deeply (e.g. `embedding.remote.apiKey`), the parent key would be `"remote"`, `isEmbedding` would be `false`, and the LLM generic fallbacks (`OPENCODE_GO_API_KEY`) would be incorrectly applied to an embedding API key — precisely the cross-contamination the comment warns against.

A more robust approach is to derive the embedding check from the top-level section name (`keys[0]`), which is stable regardless of nesting depth.

**Before:**
```
      const isEmbedding = keys[keys.length - 2] === "embedding";
```

**After:**
```suggestion
      const isEmbedding = keys[0] === "embedding";
```

</details>

Memtensor-AI:

**Details:** Executor error: Command failed: git merge --no-edit base/main
**Branch:** `fix/config-apikey-env-fallback`

kiwipaulrob:
Thanks for the review — all four findings are fair, and I've folded them into the revised branch (now rebased onto the current main, 8d310a7a).

**1. Provider fallbacks leaking across secret fields (L126–L130)** — agreed, good catch. The OPENCODE_GO/ZEN fallbacks were meant to spare opencode-go/zen users from defining a second env var, but applying them to `embedding.apiKey`, and to explicit `${VAR}` references whose named variable is unset, is wrong: it can hand an LLM key to an embedding provider's calls, or silently substitute a different key than the one the user named. Fix: an explicit `${VAR}` now resolves exactly that variable or stays untouched, and the `OPENCODE_*` fallbacks apply only on the inferred sentinel/empty path, only for LLM-class fields (parent != `embedding`). Tests cover both cases: `embedding.apiKey` resolves from `EMBEDDING_API_KEY` and never from an LLM key, and an unset `${VAR}` stays literal even when generic keys are present.

**2. Unrestricted `${VAR}` expansion (L110–L112)** — agreed that expansion should be bounded, with one adjustment to the proposed allowlist: restricting to `*_API_KEY` names would break a designed use case of this PR, namely hub tokens (`hub.teamToken` / `hub.userToken`) set via explicit `${VAR}` — they have no other env convention. The allowlist is now `^[A-Z][A-Z0-9_]*_(API_KEY|TOKEN)$`; names outside it warn and are left untouched. I'd rate the practical severity here as modest for a local single-user plugin (user-owned config file, auth-gated viewer patch path), but the same change has a robustness payoff beyond the threat model: it stops accidental expansion of unrelated variables (a typo'd `${HOME}` or `${PATH}`) from becoming a bogus credential — so I'm happy to take it.

**3. In-place mutation of `raw` (L130)** — fair. Both current call sites are safe (`loadConfig` parses fresh YAML; `writer.ts` passes `doc.toJS()`), but the signature didn't promise non-mutation. Rather than just documenting the contract I made the pass non-mutating: `pruneUnknown` already returns a fresh copy, so resolution now runs on that cleaned object — identical semantics (user-provided leaves only), no writes to the caller's object. The JSDoc states the input is never mutated, and a test asserts the raw object is untouched after resolution.

**4. Shadowed `leaf` declaration (L113–L118)** — agreed, a leftover from an earlier edit. Removed the inner declaration; the outer one is in scope.

**On the automated test run ("ENV ISSUE": `git merge --no-edit base/main` failed)** — that was a genuine merge conflict rather than an environment problem: main had since changed this same file (the viewer-port migration, #2230, gives `loadConfig`/`resolveConfig` an `agent` parameter and `effectiveViewerPort` handling), and GitHub reported the branch as not mergeable (`mergeable_state: dirty`). I've merged the latest main into the branch, keeping both sets of changes (agent/viewer-port handling + env resolution), and pushed the rebased branch. The executor should pick the update up on the next run; if a manual re-run is needed, that would be appreciated.

Validation on the revised branch: config suite **71/71** (5 files — includes main's new `hermes-migration` tests) and `tsc --noEmit` clean.

Thanks again — the embedding-key leak in particular was a sharp catch.

Memtensor-AI:

All tests passed (10/10 executed). memos_local_plugin/unit: 10/10. Duration: 3s

**Branch:** `fix/config-apikey-env-fallback`

kiwipaulrob:
**Deployment note (from the production deployment that hit this)**

The fix resolves masked keys from the daemon's process environment — which only works if the daemon actually has them. For systemd users:

```ini
# /etc/systemd/system/memos-bridge.service.d/env.conf
[Service]
EnvironmentFile=/path/to/durable/bridge.env
```

where `bridge.env` is a filtered copy of your secrets file, e.g. `grep -v '^CUSTOM:' .env > <plugin-home>/daemon/bridge.env && chmod 600 <plugin-home>/daemon/bridge.env`.

**⚠️ Do NOT place the EnvironmentFile in /tmp.** It is wiped on reboot; the unit then crash-loops with `Failed to load environment files: No such file or directory` (result 'resources', NRestarts +1 per 5s), and because the daemon is down a session host's `ensure_viewer_daemon()` spawns a rogue `bridge.mjs --daemon` that binds the viewer port — so `/api/v1/health` keeps answering while the systemd service is dead (observed live 2026-08-12 after a container reboot; the daemon must be restarted so systemd binds :18800 first).

Env var contract implemented by this PR:
- generic: `LLM_API_KEY` (llm/skillEvolver/l3Llm apiKey), `EMBEDDING_API_KEY` (embedding.apiKey only — never an LLM key)
- OpenCode fallbacks (LLM-class only): `OPENCODE_GO_API_KEY` → `OPENCODE_ZEN_API_KEY`
- explicit `${VAR}`: allowlisted names matching `^[A-Z][A-Z0-9_]*_(API_KEY|TOKEN)$` (hub tokens use this — they have no other convention)

Memtensor-AI:

All tests passed (10/10 executed). memos_local_plugin/unit: 10/10. Duration: 3s [advisory, non-gating] AI-generated tests on branch test/auto-gen-1f83be90d13816eb-20260814063936: 47/47 passed — these do NOT affect the PR verdict; review the branch manually.

**Branch:** `fix/config-apikey-env-fallback`

Memtensor-AI:

All tests passed (10/10 executed). memos_local_plugin/unit: 10/10. Duration: 4s [advisory, non-gating] AI-generated tests on branch test/auto-gen-746e1ab7fb4fd29d-20260815032618: 40/40 passed — these do NOT affect the PR verdict; review the branch manually.

**Branch:** `fix/config-apikey-env-fallback`

Memtensor-AI:

**Details:** Environment preparation failed before any gating tests executed. Failed scopes: memos_local_plugin
**Branch:** `fix/config-apikey-env-fallback`

Hun-ger:
Thanks for identifying and thoroughly testing the masked-secret restart issue. The core fix has now been incorporated through #2246, followed by additional hardening in e1530908: path-specific environment resolution, provider/endpoint-scoped OpenCode fallbacks, prevention of cross-provider key leakage, and correct handling of intentionally empty optional secrets.

The packaged plugin has also been installed and validated on both macOS and Windows: masked disk secrets are restored successfully, the bridge reaches `pipeline.ready`, and no configuration warnings are emitted.

Since this PR is now superseded, and its remaining changes include unrelated generated/variant files, we will not merge it directly. Closing in favor of #2246 and e1530908. Thank you for the contribution and detailed investigation.

Issue #1808, opened 2026-05-27
fix: [memos-local-plugin] Dreaming background processing starves Gateway event loop, causing WebSocket timeouts


tianxin8206:
### Pre-submission checklist | 提交前检查

- [x] I have searched existing issues and this hasn't been mentioned before | 我已搜索现有问题，确认此问题尚未被提及
- [x] I have read the project documentation and confirmed this issue doesn't already exist | 我已阅读项目文档并确认此问题尚未存在
- [x] This issue is specific to MemOS and not a general software issue | 该问题是针对 MemOS 的，而不是一般软件问题

### Bug Description | 问题描述

The memos-local-plugin's "dreaming" feature performs LLM inference calls synchronously on the Gateway's main Node.js event loop. When processing orphan or dirty episodes, this blocks the event loop for extended periods (observed eventLoopMax of 3.5s+), causing WebSocket handshake timeouts for OpenClaw TUI and Control UI connections.

### How to Reproduce | 如何重现

**Steps to Reproduce** 
1. Enable memos-local-plugin with dreaming enabled
2. Accumulate a large number of episodes (30K+ traces in our case)
3. Have one or more episodes enter an orphan/dirty state (e.g., r_task = NULL)
4. Restart the openclaw gateway
5. Attempt to connect via TUI or Control UI during dreaming processing

**Observed Behavior**

- OpenClaw Gateway event loop blocked for 3-5+ seconds during dreaming LLM calls
- WebSocket upgrade requests timeout (default 3s budget for read probe)
- TUI reports "cannot connect to gateway"
- Control UI login page fails to load
- Gateway logs show repeated Read probe: failed - timeout errors
- Orphan episodes with failed LLM calls are retried indefinitely with no backoff

### Environment | 环境信息

- OpenClaw version: 2026.5.22
- OS: openeuler (systemd user service)
- memos.db size: 324MB, 30K+ traces
- Model provider: DeepSeek (v4-flash for dreaming)

### Willingness to Implement | 实现意愿

- [ ] I'm willing to implement this myself | 我愿意自己解决
- [ ] I would like someone else to implement this | 我希望其他人来解决

Memtensor-AI:
## Sentinel Classification

| Field | Value |
|-------|-------|
| Module | `plugin` |
| Priority | `P2:normal` |
| Auto-fixable | No |

**Reason:** Performance issues require profiling and benchmarking that the agent cannot perform.

This issue has been classified and requires human attention.

---
[...]

PR #2252, opened 2026-08-15
fix(core): bound startup recovery wait in shutdown to 15s

kiwipaulrob:
## Problem

`core.shutdown()` awaited `startupRecoveryPromise` with **no timeout**. With a large dirty episode and a slow/flaky LLM, the startup-recovery reflect chain (up to 163 sequential LLM calls at ~2min each) can take minutes — holding shutdown hostage until the systemd kill timer fires.

Observed 15 Aug 2026 on a production bridge: SIGTERM at 08:00:23, daemon stuck mid-recovery, systemd `TimeoutStopSec=600` expired, SIGKILL at 08:10:23. A 10-minute stop-sigterm wedge from a single unguarded await.

## Fix

Wrap the recovery wait in the existing `withTimeout()` helper (15s):

```ts
await withTimeout(startupRecoveryPromise, 15_000, "startup_recovery_shutdown_timeout");
```

**Why 15s is safe:**
- The wait was introduced for issue #1808 to prevent a fast `init → shutdown` race closing SQLite mid-flush. A 15s grace still covers that (recovery of 0–1 episodes finishes in ms).
- Recovery is **resumable by design**: dirty episodes carry `rewardDirty.failedAttempts`, and the periodic 10-minute rescore re-runs them. Proceeding after the grace loses no data.
- After the timeout, `handle.shutdown()` detaches subscribers and the daemon's `process.exit(0)` fires — lingering LLM calls can no longer hold the process hostage.

## Verification

- `tsc --noEmit` clean
- Full unit suite: 168 files / 1365 tests passed (incl. the issue #1808 recovery-stall test)
- Production companion change: `TimeoutStopSec` tightened 600s → 90s on the systemd unit (ops-side belt-and-braces; with this fix graceful shutdown completes in ~20s)

### DIFF
--- a/apps/memos-local-plugin/core/pipeline/memory-core.ts
+++ b/apps/memos-local-plugin/core/pipeline/memory-core.ts
@@ -1933,7 +1933,14 @@ export function createMemoryCore(
       // gateway reload would close SQLite while reflect / reward is
       // mid-flush, producing `SQLITE_MISUSE` noise on the way down.
       try {
-        await startupRecoveryPromise;
+        // Bound the wait: a slow / flaky LLM during startup recovery of a
+        // large dirty episode must not hold shutdown hostage until the
+        // systemd kill timer (15 Aug 2026 stop-sigterm wedge: SIGTERM at
+        // 08:00:23, SIGKILL at 08:10:23). Recovery is resumable — dirty
+        // episodes carry rewardDirty.failedAttempts and the periodic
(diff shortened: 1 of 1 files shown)


Memtensor-AI:

All tests passed (41/41 executed). memos_local_plugin/unit: 41/41. Duration: 12s

**Branch:** `fix/shutdown-recovery-timeout`

PR #1998, opened 2026-06-30
Fix #1736: [Bug] memos-local-hermes-plugin: bridge fails to start on Node.js v22 (CJS/ESM m

Memtensor-AI:
## Description

Fixes #1736 — Hermes plugin bridge fails to start on Node 22 with `Cannot read properties of undefined (reading 'exports')`. Root cause is the CJS↔ESM trampoline inside `dist/bridge.cjs` (CommonJS wrapper bootstrapping a pure-ESM core via a `new Function("...import...")` indirection). Although commit d62ff452 already mitigated the `__dirname` and `require(file://URL)` symptoms, the boundary remained structurally fragile across Node minor versions.

This change introduces a pure-ESM entry point: source `apps/memos-local-plugin/bridge.mts` compiles to `dist/bridge.mjs`. The new entry uses `import.meta.url`/`fileURLToPath` for path resolution, static ESM imports for Node built-ins, and plain `await import("./core/pipeline/index.js")` for peer ESM modules — no `new Function` trampoline, no `require(file://)`. The Python launcher (`bridge_client.py` + `daemon_manager.py`) now resolves the bridge script with the precedence `dist/bridge.mjs` → `dist/bridge.cjs` → `bridge.mts` → `bridge.cts`, treating `.mjs`/`.cjs` as bare `node <script>` and `.mts`/`.cts` as `tsx` invocations. The legacy `bridge.cts` and `dist/bridge.cjs` are preserved so installations that have not refreshed continue to work unchanged.

Verification on Node v22.22.2 (the reporter's exact version): the new `node dist/bridge.mjs --agent=hermes --no-viewer` reaches `bootstrapMemoryCoreFull` cleanly and never emits the historical `'exports'` symptom — downstream failure is only the unrelated `better-sqlite3` native binding (sandbox installed with `--ignore-scripts`). 11 new Python unit cases cover script-resolution + node/tsx command shape; a new vitest integration test spawns `dist/bridge.mjs` against an isolated `HOME` and asserts the symptom string is absent. Type-check and the full Python + integration + bridge/hermes vitest suites pass (37 + 29 + tsc clean); 2 pre-existing storage test failures on `bugfix/autodev-1736` are unchanged by this diff (also fail on stash-clean HEAD).

[...]

PR #2279, opened 2026-08-23
fix(bridge): start status heartbeat in --daemon mode

### LINKED ISSUE #2278: [Bug] memos-local-plugin: --daemon entry never starts the status heartbeat — health reports "stale" forever while RPC works
### Summary

Since #1998, the Hermes launcher prefers the pure-ESM `dist/bridge.mjs` entry over `dist/bridge.cjs`. However, `bridge.mts` only starts the `bridge-status.json` status heartbeat in its **stdio** branch (`if (!args.daemon)`). The **`--daemon`** branch never calls `markConnected()` / `startHeartbeat()` — so no daemon ever refreshes the status file.

The legacy `bridge.cts` daemon path has both calls right after the viewer port binds; the ESM successor lost them in translation.

### Impact

Any `--daemon` spawn writes no status updates. Readers (health endpoint, external monitors) apply the 20s staleness rule to whatever snapshot is on disk and permanently report:

```
status: "disconnected"
lastError: "Hermes bridge heartbeat is stale"
```

…while the bridge is actually healthy: RPC (`POST /api/v1/rpc`) keeps working normally because it never touches status tracking. Monitoring built on the health endpoint therefore shows a dead bridge during fully functional operation.

### Repro

1. `node dist/bridge.mjs --agent=hermes --daemon`
2. Watch `<home-root>/bridge-status.json`: once stale from a previous run, its mtime/content never changes again after daemon start (no initial write either)
3. ~20s later, the health endpoint reports `bridge.status: disconnected`, `"Hermes bridge heartbeat is stale"` — indefinitely

Observed in production after a v2.0.7 → v2.0.16 jump: the status file froze at the *pre-upgrade* daemon's last write, and every respawned daemon (fresh PID) left it untouched.

### Root cause

`apps/memos-local-plugin/bridge.mts` — `markConnected()` / `startHeartbeat()` appear only inside `if (!args.daemon)`. The `if (args.daemon)` block has none. Compare `bridge.cts`, which calls both after successful viewer bind in daemon mode.

### Suggested fix

Mirror the `.cts` daemon path — two lines after the bind-retry loop succeeds:

```ts
bridgeStatus?.markConnected();
bridgeHeartbeat = bridgeStatus?.startHeartbeat();
```

PR incoming with the fix plus a source-level regression guard test (same pattern as the existing `bridge-startup-ordering.test.ts`).

chiefmojo:
## Fix: start the status heartbeat in `--daemon` mode

Closes #2278

### Problem

#1998 made the Hermes launcher prefer `dist/bridge.mjs`, but `bridge.mts` only calls `markConnected()` / `startHeartbeat()` in its stdio (`!args.daemon`) branch. Daemon spawns therefore never refresh `bridge-status.json`; every reader applies the 20s staleness rule to a fossilized snapshot and reports `"Hermes bridge heartbeat is stale"` indefinitely — while RPC traffic keeps working, since it never touches status tracking.

### Change

Mirrors the legacy `bridge.cts` daemon path: after the bind-retry loop succeeds, call

```ts
bridgeStatus?.markConnected();
bridgeHeartbeat = bridgeStatus?.startHeartbeat();
```

Both calls sit on the success path (after the loop's `break;`) so failure exits don't write a false "connected".

Plus a source-level regression guard test (`tests/unit/bridge/bridge-daemon-heartbeat.test.ts`), following the established pattern of `bridge-startup-ordering.test.ts`: the entry point is a top-level executable script, so runtime-testing two call sites doesn't justify refactoring it into an injectable function. The guard scopes to the daemon branch specifically — `markConnected()` in the stdio branch does not satisfy it.

### Validation

```
cd apps/memos-local-plugin
npx vitest run tests/unit/bridge/ tests/unit/bridge-status.test.ts
# 8 files / 39 tests passed (includes the new regression guard,
# which fails on main without the fix)
npm run build   # dist/bridge.mjs emits both call pairs
```

The new test was verified RED against unpatched `main` (fails with "daemon branch never calls markConnected()") and GREEN with this patch.

### DIFF
--- a/apps/memos-local-plugin/bridge.mts
+++ b/apps/memos-local-plugin/bridge.mts
@@ -410,6 +410,14 @@ async function main(): Promise<void> {
       }
     }
 
+    // Port bound successfully — start reporting. Mirrors the legacy
+    // bridge.cts daemon path: without these calls the daemon never
+    // refreshes bridge-status.json and every reader applies the 20s
+    // staleness rule to a fossilized snapshot ("Hermes bridge heartbeat
+    // is stale") while RPC traffic keeps working.
+    bridgeStatus?.markConnected();
+    bridgeHeartbeat = bridgeStatus?.startHeartbeat();
+
     const shutdownDaemon = async (sig: string) => {
       process.stderr.write(`bridge: daemon received ${sig}, shutting down\n`);
       removeOwnedPidFile();
--- a/apps/memos-local-plugin/tests/unit/bridge/bridge-daemon-heartbeat.test.ts
+++ b/apps/memos-local-plugin/tests/unit/bridge/bridge-daemon-heartbeat.test.ts
@@ -1,0 +1,75 @@
+/**
+ * bridge.mts daemon-mode heartbeat regression guard.
+ *
+ * Symptom observed 2026-08-22 after deploying memos-local-plugin v2.0.16
+ * on a Hermes host: `POST /api/v1/rpc` served traffic normally, but
+ * `bridge-status.json` froze at its pre-upgrade timestamp and every health
+ * report said `disconnected` / "Hermes bridge heartbeat is stale".
+ *
+ * Root cause: PR #1998 made the Python launcher prefer the pure-ESM
+ * `dist/bridge.mjs` entry (fixing #1736). `bridge.mts` only called
+ * `markConnected()` / `startHeartbeat()` in its stdio (`!args.daemon`)
+ * branch — the `--daemon` branch never started the status heartbeat, so
+ * no daemon ever refreshed `bridge-status.json` and the 20s staleness rule
+ * permanently reported a dead bridge. The legacy `bridge.cts` daemon path
+ * always had both calls; the ESM successor lost them in translation.
+ *
+ * Why source-level (rather than runtime): `bridge.mts` is a top-level
+ * executable script — same rationale as `bridge-startup-ordering.test.ts`.
+ * Refactoring it into an injectable function just to runtime-test two call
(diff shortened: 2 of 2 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2279
**Task:** `e1d58e077b181e1f`
**Base:** `main`
**Head:** `fix/bridge-daemon-status-heartbeat`

✅ **OpenCodeReview**: Review skipped: no items were selected.

Memtensor-AI:

**Details:** Executor error: Command failed: git clone --depth 1 --branch fix/bridge-daemon-status-heartbeat git@github.com:chiefmojo/MemOS-contrib.git /data/test-workspaces/e1d58e077b181e1f/repo
Cloning into '/data/test-workspaces/e1d58e077b181e1f/repo'...
kex_exchange_identification: Connection closed by remote host
Connection closed by UNKNOWN port 65535
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
**Branch:** `fix/bridge-daemon-status-heartbeat`

chiefmojo:
Closing in favor of #2280 — it is the more complete upstream fix: dual-entry coverage (the `.cts` daemon branch has the same gap on `main`) plus heartbeat teardown in the shutdown handlers. Our version only patched `.mts` because our fork independently carries the `.cts` daemon-side calls.

Verified #2280's regression guard ourselves: RED on pristine `main` (both entries), GREEN on its own branch, mergeable state clean. Production evidence for the underlying bug is on #2278; fix confirmed live on our deployment.

PR #2285, opened 2026-08-26
fix(plugin): remember unsupported reasoning-effort routes per bridge

asyade:
## Problem

`prepareAuxiliaryCall` in the DeepSeek Harness adapter probes DSH'"'"'s branded `off` reasoning effort before **every** auxiliary LLM call, and falls back to an effort-less preparation when the exact model rejects it. That fallback is correct — but for models that advertise **no reasoning efforts at all** (e.g. a hand-declared OpenAI-completions entry), the probe fails *every* time, so each summarize / retrieval-filter call pays one doomed `prepareCall`, one logged `UNSUPPORTED_REASONING_EFFORT` error row, and a retry before doing any real work.

Observed in production while running the plugin (2.0.16) against DeepSeek Harness with OpenCode'"'"'s `x-preview-f-free`: the harness-side api log accumulated **1,220 failed** `capture.summarize` preparations vs 191 successes, and **1,267 failed** retrieval filters vs 106 ok — every success reached only through the fallback retry, doubling round trips and flooding the log.

## Fix

Each bridge now remembers the routes that rejected the probe (keyed by `provider/model`) and skips the probe on later calls:

- first call for a route: probe `off` → on `UNSUPPORTED_REASONING_EFFORT`, remember the route, prepare without effort (unchanged behavior);
- later calls for a remembered route: go straight to the effort-less preparation;
- models that *do* support `off` are unaffected — they keep getting the probe;
- non-unsupported errors still propagate untouched and do not poison the cache.

The memory is scoped **per bridge instance** (closure state, not module-global), so concurrent bridges stay independent and unit tests remain hermetic across cases.

## Tests

- new: `"remembers an unsupported-effort route and probes it only once per bridge"` — asserts the exact preparation sequence `[off, plain, plain]` across two completes;
- new: `"keeps remembered effort routes scoped to a single bridge instance"` — guards against reintroducing module-global state;
- existing suite passes unchanged (`20/20` in `tests/unit/adapters/deepseek-harness-host-llm.test.ts`);
- `tsc -p tsconfig.json --noEmit` clean.

### DIFF
--- a/apps/memos-local-plugin/adapters/deepseek-harness/host-llm.ts
+++ b/apps/memos-local-plugin/adapters/deepseek-harness/host-llm.ts
@@ -96,6 +96,9 @@ export interface CreateDeepSeekHarnessHostLlmBridgeOptions {
 export function createDeepSeekHarnessHostLlmBridge(
   options: CreateDeepSeekHarnessHostLlmBridgeOptions,
 ): HostLlmBridge {
+  // Per-bridge, not module-global: each runtime owns its route knowledge, so
+  // concurrent bridges stay independent and tests remain hermetic.
+  const effortUnsupportedRoutes: EffortUnsupportedRoutes = new Set();
   return {
     id: HOST_LLM_BRIDGE_ID,
     async complete(input: HostLlmCompleteInput): Promise<HostLlmCompletion> {
@@ -122,6 +125,7 @@ export function createDeepSeekHarnessHostLlmBridge(
           input,
           route,
           callDeadline.signal,
+          effortUnsupportedRoutes,
         );
         const request = createGenerateOptions(input, route, prepared.config);
         request.signal = callDeadline.signal;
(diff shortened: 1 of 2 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2285
**Task:** `392e613ed310c57b`
**Base:** `main`
**Head:** `fix/host-llm-effort-probe-cache`

✅ **OpenCodeReview**: Review complete: 0 finding(s) across 1 selected item(s).

Memtensor-AI:

All tests passed (20/20 executed). memos_local_plugin/unit: 20/20. Duration: 3s [advisory, non-gating] AI-generated tests on branch test/auto-gen-392e613ed310c57b-20260826190007: 41/43 passed, 2 failed — these do NOT affect the PR verdict; review the branch manually.

**Branch:** `fix/host-llm-effort-probe-cache`

asyade:
Closing at the author's request — withdrawn for now.

Issue #2300, opened 2026-08-28
Serialized activation KV cache carries no model identity, so a cache built by one model loads silently into another


jeojdi1:
> **Re-checked 2026-08-28** against `main` at `185ebdb` ("Dev v2.0.32", #2295). `src/memos/memories/activation/kv.py` is unchanged since #797 (2025-12-30) and `_concat_caches` is byte-identical on `main` and `dev-v2.0.30`, so everything below still applies at HEAD. The line numbers cite `9119efe`, where it was measured.

### Environment

- MemOS at `main` / `9119efe5554e61a94b669df5eb84cc1b8ef3c0ab`
- `torch 2.13.0`, `transformers 5.15.1`, CPU, fp32
- Models: `Qwen/Qwen2.5-0.5B-Instruct` (writer) and `Qwen/Qwen2.5-0.5B` (reader) — same architecture
  and identical tensor shapes, different weights

### Summary

`KVCacheMemory.dump` writes `{"kv_cache_memories": self.kv_cache_memories}` and nothing else
(`src/memos/memories/activation/kv.py:195-198`). The only metadata attached to an item is
`{"source_text", "extracted_at"}` (`kv.py:47-50`), and `KVCacheItem`
(`src/memos/memories/activation/item.py:32-43`) has no field for the producing model. `KVCacheMemory.load`
(`kv.py:139-181`) unpickles and installs whatever it finds without checking anything about the
producer. The only identity check in the cube load path is a config *schema* version string
(`src/memos/mem_cube/general.py:60-66`).

A KV cache is only valid for the exact weights, tokenizer and dtype that produced it, and caches are
explicitly designed to move between machines — `GeneralMemCube.init_from_remote_repo` clones a cube
from `https://huggingface.co/datasets/<cube_id>` and loads it (`mem_cube/general.py:161-184`).
Two further details make a mismatch easy to hit without anyone doing anything wrong:

- `merge_config_with_default` rebuilds the config from the *caller's* default and preserves only
  `{"user_id", "cube_id", "config_filename", "model_schema"}` (`mem_cube/utils.py:148-153`), so the
  model recorded in a cube's own `config.json` is discarded on that path.
- `model_name_or_path` is not a weights identity: `HFLLM.__init__` loads with `torch_dtype="auto"`
  and no `revision` pin (`llms/hf.py:44-54`), and `HFLLMConfig` (`configs/llm.py:130-138`) has no
[...]

PR #2204, opened 2026-08-04
Fix #2203: skill-memory zip extractall is path-traversal (zip-slip); activation cache uses 

Memtensor-AI:
## Description

Hardens the two unauthenticated skill-memory sinks flagged in issue #2203.

**1. Zip-slip (CWE-22) in `_extract_and_parse_skill_zip`.** Added a private `_safe_extract_zip(zf, extract_dir)` helper next to the caller in `src/memos/mem_reader/read_skill_memory/upload_skill_memory.py`. Before extraction it walks `zf.infolist()` and rejects: absolute-path entries (POSIX or NT), `..` traversal entries whose resolved destination escapes `extract_dir`, and symlink entries (external_attr bit `0o120000`). Any hostile entry causes the entire extraction to abort with `ValueError` — no silent skip, because the codepath is unauthenticated and partial extraction would produce a subtly-wrong SkillMemory that hides the attack.

**2. Unsafe pickle.load (CWE-502) in `KVCacheMemory.load` and `VLLMKVCacheMemory.load`.** Added a nested `_SafeUnpickler(pickle.Unpickler)` per module that overrides `find_class` and enforces a per-module allowlist. `kv.py` allows `DynamicCache`/`DynamicLayer`, `KVCacheItem`, `KVCacheRecords`, stdlib containers, `datetime.*`, and the `torch.*` tensor rebuild helpers needed by DynamicCache. `vllmkv.py` allows `VLLMKVCacheItem`, `KVCacheRecords`, stdlib containers, `datetime.*` (no torch — the prompt is a `str`). Any other `find_class` call raises `pickle.UnpicklingError` before the class is imported, so a hostile `__reduce__` never fires. The existing except-block resets `kv_cache_memories = {}` on failure and logs a WARN so operators see the attempted attack.

**Tests + verification.** 14 new tests: 9 zip-slip cases (`../` / `../../../` / absolute-path / symlink / legit round trip / high-level parser guard / two platform-defensive edges) and 5 pickle cases (kv+vllm hostile-payload block, kv arbitrary-class block, kv+vllm legit round trip). TDD-verified: pre-fix `_OsSystemPayload` PoC actually wrote the canary file, confirming the sink was live; post-fix the canary is not created and the cache resets to empty. `ruff check` + `ruff format --check` pass on all changed files. Two pre-existing `test_kv.py` failures (`test_get_cache_merge`, `test_delete_and_get_all`) are transformers-API-version issues unrelated to this patch — reproducible on the base branch before the fix is applied.

[...]

Issue #2203, opened 2026-08-04
activation cache uses pickle.load (unsafe deserialization) — correction: the zip-slip vector does NOT reproduce


EvolveAegis:
## Correction (2026-08-04)

I need to correct this report. My original write-up claimed a zip-slip → pickle RCE chain. I re-tested the zip-slip leg locally and **it does not reproduce**: on Python 3.13, `zipfile.ZipFile.extractall` sanitizes traversal — entries with `..`, absolute paths, or `sub/../` all land inside the target directory and do not escape. So the zip-slip write primitive I described is not real, and the "RCE chain" framing is withdrawn.

What does hold:

## What still holds: pickle.load on the activation cache (unsafe deserialization, CWE-502)

- `src/memos/memories/activation/kv.py:158` and `src/memos/memories/activation/vllmkv.py:164` both do `data = pickle.load(f)` over cache files.
- A crafted pickle runs arbitrary code on load (confirmed: `pickle.load` of a `__reduce__`-based payload executed the canary).

I have **not** demonstrated a working write primitive that lets an attacker plant a malicious cache file on disk, so I am not claiming this is a proven RCE — it is an unsafe-deserialization sink that should be hardened regardless (treat the cache as untrusted input: use a non-executable format like JSON/safetensors, or load only from a trusted, integrity-checked location).

Apologies for the over-stated original report; I've edited it down to the verified sink.

## Suggested change
Replace `pickle.load` on the activation cache with a safe format (JSON/safetensors) or a safe-pickle variant, and/or load only from a trusted, non-writable location with an integrity check.

Memtensor-AI:
🤖 AutoDev has picked up this issue and started working on it.

**Task ID:** `5845caaa9f3b7735`
**Working branch:** `bugfix/autodev-2203-20260804015142516`
**Target branch:** `latest dev* branch`
**Workflow:** opsp (analysis → coding → testing → PR)

I will post the PR link here once done. If I need more information, I will ask in the comments.

EvolveAegis:
## Correction (2026-08-04)

I need to correct this report. My original write-up claimed a zip-slip → pickle RCE chain. I re-tested the zip-slip leg locally and **it does not reproduce**: on Python 3.13, `zipfile.ZipFile.extractall` sanitizes traversal — entries with `..`, absolute paths, or `sub/../` all land inside the target directory and do not escape. So the zip-slip write primitive I described is not real, and the "RCE chain" framing is withdrawn.

What does hold:

[...]

PR #2302, opened 2026-08-28
fix: record and check model identity on serialized KV caches

jeojdi1:
## Description

A KV cache is the internal activation state of one specific set of weights — it is not portable data. `KVCacheMemory.dump()` currently writes only `{"kv_cache_memories": ...}`, and `load()` restores it unconditionally. There is nothing recording which model produced a cache and nothing checking it on the way back in.

The consequence is silent: a cache dumped under one model and loaded under another is accepted with **no error**, and the model simply produces different tokens. On a close fine-tune pair I measured a next-token KL shift of **0.08–0.92** with the top-1 token flipping on **2 of 5** probes. A distant *architecture* does raise, but only as an opaque `RuntimeError` about tensor sizes, which does not tell the user what actually went wrong.

This PR:

* records `model_identity` in the dumped payload (`model_name_or_path`, best-effort from the configured extractor LLM);
* checks it on `load()` and **warns** on mismatch, naming both models and saying what to do about it.

It is a warning rather than an exception on purpose: caches dumped before this field existed carry no identity, and refusing to load them would break every existing store. `_model_identity()` returns `None` when the LLM config exposes no name, so a dump never fails because identity could not be determined, and a `None` on either side skips the check.

Related Issue (Required): #2300

**Note on overlap:** issue #2203 (closed) and PR #2204 cover `pickle.load` in this same `load()` path as an unsafe-deserialization sink (CWE-502). This change is orthogonal — it adds a payload field and a check, and does not touch how the payload is deserialized — but it does edit adjacent lines, so it may need a trivial rebase depending on merge order.

## Type of change

## How Has This Been Tested?

- [x] Unit Test

Three tests added:

* `test_dump_records_model_identity` — the field is written.
* `test_load_warns_on_model_mismatch` — dump under `org/model-a`, load under `org/model-b`, assert the warning names the original model. Fails on the current code, which emits nothing.
* `test_load_is_quiet_when_model_matches` — no warning on the matching case, so this is not a new source of noise.

## Checklist

- [ ] I have created related documentation issue/PR in MemOS-Docs (happy to, if you consider the new payload field user-facing)
- [x] I have linked the issue to this PR (#2300)

### DIFF
--- a/src/memos/memories/activation/kv.py
+++ b/src/memos/memories/activation/kv.py
@@ -6,11 +6,15 @@ from datetime import datetime
 from transformers import DynamicCache
 
 from memos.configs.memory import KVCacheMemoryConfig
+from memos.log import get_logger
 from memos.dependency import require_python_package
 from memos.llms.factory import LLMFactory
 from memos.memories.activation.base import BaseActMemory
 from memos.memories.activation.item import KVCacheItem
 from memos.memories.textual.item import TextualMemoryItem
+
+
+logger = get_logger(__name__)
 
 
 class KVCacheMemory(BaseActMemory):
@@ -158,6 +162,7 @@ class KVCacheMemory(BaseActMemory):
                 data = pickle.load(f)
 
             if isinstance(data, dict):
+                self._check_model_identity(data.get("model_identity"))
                 # Load memories, handle both old and new formats
                 if "kv_cache_memories" in data:
                     memories = data["kv_cache_memories"]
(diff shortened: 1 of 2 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2302
**Task:** `022280e030428671`
**Base:** `main`
**Head:** `fix/kv-cache-carries-no-model-identity`

---

### 1. `src/memos/memories/activation/kv.py` (L238-L241)

The `saved` parameter is typed `dict | None` but is sourced directly from `data.get('model_identity')` after a `pickle.load`, so its actual runtime type is unconstrained. If the stored value is anything other than `dict` or `None` (e.g., a plain string `"gpt2"` from a corrupted or hand-edited file), the guard `saved is None` is `False`, execution falls through, and `saved.get(...)` raises `AttributeError`, crashing the entire `load` path with an unhelpful traceback — the opposite of the intended graceful degradation.

Replace the guard with an `isinstance` check:

**Before:**
```
        current = self._model_identity()
        if saved is None or current is None:
            return
        if saved.get("model_name_or_path") != current.get("model_name_or_path"):
```

**After:**
```suggestion
        current = self._model_identity()
        if not isinstance(saved, dict) or current is None:
            return
        if saved.get("model_name_or_path") != current.get("model_name_or_path"):
```

</details>

---

### 2. `src/memos/memories/activation/kv.py` (L220-L226)

When none of the three hardcoded attribute names matches `cfg`, the method returns `None`. `_check_model_identity` then hits `if saved is None or current is None: return` and silently skips the entire identity check — including for mismatched models. This means the mismatch warning is permanently suppressed for any LLM config that exposes its model path under a different attribute name (e.g. `model_id`, `checkpoint`), with no log message or other indication that the check was skipped. Consider logging a debug/warning when `cfg` is non-None but no known attribute was found, so that gaps in coverage are discoverable.

---

### 3. `tests/memories/activation/test_kv.py` (L139)

The negative assertion filters on the literal substring `"was built with model"`, while the positive counterpart (`test_load_warns_on_model_mismatch`) uses a completely different predicate — checking that both model name strings appear in the message. These two tests are supposed to be inverses of each other, but their predicates are not symmetric.

The production message is `"KV cache was built with model %r but is being loaded into %r"`. If that phrasing ever changes (e.g., to `"KV cache produced by model %r"`), this filter would match zero records regardless of how many warnings were actually emitted, making the assertion pass vacuously. `test_load_warns_on_model_mismatch` would fail correctly on the same change — but this test would silently give false confidence.

Suggestion: mirror the predicate from the positive test so both break together on a format change:
```python
assert not any("org/model-a" in r.getMessage() for r in caplog.records)
```

Memtensor-AI:

**Details:** Test collection failed because the `torch` module is not installed in the test environment, preventing the test module from being imported at all. [advisory, non-gating] AI-generated tests on branch test/auto-gen-b6472f335cd88b41-20260829014814: 82/82 passed — these do NOT affect the PR verdict; review the branch manually.
**Branch:** `fix/kv-cache-carries-no-model-identity`

jeojdi1:
Thanks for the review — taking one of these and pushing back on the other.

**1. `import pickle` at module level** — agreed, fixed in 56fe586.

**2. `hasattr(merged, "layers")` is dead code** — I think this one is backwards, and applying the suggestion would break the test.

`DynamicCache` does expose `.layers`, and `key_cache` is the attribute that no longer exists. In `transformers` v5.16.1, `src/transformers/cache_utils.py`:

```python
1306:  self.layers = layers if layers is not None else []
1322:  return len(self.layers)          # __len__
1375:  while len(self.layers) <= layer_idx:
1383:  keys, values = self.layers[layer_idx].update(...)
```

and `key_cache` appears **zero** times in that file. The rename landed in 4.57; `.key_cache` / `.value_cache` were the pre-4.57 layout. So the suggested `assert len(merged.key_cache) == 1` raises `AttributeError` on any current install — which is exactly the failure mode this PR series is fixing, and the reason `test_get_cache_merge` and `test_delete_and_get_all` currently fail on `main` (see #2313).

The `hasattr` branch is there deliberately because `pyproject.toml:41` declares `transformers >=4.51.3,<5.0.0`, which spans both layouts, so the test has to work on either.

That said — your parenthetical suggestion of `len(merged) == 1` is strictly better than the `hasattr` branch: `__len__` returns `len(self.layers)` and is stable across both layouts. Happy to push that if a maintainer prefers it.

Memtensor-AI:

**Details:** Test collection failed because the `torch` module is not installed in the test environment, preventing the test file from being imported. [advisory, non-gating] AI-generated tests on branch test/auto-gen-022280e030428671-20260901164722: 0/54 passed, 54 failed — these do NOT affect the PR verdict; review the branch manually.
**Branch:** `fix/kv-cache-carries-no-model-identity`

Issue #2301, opened 2026-08-28
The stored activation DynamicCache is mutated in place by generation, so activation memory grows every turn


jeojdi1:
> **Re-checked 2026-08-28** against `main` at `185ebdb` ("Dev v2.0.32", #2295). `src/memos/memories/activation/kv.py` is unchanged since #797 (2025-12-30) and `_concat_caches` is byte-identical on `main` and `dev-v2.0.30`, so everything below still applies at HEAD. The line numbers cite `9119efe`, where it was measured.

### Environment

MemOS at `main` / `9119efe`; `torch 2.13.0`, `transformers 5.15.1`, CPU, `Qwen/Qwen2.5-0.5B-Instruct`.

### Summary

`mem_os/core.py:324-326` takes the stored cache object and passes it straight through:

```python
kv_cache = next(iter(mem_cube.act_mem.get_all()), None)
past_key_values = kv_cache.memory if (kv_cache and hasattr(kv_cache, "memory")) else None
...
response = self.chat_llm.generate(current_messages, past_key_values=past_key_values)
```

`HFLLM._prefill` (`llms/hf.py:305-323`) hands that object to `self.model(...)`, and
`DynamicLayer.update` appends in place (`self.keys = torch.cat([self.keys, key_states], dim=-2)`).
The stored `KVCacheItem.memory` therefore accumulates the prompt and the generated tokens of every
turn. `_concat_caches` returning `caches[0]` unchanged for a single id (`kv.py:208-209`) hands out the
same aliased object, and `ActivationMemoryManager` re-dumps it
(`mem_scheduler/memory_manage_modules/activation_memory_manager.py:95`), so the growth persists to
disk.

### Minimal reproduction

```python
import torch
from transformers import AutoModelForCausalLM, AutoTokenizer
M = "Qwen/Qwen2.5-0.5B-Instruct"
tok = AutoTokenizer.from_pretrained(M)
m = AutoModelForCausalLM.from_pretrained(M, dtype=torch.float32).eval()
ids = tok("The user lives in Berlin.", return_tensors="pt").input_ids
with torch.no_grad():
    stored = m(input_ids=ids, use_cache=True, return_dict=True).past_key_values
print("after build:", stored.get_seq_length())
for turn in range(1, 4):
    q = tok(f"\nQ{turn}: tell me something.", return_tensors="pt",
            add_special_tokens=False).input_ids
    with torch.no_grad():
[...]

PR #2303, opened 2026-08-28
fix: stop get_cache handing out the stored KV cache by reference

jeojdi1:
## Description

`_concat_caches` returns `caches[0]` unchanged when a single cache id is requested, so `get_cache()` hands the caller the stored `KVCacheItem.memory` **object itself**. The caller passes that cache to `generate()`, which appends to it in place — so the saved activation memory grows on every chat turn.

Measured across three turns, the stored cache length went **6 → 19 → 32 → 45**. Nothing in the API suggests that *retrieving* a memory mutates it, and the multi-cache path already builds a fresh container, so only this early return leaks the reference.

This PR returns a new `DynamicCache` sharing the same tensors. The tensors are deliberately **not** cloned: `generate()` appends along the sequence axis rather than writing into existing rows, so a fresh container is sufficient to protect the stored item without paying to duplicate the cache.

Related Issue (Required): #2301

## Type of change

## How Has This Been Tested?

- [x] Unit Test

`test_get_cache_does_not_alias_stored_memory` asserts the returned object is not the stored one, then simulates `generate()` by appending a step to the returned cache and asserts the stored length is unchanged.

Verified fails-before / passes-after on `transformers 5.16.1`:

```
# upstream
FAILED tests/memories/activation/test_kv.py::test_get_cache_does_not_alias_stored_memory
1 failed

# with this change
1 passed
```

## Checklist

- [ ] I have created related documentation issue/PR in MemOS-Docs (not applicable)
- [x] I have linked the issue to this PR (#2301)

### DIFF
--- a/src/memos/memories/activation/kv.py
+++ b/src/memos/memories/activation/kv.py
@@ -80,6 +80,42 @@ class KVCacheMemory(BaseActMemory):
 
         return self._concat_caches(caches_to_merge)
 
+    @staticmethod
+    def _copy_cache(cache: DynamicCache) -> DynamicCache:
+        """Shallow-structure copy of a DynamicCache that shares no mutable state.
+
+        The tensors themselves are not cloned -- ``generate`` appends along the
+        sequence axis rather than writing into the existing rows, so a fresh
+        container with the same tensor objects is enough to protect the stored
+        item, without paying to duplicate the cache.
+        """
+        import torch  # noqa: F401  (kept local, matching this module's style)
+
+        copy = DynamicCache()
+        if hasattr(cache, "layers"):
+            if not hasattr(copy, "layers"):
+                copy.layers = []
+            layer_cls = type(cache.layers[0]) if cache.layers else None
+            while layer_cls is not None and len(copy.layers) < len(cache.layers):
+                copy.layers.append(layer_cls())
+            for i, src in enumerate(cache.layers):
+                dst = copy.layers[i]
+                if not getattr(dst, "is_initialized", True) and hasattr(
+                    dst, "lazy_initialization"
+                ):
+                    dst.lazy_initialization(src.keys, src.values)
+                dst.keys = src.keys
+                dst.values = src.values
+        elif hasattr(cache, "key_cache"):
+            for i in range(len(cache.key_cache)):
+                copy.key_cache.append(cache.key_cache[i])
+                copy.value_cache.append(cache.value_cache[i])
+        else:
+            raise AttributeError(
+                "DynamicCache object has neither 'layers' nor 'key_cache' attributes"
+            )
+        return copy
+
     def get(self, memory_id: str) -> KVCacheItem | None:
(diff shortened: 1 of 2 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2303
**Task:** `110bb783284bc7e7`
**Base:** `main`
**Head:** `fix/kv-cache-mutated-in-place-by-generate`

---

### 1. `src/memos/memories/activation/kv.py` (L92)

`copy` shadows the standard-library module of the same name. While harmless here since `copy.copy`/`copy.deepcopy` are not used inside this function, it can confuse readers and static analysis. Prefer a name like `new_cache` or `dst_cache`.

**Before:**
```
        copy = DynamicCache()
```

**After:**
```suggestion
        new_cache = DynamicCache()
```

</details>

Memtensor-AI:

**Details:** Test collection failed because the `torch` module is not installed in the test environment, preventing the test file from being imported. [advisory, non-gating] AI-generated tests on branch test/auto-gen-c9892125fee30ed9-20260829013054: 36/36 passed — these do NOT affect the PR verdict; review the branch manually.
**Branch:** `fix/kv-cache-mutated-in-place-by-generate`

jeojdi1:
Thanks for the review — one accepted, one I'd push back on.

**1. Unused `import torch`** — correct, removed in a8b1bbf.

**2. The conditional `lazy_initialization` call** — this one needs to stay, and removing it would reintroduce the bug in #2313.

`lazy_initialization` does more than set `keys` / `values`. From `transformers` v5.16.1 `cache_utils.py:121-125`:

```python
def lazy_initialization(self, key_states, value_states) -> None:
    self.dtype, self.device = key_states.dtype, key_states.device
    self.keys = torch.tensor([], dtype=self.dtype, device=self.device)
    self.values = torch.tensor([], dtype=self.dtype, device=self.device)
    self.is_initialized = True
```

The load-bearing line is the last one. `DynamicLayer.get_seq_length()` at `:154-158` reads:

```python
    return 0
```

So a layer that gets `.keys` / `.values` assigned without `is_initialized` being set reports length 0 and has its contents discarded on the first forward pass. That is precisely the defect in #2313 / #2314. It also sets `dtype` and `device`, and `DynamicSlidingWindowLayer` overrides it to move `_sliding_window_tensor` onto the right device (`:225-227`).

On the second option — an `else` branch so the assignments only run when `lazy_initialization` did not — that would be worse: `lazy_initialization` sets `keys` / `values` to *empty* tensors, so the copy would come out empty. The call has to run first and the assignments have to overwrite it, which is what the current code does.

I've left the ordering as-is, but it clearly isn't obvious from reading it — happy to add a comment making the dependency explicit if you'd like.

Memtensor-AI:

**Details:** Test collection failed because the `torch` module is not installed in the test environment, preventing the test file from being imported.
**Branch:** `fix/kv-cache-mutated-in-place-by-generate`

PR #2305, opened 2026-08-28
Fix #2300: Serialized activation KV cache carries no model identity, so a cache built by on

### LINKED ISSUE #2300: Serialized activation KV cache carries no model identity, so a cache built by one model loads silently into another

Memtensor-AI:
## Description

Fixes issue #2300 — serialized activation KV caches now carry a producer fingerprint that is verified at load time, replacing the silent cross-model swap the issue reproduced with a clear diagnostic.

**What changed.** `KVCacheMemory.extract` / `from_textual_memory` and their `VLLMKVCacheMemory` counterparts now stamp `metadata["producer"]` with `model_name_or_path`, `backend`, `config_fingerprint` (sha256 of `model.config.to_diff_dict()`), `tokenizer_fingerprint` (sha256 of `backend_tokenizer.to_str()`), `torch_dtype`, `quantization`, plus soft hints (`architectures`, `num_hidden_layers`, `num_kv_heads`, `head_dim`, `transformers_version`). On load, every item's saved fingerprint is compared against a freshly-computed live one across the strict subset `{model_name_or_path, config_fingerprint, tokenizer_fingerprint, torch_dtype, quantization}`. Mismatched items are dropped with a single `logger.error` line naming every offending field and the item id; other items keep loading. Items with no `producer` key (pre-fix caches) are kept with a `logger.warning` — the deprecation window the issue requested. The silent `except (EOFError, pickle.UnpicklingError, Exception): self.kv_cache_memories = {}` at the load site now emits `logger.warning(..., exc_info=True)` so a corrupt cache is distinguishable from an empty one in production (issue's second suggestion).

**Backward compatibility.** Fully backward compatible — no config schema change, no on-disk file layout change (still a pickled dict at `os.path.join(dir, memory_filename)`); metadata simply gains one additive sub-key per item.

**Tests.** 14 new pytest cases under `tests/memories/activation/test_kv.py` and new `tests/memories/activation/test_vllmkv.py` cover: fingerprint capture on extract / from_textual_memory / on a partially-broken introspection source; matched load stays silent; config / tokenizer / dtype mismatch each drops the item with an ERROR line; mixed batch keeps the good item and drops only the bad one; missing-producer pre-fix cache installs with WARNING; corrupt pickle logs WARNING and resets to empty. Two pre-existing tests (`test_get_cache_merge`, `test_delete_and_get_all`) were failing on `main` against `transformers 4.57.6` because they used the removed `DynamicCache.key_cache` API; rewritten to use `DynamicCache.update`. Full activation module = 25/25 pass; broader `tests/memories/` + `tests/llms/` = 125/125 pass; no regressions attributable to this change. Ruff clean.

Related Issue (Required):  Fixes #2300

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Automated tests are pending.

- [ ] Unit Test

## Checklist

## Reviewer Checklist
- [x] closes #2300
- [ ] Made sure Checks passed

### DIFF
--- a/src/memos/memories/activation/kv.py
+++ b/src/memos/memories/activation/kv.py
@@ -1,16 +1,143 @@
+import hashlib
+import json
 import os
 import pickle
 
 from datetime import datetime
+from typing import Any
 
 from transformers import DynamicCache
 
 from memos.configs.memory import KVCacheMemoryConfig
 from memos.dependency import require_python_package
 from memos.llms.factory import LLMFactory
+from memos.log import get_logger
 from memos.memories.activation.base import BaseActMemory
 from memos.memories.activation.item import KVCacheItem
 from memos.memories.textual.item import TextualMemoryItem
+
+
+logger = get_logger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Producer fingerprint helpers (see openspec change 2300-... for the contract)
+# ---------------------------------------------------------------------------
+
+# The strict subset of fingerprint fields we compare on load. A field mismatches
+# only if both sides are non-None and unequal — unknown fields never fail.
+_STRICT_FIELDS: tuple[str, ...] = (
+    "model_name_or_path",
+    "config_fingerprint",
+    "tokenizer_fingerprint",
+    "torch_dtype",
+    "quantization",
+)
+
+
+def _compute_producer_fingerprint(llm: Any) -> dict[str, Any]:
+    """Return an identity fingerprint for the LLM that produced a KV cache.
+
+    Every field is best-effort — a missing attribute becomes ``None`` rather
+    than an exception. The loader treats ``None`` as unknown, not as mismatch,
+    so a partial fingerprint stays useful.
+    """
+    fp: dict[str, Any] = {
+        "model_name_or_path": None,
+        "backend": None,
+        "config_fingerprint": None,
+        "tokenizer_fingerprint": None,
+        "torch_dtype": None,
+        "quantization": None,
+        "architectures": None,
+        "num_hidden_layers": None,
+        "num_kv_heads": None,
+        "head_dim": None,
+        "transformers_version": None,
+    }
+
+    llm_config = getattr(llm, "config", None)
+    if llm_config is not None:
+        fp["model_name_or_path"] = getattr(llm_config, "model_name_or_path", None)
+        fp["backend"] = type(llm_config).__name__
+
+    model = getattr(llm, "model", None)
+    if model is not None:
+        try:
+            model_config = getattr(model, "config", None)
+            if model_config is not None:
+                to_diff_dict = getattr(model_config, "to_diff_dict", None)
+                if callable(to_diff_dict):
+                    diff = to_diff_dict()
+                    fp["config_fingerprint"] = hashlib.sha256(
+                        json.dumps(diff, sort_keys=True, default=str).encode()
+                    ).hexdigest()
+                quant = getattr(model_config, "quantization_config", None)
+                fp["quantization"] = str(quant) if quant is not None else None
+                arch = getattr(model_config, "architectures", None)
+                if arch:
+                    fp["architectures"] = list(arch)
+                fp["num_hidden_layers"] = getattr(model_config, "num_hidden_layers", None)
+                fp["num_kv_heads"] = getattr(model_config, "num_key_value_heads", None) or getattr(
+                    model_config, "num_attention_heads", None
+                )
+                head_dim = getattr(model_config, "head_dim", None)
+                if head_dim is None:
+                    hidden = getattr(model_config, "hidden_size", None)
+                    n_heads = getattr(model_config, "num_attention_heads", None)
+                    if hidden and n_heads:
+                        head_dim = hidden // n_heads
+                fp["head_dim"] = head_dim
+            dtype = getattr(model, "dtype", None)
+            if dtype is not None:
+                fp["torch_dtype"] = str(dtype)
+        except Exception:
+            logger.warning(
+                "Failed to introspect HF model config for KV cache fingerprint",
+                exc_info=True,
+            )
+
+    tokenizer = getattr(llm, "tokenizer", None)
+    if tokenizer is not None:
+        try:
+            backend_tokenizer = getattr(tokenizer, "backend_tokenizer", None)
+            if backend_tokenizer is not None and hasattr(backend_tokenizer, "to_str"):
+                fp["tokenizer_fingerprint"] = hashlib.sha256(
+                    backend_tokenizer.to_str().encode()
+                ).hexdigest()
+        except Exception:
+            logger.warning(
+                "Failed to compute tokenizer fingerprint for KV cache",
+                exc_info=True,
+            )
+
+    try:
+        import transformers
+
+        fp["transformers_version"] = transformers.__version__
+    except Exception:
+        pass
+
+    return fp
+
+
+def _fingerprint_mismatch_reasons(saved: dict[str, Any] | None, live: dict[str, Any]) -> list[str]:
+    """Return a list of human-readable mismatch reasons.
+
+    Empty list means "safe to install" (either every field matched or one side
+    was unknown). Only strict fields participate in the decision.
+    """
+    if not isinstance(saved, dict):
+        return []
+    reasons: list[str] = []
+    for field in _STRICT_FIELDS:
+        s_val = saved.get(field)
+        l_val = live.get(field)
+        if s_val is None or l_val is None:
+            continue
+        if s_val != l_val:
+            reasons.append(f"{field}: saved={s_val!r} live={l_val!r}")
+    return reasons
 
 
 class KVCacheMemory(BaseActMemory):
@@ -29,6 +156,10 @@ class KVCacheMemory(BaseActMemory):
         self.llm = LLMFactory.from_config(config.extractor_llm)
         self.kv_cache_memories: dict[str, KVCacheItem] = {}
 
+    def _producer_fingerprint(self) -> dict[str, Any]:
+        """Fingerprint of the LLM currently wired into this memory."""
+        return _compute_producer_fingerprint(self.llm)
+
     def extract(self, text: str) -> KVCacheItem:
         """Extract memory based on the text.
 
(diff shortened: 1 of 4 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2305
**Task:** `b7c23a4c04476531`
**Base:** `dev-v2.0.30`
**Head:** `bugfix/autodev-2300-20260828172611164`
**Head SHA:** `2be4dbb4067351752e7fdafd4df64bb2ca7910b6`

---

### 1. `src/memos/memories/activation/kv.py` (L9)

Finding #1 is still unresolved. `DynamicCache` does not expose a `.layers` property — the public API surfaces per-layer data via `.key_cache` and `.value_cache` (plain Python lists). Accessing `.layers` on a `DynamicCache` instance will raise `AttributeError` at runtime.

Fix: replace with `assert len(merged.key_cache) == 1` (and optionally `assert len(merged.value_cache) == 1`).

**Before:**
```
from transformers import DynamicCache
```

**After:**
```suggestion
    assert len(merged.key_cache) == 1
    assert len(merged.value_cache) == 1
```

</details>

---

### 2. `src/memos/memories/activation/kv.py` (L314-L318)

Finding #6 is still unresolved in kv.py. A fingerprint mismatch is an expected operational event (model changed, cache is stale) rather than an error condition. Logging it at `ERROR` level will trigger false-alarm alerts in production monitoring. The equivalent path in `vllmkv.py` was correctly fixed to use `logger.warning`, but `kv.py` still uses `logger.error`.

Fix: change `logger.error` to `logger.warning` here, consistent with the vllmkv.py fix.

**Before:**
```
logger.error(
    "KV cache item %s dropped: producer fingerprint mismatch (%s)",
    item_id,
    "; ".join(reasons),
)
```

**After:**
```suggestion
                logger.warning(
                    "KV cache item %s dropped: producer fingerprint mismatch (%s)",
                    item_id,
                    ";".join(reasons),
                )
```

</details>

---

### 3. `tests/memories/activation/test_kv.py` (L274)

**Finding #1 is NOT fixed.** `DynamicCache` (transformers `>=4.51.3`, as pinned in `pyproject.toml`) does not expose a `.layers` property. The public, stable API uses `.key_cache` and `.value_cache` (both plain Python `list`). Calling `merged.layers` will raise `AttributeError` at runtime, making this test fail unconditionally.

The previous code `assert len(merged.key_cache) == 1` and `assert len(merged.value_cache) == 1` was correct — the fix replaced it with an invalid attribute access instead of fixing the assertion logic.

**Before:**
```
    assert len(merged.layers) == 1
```

**After:**
```suggestion
    assert len(merged.key_cache) == 1
    assert len(merged.value_cache) == 1
```

</details>

---

Memtensor-AI:

Open Code Review found **9** issue(s). I have resumed the development Agent to fix them.

- Task: `b7c23a4c04476531`
- Fix attempt: `1/2`
- Finding delta: `0 repeated / 9 new / 0 likely resolved`

Memtensor-AI:

**Details:** Test collection failed because the `torch` module is not installed in the test environment, preventing the test file from being imported. [advisory, non-gating] AI-generated tests on branch test/auto-gen-b7c23a4c04476531-20260829020727: 92/100 passed, 8 failed — these do NOT affect the PR verdict; review the branch manually.
**Branch:** `bugfix/autodev-2300-20260828172611164`

PR #2310, opened 2026-08-31
fix(retrieval): order the tier-2 vector scan window by recency, not physical position

### LINKED ISSUE #2233: fix: Tier-2 vector recall applies unordered LIMIT before cosine ranking, causing false negatives
## Pre-submission checklist | 提交前检查

## Bug Description | 问题描述

Tier-2 vector recall applies an unordered SQL `LIMIT` before cosine ranking. As a result, the globally best vector can be excluded from similarity computation solely because it appears after the physical scan prefix.

This is not a claim that every retrieval must scan the full table. A bounded scan is a reasonable performance safeguard. The correctness problem is that the bounded candidate set is not selected by an ANN index, vector relevance, deterministic recency order, or another semantic pre-filter: `scanAndTopK()` executes a query of the following shape:

```sql
SELECT ... FROM traces WHERE ... LIMIT <hardCap>
```

There is no `ORDER BY`. `runTier2()` then passes `hardCap: vecPoolSize * 4`. Therefore records outside the arbitrary SQLite prefix have zero probability of vector recall, even if their cosine similarity is `1.0`.

With the v2.0.14 defaults, `tier2TopK=5` and `candidatePoolFactor=4`, so `vecPoolSize=20` and the effective vector scan cap is only 80 qualifying trace rows per vector channel.

Expected behavior: a Tier-2 vector Top-K search should either select candidates using a relevance-aware strategy, or explicitly expose/document approximate bounded-search semantics with a deterministic, meaningful candidate policy. It should not silently treat an unordered SQL prefix as the vector search corpus.

Actual behavior: the best vector is deterministically missed when it is immediately outside the prefix. Increasing `tier2TopK` by one changes the cap and suddenly makes the same vector rank first, although neither the data nor query changed.

## How to Reproduce | 如何重现

The reproduction directly imports the compiled implementation from the official `@memtensor/memos-local-plugin@2.0.14` npm tarball.

1. Create an in-memory SQLite `traces` table with 400 qualifying rows and the same priority/timestamp.
2. Create an index on `priority` so the chosen SQL prefix is observable and reproducible.
3. Store vector `[1, 0]` at row 335. Use query vector `[1, 0]`, so its cosine is exactly `1.0`.
4. Store vectors with cosine `0.3` in every other row. This remains above the v2.0.14 `minTraceSim=0.25` default.
5. Call the official `scanAndTopK()` with `hardCap=320`, then with `hardCap=400`.
6. Call the official `runTier2()` with `tier2TopK=20`, then with `tier2TopK=21`, keeping every other input/config value unchanged.

Observed output:

```text
scanAndTopK(hardCap=320): trace-335 absent; best returned cosine ~= 0.3
scanAndTopK(hardCap=400): trace-335 is rank 1; cosine = 1.0

runTier2(tier2TopK=20): observed hardCap=320; trace-335 absent
runTier2(tier2TopK=21): observed hardCap=336; trace-335 present at rank 1
```

The corresponding query plan is:

```text
SEARCH traces USING INDEX idx_traces_priority (priority>?)
```

This boundary test isolates the cause:

- The target vector is valid and is the unique global best match.
- It satisfies the same SQL filters and similarity threshold as the other rows.
- The target becomes rank 1 as soon as the scan boundary includes it.
- Therefore the miss happens before cosine Top-K ranking, not in embedding generation, thresholding, hydration, or later re-ranking.

The same mechanism was independently reproduced against the official v2.0.10 package as well.

## Environment | 环境信息

- Package: `@memtensor/memos-local-plugin@2.0.14`
- Official npm tarball SHA-1: `32639d241918c7da8d536e52eac7e0a7c42c312e`
- Tarball SHA-256: `575af9121f86ad3eacfc6aca9d8a3e3e0856b2caa91a7d537a3c37ad3ee43907`
- npm integrity: `sha512-yEAroCSBfdf7urP47Hyr2MzTg4BPLIWqlno5r0imHb69s8fh7uXZRuPK23IWCzDFIWuPK/SuZfk8u3MdGQOzLg==`
- Node.js: v22.23.0
- Database driver: `better-sqlite3`
- Operating system: Linux, aarch64 container

## Additional Context | 其他信息

This may have been introduced or made more visible by the safeguards added after reports of synchronous full-table vector scans causing high CPU/RSS. The performance motivation is valid, but the current cap trades away recall based on physical/query-plan order rather than an explicit approximation strategy.

Possible directions, without prescribing a particular fix:

- use an ANN/vector index;
- preselect candidates using a documented semantic or temporal policy before cosine ranking;
- make the scan bound configurable and observable, with explicit approximate-recall semantics;
- avoid an unordered `LIMIT` as the sole candidate generator.

## Willingness to Implement | 实现意愿

- [x] I would like someone else to implement this

simpleqt:
Fixes #2233

## Problem

`scanAndTopK` bounded the cosine candidate set with a bare `LIMIT <hardCap>` — no `ORDER BY` — so the window was SQLite's arbitrary physical scan order. The globally best vector could be excluded from similarity computation purely by physical position (cosine 1.0, zero recall probability). With the v2.0.14 defaults (`tier2TopK=5`, `candidatePoolFactor=4`) the effective vector scan cap was ~80–100 arbitrary rows per channel.

While writing the regression test I also confirmed the arbitrariness empirically: the same query over the same rows inserted in a different physical order returns a different winner under the unordered cap.

## Fix

- `scanAndTopK` takes an optional `orderBy` (applied before the `hardCap` LIMIT); the option's contract is repo-internal constant strings, never request input.
- All four vector repos pass their recency column: `traces` → `ts DESC, id DESC`; `policies` / `world_model` / `skills` → `updated_at DESC, id DESC`. Every one of these is already served by an existing `DESC` index (`idx_traces_owner/share/session_ts`, `idx_policies_owner/status/...`, `idx_skills_owner`), so the bounded window stays index-friendly.
- The bounded-scan semantics are documented on `scanAndTopK`: the window is now a deterministic candidate policy (most recent qualifying rows first) rather than a physical accident.

This is the "deterministic, meaningful candidate policy" alternative the issue lists — it does not change the bounded-scan approximation itself, it makes the bound well-defined.

## Testing

New case in `tests/unit/storage/vector-stream.test.ts`: two perfect-match vectors (one old, one new) plus orthogonal fillers; with `orderBy: "ts DESC, id DESC"` and `hardCap: 1`, the window deterministically resolves to the newest perfect match **in both physical row layouts** (insertion order and reversed). The unordered path is deliberately left un-asserted — which row it returns is layout-dependent, which is exactly the bug.

- `vitest run tests/unit` → 1550 passed, 2 skipped (179 files)
- `tsc -p tsconfig.json --noEmit` → clean

### DIFF
--- a/apps/memos-local-plugin/core/storage/repos/policies.ts
+++ b/apps/memos-local-plugin/core/storage/repos/policies.ts
@@ -185,6 +185,7 @@ export function makePoliciesRepo(db: StorageDb) {
           where: whereParts.join(" AND "),
           params,
           hardCap: opts.hardCap,
+          orderBy: "updated_at DESC, id DESC",
         },
       );
     },
--- a/apps/memos-local-plugin/core/storage/repos/skills.ts
+++ b/apps/memos-local-plugin/core/storage/repos/skills.ts
@@ -223,6 +223,7 @@ export function makeSkillsRepo(db: StorageDb) {
           where: whereParts.join(" AND "),
           params,
           hardCap: opts.hardCap,
+          orderBy: "updated_at DESC, id DESC",
         },
       );
     },
--- a/apps/memos-local-plugin/core/storage/repos/traces.ts
+++ b/apps/memos-local-plugin/core/storage/repos/traces.ts
@@ -552,6 +552,7 @@ export function makeTracesRepo(db: StorageDb) {
           where: whereParts.join(" AND "),
           params,
           hardCap: opts.hardCap,
+          orderBy: "ts DESC, id DESC",
         },
       );
     },
--- a/apps/memos-local-plugin/core/storage/repos/traces.ts
+++ b/apps/memos-local-plugin/core/storage/repos/traces.ts
@@ -552,7 +552,7 @@ export function makeTracesRepo(db: StorageDb) {
           where: whereParts.join(" AND "),
           params,
           hardCap: opts.hardCap,
-          orderBy: "ts DESC, id DESC",
+          orderBy: "ts DESC, id DESC", // repo-internal constant; validated in scanAndTopK
         },
       );
     },
--- a/apps/memos-local-plugin/core/storage/repos/world_model.ts
+++ b/apps/memos-local-plugin/core/storage/repos/world_model.ts
@@ -172,6 +172,7 @@ export function makeWorldModelRepo(db: StorageDb) {
         vecColumn: "vec",
         where,
         hardCap: opts.hardCap,
+        orderBy: "updated_at DESC, id DESC",
       });
     },
 
--- a/apps/memos-local-plugin/core/storage/vector.ts
+++ b/apps/memos-local-plugin/core/storage/vector.ts
@@ -218,6 +218,15 @@ export interface ScanRow {
 export const DEFAULT_SCAN_HARD_CAP = 5_000;
 
 /**
+ * `orderBy` is interpolated into SQL, so it must be a repo-internal constant.
+ * This allowlist (identifier[, identifier]… each optionally ASC/DESC) rejects
+ * anything else at the boundary, so a future caller passing request-derived
+ * input fails loudly instead of opening SQL injection.
+ */
+const SAFE_ORDER_BY_RE =
+  /^[a-z_][a-z0-9_]*(\s+(asc|desc))?(,\s*[a-z_][a-z0-9_]*(\s+(asc|desc))?)*$/i;
+
+/**
  * Stream rows from `table`, decode vectors, and run top-K cosine against
  * `query`. `selectExtra` lets callers bring along columns that will surface in
  * `VectorHit.meta`.
@@ -245,6 +254,9 @@ export function scanAndTopK<TMeta = undefined>(
   if (k <= 0 || query.length === 0) return [];
 
   const { vecColumn, norm2Column, where, params, hardCap, orderBy } = opts;
+  if (orderBy !== undefined && !SAFE_ORDER_BY_RE.test(orderBy)) {
+    throw new Error(`scanAndTopK: unsafe orderBy value: ${JSON.stringify(orderBy)}`);
+  }
   const cap = hardCap ?? DEFAULT_SCAN_HARD_CAP;
   const cols = ["id", vecColumn, ...(norm2Column ? [norm2Column] : []), ...selectExtra];
   const sql = [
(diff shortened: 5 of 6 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2310
**Task:** `d7dd1c5d708f6868`
**Base:** `main`
**Head:** `fix/tier2-recency-ordered-scan`

---

### 1. `apps/memos-local-plugin/core/storage/vector.ts` (L226-L227)

The allowlist regex rejects most injection attempts, but `\s+` matches any whitespace including newlines and tabs. A value like `"updated_at\nDESC"` would pass the regex, and while SQLite normalises embedded newlines harmlessly today, it is surprising behaviour and a fragile assumption. Tighten to `[ ]+` (a plain space or `\x20+`) so only a single-space separator is valid — this matches all current caller strings and closes the whitespace ambiguity.

Suggestion:
  /^[a-z_][a-z0-9_]*( +(asc|desc))?(, *[a-z_][a-z0-9_]*( +(asc|desc))?)*$/i

---

### 2. `apps/memos-local-plugin/core/storage/vector.ts` (L257-L258)

The error message includes the raw `orderBy` value via `JSON.stringify`. If `orderBy` ever originates from a context where it could contain sensitive schema information (column names reflecting internal data classifications, etc.), logging or surfacing it in an unhandled-rejection handler could leak that. Consider omitting the value and logging only a fixed string, or truncating to a safe length, since the calling code already knows what it passed.

Suggestion:
    throw new Error(`scanAndTopK: orderBy value failed safety validation`);

simpleqt:
Both open-code-review findings addressed in 8e038cc: `scanAndTopK` now validates `orderBy` against an identifier-list allowlist (optional ASC/DESC per column) and throws `unsafe orderBy` before any SQL is built, so the public option can't become an injection vector even for a future caller that passes request-derived input. The repos' constant strings are unchanged, and a new test pins the guard (a `ts DESC; DROP TABLE bench --` value throws and leaves the table untouched).

`vitest run tests/unit/storage/` → 85 passed; `tsc -p tsconfig.json --noEmit` → clean.

Memtensor-AI:

All tests passed (20/20 executed). memos_local_plugin/unit: 20/20. Duration: 5s [advisory, non-gating] AI-generated tests on branch test/auto-gen-d7dd1c5d708f6868-20260831101902: 93/93 passed — these do NOT affect the PR verdict; review the branch manually.

**Branch:** `fix/tier2-recency-ordered-scan`

PR #2314, opened 2026-09-01
fix: stop merged KV caches being silently discarded on transformers >= 4.57

### LINKED ISSUE #2313: Merged KV caches report get_seq_length() == 0 and are silently discarded on transformers >= 4.57
### Environment

- MemOS at `main` (verified at `9119efe`, re-checked at `185ebdb` and still current — `src/memos/memories/activation/kv.py` is unchanged upstream since #797, 2025-12-30)
- `torch 2.13.0`, `transformers 5.15.1`, CPU, fp32, `Qwen/Qwen2.5-0.5B-Instruct`
- Version-gated: applies to `transformers >= 4.57.0`, which is inside the declared range in `pyproject.toml:41` (`>=4.51.3,<5.0.0`). `poetry.lock` pins `4.53.2`, where this does not occur.

### Summary

`_concat_caches` (`src/memos/memories/activation/kv.py:200-255`) builds each merged layer with `layer_cls()` (line 228) and then assigns `.keys` / `.values` directly. A layer constructed that way never runs `DynamicLayer.lazy_initialization`, so `is_initialized` stays `False`.

From `transformers` v4.57.0 onward, `DynamicLayer.get_seq_length()` short-circuits on exactly that flag:

```python
def get_seq_length(self) -> int:
        return 0
    return self.keys.shape[-2]
```

So a correctly concatenated cache reports a length of **0**. The attention mask is then sized for the query alone, and `DynamicLayer.update` calls `lazy_initialization`, which resets `keys`/`values` to empty tensors — **every merged token is dropped on the first forward pass.**

There is no exception and no warning. Activation memory simply has no effect, and the only symptom is that the model behaves as if the memory had never been loaded. Below v4.57 the gate is `self.keys is None`, so direct assignment survives and the bug does not appear.

This is very likely the same thing as the `test_get_cache_merge` failure that PR #2204 describes as a *"transformers-API-version issue unrelated to this patch"* — it is a version issue, but it is silent data loss rather than a test nuisance.

### Minimal reproduction

Needs only `torch` + `transformers` + a small model. The merge function is copied verbatim from `kv.py` so no MemOS install is required. Runs on CPU in under a minute.

```python
import torch, transformers
from transformers import AutoModelForCausalLM, AutoTokenizer, DynamicCache

MODEL = "Qwen/Qwen2.5-0.5B-Instruct"

# --- verbatim from src/memos/memories/activation/kv.py:200-255 (`self` removed) ---
def _concat_caches(caches):
    assert caches, "Need at least one cache"
    if len(caches) == 1:
        return caches[0]
    merged = DynamicCache()
    if hasattr(caches[0], "layers"):
        num_layers = len(caches[0].layers)
        if not hasattr(merged, "layers"):
            merged.layers = []
        if num_layers > 0:
            layer_cls = type(caches[0].layers[0])
            while len(merged.layers) < num_layers:
                merged.layers.append(layer_cls())
        for layer in range(num_layers):
            keys = [c.layers[layer].keys for c in caches]
            vals = [c.layers[layer].values for c in caches]
            merged.layers[layer].keys = torch.cat(keys, dim=-2)
            merged.layers[layer].values = torch.cat(vals, dim=-2)
    elif hasattr(caches[0], "key_cache"):
        num_layers = len(caches[0].key_cache)
        for layer in range(num_layers):
            keys = [c.key_cache[layer] for c in caches]
            vals = [c.value_cache[layer] for c in caches]
            merged.key_cache.append(torch.cat(keys, dim=-2))
            merged.value_cache.append(torch.cat(vals, dim=-2))
    else:
        raise AttributeError("no layers / key_cache")
    return merged
# ----------------------------------------------------------------------------------

print("transformers", transformers.__version__)
tok = AutoTokenizer.from_pretrained(MODEL)
model = AutoModelForCausalLM.from_pretrained(MODEL, dtype=torch.float32).eval()

def prefill(ids, cache=None):
    with torch.no_grad():
        return model(input_ids=ids, use_cache=True, past_key_values=cache,
                     return_dict=True).past_key_values

A = tok("The user's name is Alice and she lives in Berlin.", return_tensors="pt").input_ids
L = A.shape[-1]

merged = _concat_caches([prefill(A), prefill(A)])
print("get_seq_length():", merged.get_seq_length(), "expected", 2 * L,
      "| is_initialized:", getattr(merged.layers[0], "is_initialized", "n/a"))

q = tok("\nQ: In which city does the user live?", return_tensors="pt",
print("K rows before forward:", merged.layers[0].keys.shape[-2],
      "-> after forward:", prefill(q, merged).layers[0].keys.shape[-2],
      "(query length was", q.shape[-1], ")")
```

The same thing through the MemOS API — `examples/core_memories/kv_cache_memory.py` step 6 already builds this object:

```python
merged = kv_mem.get_cache([item1.id, item2.id])
assert merged.get_seq_length() == merged.layers[0].keys.shape[-2]   # fails on transformers >= 4.57
```

### Observed

```
transformers 5.15.1
get_seq_length(): 0 expected 24 | is_initialized: False
K rows before forward: 24 -> after forward: 11 (query length was 11 )
```

24 cached rows go in; 11 come out, which is the query alone. The merged content is gone.

### Expected

`merged.get_seq_length()` should equal `L1 + L2`, and the merged content should survive the first forward pass.

### Why it matters

`get_cache()` is documented and shipped (`examples/core_memories/kv_cache_memory.py:113`). On any `transformers >= 4.57` install — permitted by the declared dependency range — a user who merges two activation memories gets a cache that contributes nothing, with no error, no warning, and no log line. The failure is invisible: the model just answers as though no memory were attached.

### Suggested fix

Initialize the layer through its public `lazy_initialization` path before assigning, guarded by `hasattr` so the older `key_cache` layout is untouched. I have a patch ready with a regression test that fails on current `main` and passes with the fix, and will open it against this issue.

Note that the test fixture in `tests/memories/activation/test_kv.py` also needs repairing to see this: `make_filled_cache` appends to `cache.key_cache`, which was removed in 4.57, so `test_get_cache_merge` and `test_delete_and_get_all` fail with `AttributeError` on any current install and never reach the `layers` path where the bug lives.

### Related

A second, version-independent problem in the same function — merged fragments are not re-rotated to their landing positions — is tracked separately in draft PR #2304, and is deliberately kept apart from this one because it is a design question rather than a clear-cut bug.

jeojdi1:
## Description

Fixes #2313.

`_concat_caches` builds each merged layer with `layer_cls()` and then assigns `.keys` / `.values` directly. A layer constructed that way is never marked initialized, and on `transformers >= 4.57` `DynamicLayer.get_seq_length()` short-circuits on that flag:

```python
def get_seq_length(self) -> int:
        return 0
    return self.keys.shape[-2]
```

So a correctly concatenated cache reports a length of **0**, the model treats it as empty, and **every merged token is dropped on the first forward pass**. No exception, no warning — activation memory simply has no effect, and the only symptom is that the model behaves as if the memory were never loaded.

The fix initializes the layer through its public `lazy_initialization` path before assigning, guarded by `hasattr` so the older `key_cache` layout is untouched.

**This PR also repairs the test fixture, which is what hid the bug.** `make_filled_cache` appended to `cache.key_cache`, removed in 4.57, so `test_get_cache_merge` and `test_delete_and_get_all` failed with `AttributeError` on any current install. PR #2204's description notes these as *"pre-existing `test_kv.py` failures … transformers-API-version issues unrelated to this patch"* — they are version issues, but rebuilding the fixture on the public `update` API both makes them pass **and** makes them exercise the `layers` path, which the old fixture never did. That is what surfaces the real bug.

Related Issue (Required): #2313

## Type of change

## How Has This Been Tested?

- [x] Unit Test

Added `test_concat_caches_preserves_seq_length`, which builds two caches through the public `update` API, merges them, and asserts the merged length is the sum.

Verified against the requirement that a bug fix ships a regression test that fails on the old code and passes on the new, on `transformers 5.16.1`:

```
# upstream src/memos/memories/activation/kv.py
FAILED tests/memories/activation/test_kv.py::test_concat_caches_preserves_seq_length
1 failed

# with this change
1 passed
```

Whole file, before and after:

```
before:  2 failed, 3 passed   (test_get_cache_merge, test_delete_and_get_all)
after:   5 passed
```

## Checklist

- [ ] I have created related documentation issue/PR in MemOS-Docs (not applicable — no user-facing API change)

`CONTRIBUTING.md:283` says to branch off `dev` and open PRs against `dev`, but there is no plain `dev` branch on this repository — the branches are `main` plus `dev-v2.0.28` … `dev-v2.0.32`. This PR targets `main`, which is the default branch, where I verified the behaviour, and where 11 of the last 15 merged PRs landed. `src/memos/memories/activation/kv.py` is byte-identical on `main` and `dev-v2.0.30`, so the rebase is trivial either way — happy to retarget to whichever branch you prefer, just say which.

## Relationship to the other KV cache PRs

This is one of three independent fixes to `src/memos/memories/activation/kv.py`, each kept to a single logical change per `CONTRIBUTING.md`. They touch adjacent lines and are easiest to review in this order:

- this PR — merged caches report length 0 and are discarded (#2313)
- #2302 — serialized caches carry no model identity (#2300)
- #2303 — `get_cache` hands out the stored cache by reference, so it grows every turn (#2301)
- #2304 (draft) — fragments are not re-rotated to their landing positions

Happy to rebase them into whatever order suits review.

### DIFF
--- a/src/memos/memories/activation/kv.py
+++ b/src/memos/memories/activation/kv.py
@@ -232,8 +232,27 @@ class KVCacheMemory(BaseActMemory):
                 keys = [c.layers[layer].keys for c in caches]
                 vals = [c.layers[layer].values for c in caches]
                 # single concat per layer
-                merged.layers[layer].keys = torch.cat(keys, dim=-2)
-                merged.layers[layer].values = torch.cat(vals, dim=-2)
+                merged_keys = torch.cat(keys, dim=-2)
+                merged_values = torch.cat(vals, dim=-2)
+
+                # A layer constructed directly is not "initialized", and on
+                # transformers >= 4.57 ``DynamicLayer.get_seq_length()`` returns 0
+                # for an uninitialized layer *regardless of the tensors it holds*:
+                #
+                #     if not self.is_initialized or self.keys.numel() == 0:
+                #         return 0
+                #
+                # Assigning ``.keys``/``.values`` therefore leaves the merged cache
+                # reporting length 0, and the model treats it as empty and discards
+                # every merged token on the first forward pass -- silently, with no
+                # error. Initialize through the public path so the length is real.
+                merged_layer = merged.layers[layer]
+                if not getattr(merged_layer, "is_initialized", True) and hasattr(
+                    merged_layer, "lazy_initialization"
+                ):
+                    merged_layer.lazy_initialization(merged_keys, merged_values)
+                merged_layer.keys = merged_keys
+                merged_layer.values = merged_values
 
         # Check for old structure (key_cache)
         elif hasattr(caches[0], "key_cache"):
--- a/tests/memories/activation/test_kv.py
+++ b/tests/memories/activation/test_kv.py
@@ -33,11 +33,22 @@ def kv_memory(dummy_config):
         yield KVCacheMemory(dummy_config)
 
 
-def make_filled_cache():
-    # Create a DynamicCache with at least one dummy tensor layer
+def make_filled_cache(seq_len: int = 3, n_layers: int = 1):
+    """Create a DynamicCache with dummy tensors, on any transformers version.
+
+    This previously appended to ``cache.key_cache`` directly. That attribute was
+    removed in transformers 4.57 in favour of ``cache.layers``, so
+    ``test_get_cache_merge`` and ``test_delete_and_get_all`` failed with
+    ``AttributeError: 'DynamicCache' object has no attribute 'key_cache'`` on any
+    current install. Going through the public ``update`` API works on both the
+    old and new layouts -- and, unlike direct assignment, produces a properly
+    initialized cache, which is what makes the merge assertions meaningful.
+    """
     cache = DynamicCache()
-    cache.key_cache.append(torch.zeros(1, 2, 3))
-    cache.value_cache.append(torch.zeros(1, 2, 3))
+    for layer_idx in range(n_layers):
+        k = torch.zeros(1, 2, seq_len, 4)
+        v = torch.zeros(1, 2, seq_len, 4)
+        cache.update(k, v, layer_idx)
     return cache
 
 
@@ -58,9 +69,15 @@ def test_get_cache_merge(kv_memory):
     kv_memory.add([item1, item2])
     merged = kv_memory.get_cache([item1.id, item2.id])
     assert isinstance(merged, DynamicCache)
-    # Check the number of layers in merged key/value cache
-    assert len(merged.key_cache) == 1
-    assert len(merged.value_cache) == 1
+    # Check the number of layers in the merged cache. ``key_cache`` was removed
+    # in transformers 4.57; ``layers`` is the current layout, so accept either.
+    if hasattr(merged, "layers"):
+        assert len(merged.layers) == 1
+        assert merged.layers[0].keys.shape[-2] == 6
+        assert merged.layers[0].values.shape[-2] == 6
+    else:
+        assert len(merged.key_cache) == 1
+        assert len(merged.value_cache) == 1
 
 
 def test_delete_and_get_all(kv_memory):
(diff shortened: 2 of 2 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2314
**Task:** `67a84472b3458e97`
**Base:** `main`
**Head:** `fix/merged-kv-cache-reports-zero-length`

---

### 1. `src/memos/memories/activation/kv.py` (L250-L255)

When `is_initialized` is `False` but `lazy_initialization` is absent (e.g., a different cache implementation or a future transformers API change), both conditions fail and execution silently falls through to the bare direct assignments on lines 254–255. The comment in this very block explains that direct assignment while `is_initialized` is `False` causes silent data loss — so this fallthrough silently reproduces the bug being fixed. Consider logging a warning on the else path to make this observable.

**Before:**
```
                if not getattr(merged_layer, "is_initialized", True) and hasattr(
                    merged_layer, "lazy_initialization"
                ):
                    merged_layer.lazy_initialization(merged_keys, merged_values)
                merged_layer.keys = merged_keys
                merged_layer.values = merged_values
```

**After:**
```suggestion
                if not getattr(merged_layer, "is_initialized", True):
                    if hasattr(merged_layer, "lazy_initialization"):
                        merged_layer.lazy_initialization(merged_keys, merged_values)
                    else:
                        logger.warning(
                            "Merged cache layer is uninitialized and has no "
                            "lazy_initialization; get_seq_length() may return 0 "
                            "and merged tokens may be silently discarded."
                        )
                merged_layer.keys = merged_keys
                merged_layer.values = merged_values
```

</details>

---

### 2. `src/memos/memories/activation/kv.py` (L253-L255)

`lazy_initialization(merged_keys, merged_values)` receives the real tensors and almost certainly stores them internally (setting `is_initialized = True` and populating `self.keys`/`self.values`). The unconditional assignments that immediately follow bypass any property setters or internal bookkeeping the method applies, and are at best redundant. If the direct assignments are still required after the call, the intent should be made explicit with a comment; otherwise they should be placed in an `else` branch so they only run when `lazy_initialization` was not called.

**Before:**
```
                    merged_layer.lazy_initialization(merged_keys, merged_values)
                merged_layer.keys = merged_keys
                merged_layer.values = merged_values
```

**After:**
```suggestion
                    merged_layer.lazy_initialization(merged_keys, merged_values)
                else:
                    merged_layer.keys = merged_keys
                    merged_layer.values = merged_values
```

</details>

Memtensor-AI:

**Details:** Test collection failed because the `torch` module is not installed in the test environment, preventing the test file from being imported.
**Branch:** `fix/merged-kv-cache-reports-zero-length`

PR #2326, opened 2026-09-02
fix(skill): back off crystallization retries after repeated failures

### LINKED ISSUE #2319: fix: skill crystallization retries the same failing policies forever
### Pre-submission checklist

### Bug Description

Skill crystallization has no failure back-off. When `crystallize()` or the verifier fails for a policy, nothing changes in that policy's eligibility — so every skill tick re-attempts the same failing policy forever.

In a 25-day audit of our local deployment, this accumulated **2,640 repeated crystallization failures** across a small set of permanently failing policies (LLM refusals, verifier mismatches, permanently unqualifying evidence). Each retry burns LLM budget and log volume for a policy that can never succeed.

### How to Reproduce

1. Seed one graduated policy whose evidence can never pass the crystallizer (e.g. its traces contain no anchorable tool results)
2. Let skill ticks run repeatedly
3. Observe the same policy selected for crystallization every tick, failing every time, with no upper bound

### Environment

- memos-local-plugin: 2.0.18 (`apps/memos-local-plugin`)
- Node.js: 26
- OS: macOS

### Willingness to Implement

- [x] I'm willing to implement this myself

smoryan:
## Summary

A crystallize or verify failure leaves the policy untouched, so every skill
tick retried the same policy forever. On one long-running production install,
the 2026-08-28 audit counted **2,640 repeated crystallize failures over
25 days** — the bulk of wasted skill invocations. This PR bounds that retry
cost: after 3 consecutive failures a policy is taken off the crystallization
path; at any earlier point a successful crystallization clears the counter
and failures start counting from zero again.

## Problem

`runSkill` treats both failure modes (crystallizer skip and verifier
rejection) as advisory: it logs a warning and moves on, but the policy's
`skill_eligible` flag stays `true`. The next trigger re-selects the same
policy and the same failure repeats — indefinitely for policies whose failure
cause is permanent (e.g. evidence that can never satisfy the verifier).

## Change

- `core/skill/skill.ts` — count consecutive failures per policy in kv
  (`skill.failCount:<id>`); both failure paths call `bumpFailureBackoff()`
  before `continue`, and a successful crystallization clears the counter.
  After `SKILL_FAILURE_BACKOFF_LIMIT` (3) consecutive failures the policy's
  `skill_eligible` flag is turned off. The one deliberate exception is
  `"llm-disabled"`: that skip reason is a global configuration state, not a
  failure of the policy, so it never counts (otherwise switching the LLM off
  would trip the whole candidate pool with no recovery path).
- `core/skill/eligibility.ts` — `decide()` now reports `skill_eligible=false`
  as its own skip reason (`"policy.skillEligible=false (backoff or manual)"`),
  so a tripped backoff stays distinguishable from a manual toggle; the hidden
  check inside `hasSuccessAnchor()` is removed so the two sources don't blur.
- `core/storage/repos/policies.ts` — new `setSkillEligible()` repo method. It
  deliberately leaves `updated_at` untouched: bumping it would flip the
  rebuild heuristic for an existing skill as a side effect.
- `tests/unit/skill/backoff.test.ts` — new regression suite (5 tests): trip
  after the 3rd consecutive failure and not before, tripped policies are
  skipped by the eligibility gate with the counter cleared on trip, a
  successful crystallization clears the counter, a manually re-enabled
  policy gets a fresh backoff window, and `llm-disabled` ticks never count.

The counter is cleared when the backoff trips, so re-enabling the
`skill_eligible` flag gives the policy a fresh 3-failure window.

## Tests

- `npx vitest run tests/unit/skill` → **74 passed (12 files)** (69
  pre-existing + 5 new)
- `npx tsc --noEmit` → clean (exit 0)

## Related

- Complements #2253: that PR removed one large *source* of crystallizer
  failures (validator rejections of valid-but-incomplete drafts); this PR
  bounds the retry cost of the failures that remain (LLM refusal, verifier
  mismatches, permanently unqualifying evidence). Fixes #2319.

## Type of change

- [ ] Refactor (does not change functionality, e.g. code style improvements, lint)
- [ ] Documentation update

## How Has This Been Tested?

- [x] Unit tests (`npx vitest run tests/unit/skill` — 74 passed)
- [x] Type-check (`tsc --noEmit` clean)

## Checklist

- [ ] I have linked the issue to this PR (if applicable)

### DIFF
--- a/apps/memos-local-plugin/core/skill/eligibility.ts
+++ b/apps/memos-local-plugin/core/skill/eligibility.ts
@@ -66,6 +66,18 @@ function decide(
   existing: SkillRow | null,
   cfg: SkillConfig,
 ): EligibilityDecision {
+  // WHY: skill_eligible=false can mean either the crystallize-failure
+  // backoff (3 consecutive failures) or a manual toggle. Reporting it as
+  // its own skip reason keeps the two sources distinguishable from the
+  // experience-type success-anchor check below.
+  if (policy.skillEligible === false) {
+    return {
+      policy,
+      existingSkill: existing,
+      action: "skip",
+      reason: "policy.skillEligible=false (backoff or manual)",
+    };
+  }
   if (policy.status !== "active") {
     return {
       policy,
@@ -125,7 +137,6 @@ function decide(
 }
 
 function hasSuccessAnchor(policy: PolicyRow): boolean {
-  if (policy.skillEligible === false) return false;
   const type = policy.experienceType ?? "success_pattern";
   if (type === "failure_avoidance" || type === "repair_instruction" || type === "preference") {
     return false;
--- a/apps/memos-local-plugin/core/skill/skill.ts
+++ b/apps/memos-local-plugin/core/skill/skill.ts
@@ -28,6 +28,7 @@ import type { Repos } from "../storage/repos/index.js";
 import { now as nowMs } from "../time.js";
 import { ids } from "../id.js";
 import type {
+  PolicyId,
   PolicyRow,
   SkillId,
   SkillRow,
@@ -154,6 +155,12 @@ export async function runSkill(
         reason: crystResult.skippedReason,
         modelRefusal: crystResult.modelRefusal,
       });
+      // WHY: "llm-disabled" is a global configuration state, not a failure of
+      // this policy. Counting it would trip the whole candidate pool while
+      // the LLM is merely switched off — with no recovery path back.
+      if (crystResult.skippedReason !== "llm-disabled") {
+        bumpFailureBackoff(deps, decision.policy.id);
+      }
       continue;
     }
 
@@ -176,6 +183,7 @@ export async function runSkill(
         skillId: "sk_placeholder" as SkillId,
         reason: verdict.reason ?? "verify-failed",
       });
+      bumpFailureBackoff(deps, decision.policy.id);
       continue;
     }
 
@@ -224,6 +232,10 @@ export async function runSkill(
       });
     }
     timings.persist += nowMs() - tPersist;
+
+    // A successful crystallization resets the failure counter so later
+    // failures start counting from zero again.
+    deps.repos.kv.del(failCountKey(decision.policy.id));
 
     if (decision.action === "rebuild") rebuilt += 1;
     else crystallized += 1;
@@ -337,6 +349,26 @@ export function applySkillFeedback(
 }
 
 // ─── Helpers ──────────────────────────────────────────────────────────────
+
+// WHY: a crystallize/verify failure leaves the policy untouched, so every
+// trigger retries it forever (2026-08-28 audit: 2640 repeated failures over
+// 25 days, the bulk of wasted skill invocations). Count consecutive failures
+// in kv; after 3, flip skill_eligible=false so the policy is skipped. A
+// successful crystallization clears the counter.
+const SKILL_FAILURE_BACKOFF_LIMIT = 3;
+
+function failCountKey(policyId: string): string {
+  return `skill.failCount:${policyId}`;
+}
+
+function bumpFailureBackoff(deps: RunSkillDeps, policyId: PolicyId): void {
+  const count = deps.repos.kv.get<number>(failCountKey(policyId), 0) + 1;
+  deps.repos.kv.set(failCountKey(policyId), count);
+  if (count >= SKILL_FAILURE_BACKOFF_LIMIT) {
+    deps.repos.policies.setSkillEligible(policyId, false);
+    deps.log.warn("skill.backoff.exhausted", { policyId, count });
+  }
+}
 
 function gatherPolicies(input: RunSkillInput, repos: Repos): PolicyRow[] {
   if (input.policyId) {
--- a/apps/memos-local-plugin/core/skill/skill.ts
+++ b/apps/memos-local-plugin/core/skill/skill.ts
@@ -366,6 +366,9 @@ function bumpFailureBackoff(deps: RunSkillDeps, policyId: PolicyId): void {
   deps.repos.kv.set(failCountKey(policyId), count);
   if (count >= SKILL_FAILURE_BACKOFF_LIMIT) {
     deps.repos.policies.setSkillEligible(policyId, false);
+    // Clear the counter on trip: a later manual re-enable must get a fresh
+    // window, not instant re-trip on the next single failure.
+    deps.repos.kv.del(failCountKey(policyId));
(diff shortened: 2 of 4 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2326
**Task:** `8dbd50af769761aa`
**Base:** `main`
**Head:** `fix/skill-crystallization-backoff`

---

### 1. `apps/memos-local-plugin/core/skill/skill.ts` (L364-L374)

The failure counter is written to KV before checking whether the trip threshold is reached, but both operations are not atomic. If `setSkillEligible` throws (e.g. a DB error), the counter will have been incremented but the policy will not be marked ineligible. On the next call the counter starts from 1 again because `kv.del` is only called inside the `if` block, so the window effectively resets silently on every error. More critically, if `kv.set` succeeds but the process crashes before `kv.del` runs after tripping, the counter is already deleted by the next invocation path, but the policy flag was never set — the backoff never trips.

Consider restructuring so the KV write and the policy update happen together, or at minimum ensure `kv.del` is called in a `finally`-style guard so partial state is avoided:
```ts
if (count >= SKILL_FAILURE_BACKOFF_LIMIT) {
  deps.repos.kv.del(failCountKey(policyId)); // clear first to avoid stale state on error
  deps.repos.policies.setSkillEligible(policyId, false);
  deps.log.warn("skill.backoff.exhausted", { policyId, count });
}
```
Clearing the counter before the policy write means a crash between the two leaves the counter cleared and the policy still eligible — a safe fallback that lets it retry cleanly, rather than leaving the counter stuck at the trip value.

**Before:**
```
function bumpFailureBackoff(deps: RunSkillDeps, policyId: PolicyId): void {
  const count = deps.repos.kv.get<number>(failCountKey(policyId), 0) + 1;
  deps.repos.kv.set(failCountKey(policyId), count);
  if (count >= SKILL_FAILURE_BACKOFF_LIMIT) {
    deps.repos.policies.setSkillEligible(policyId, false);
    // Clear the counter on trip: a later manual re-enable must get a fresh
    // window, not instant re-trip on the next single failure.
    deps.repos.kv.del(failCountKey(policyId));
    deps.log.warn("skill.backoff.exhausted", { policyId, count });
  }
}
```

**After:**
```suggestion
function bumpFailureBackoff(deps: RunSkillDeps, policyId: PolicyId): void {
  const count = deps.repos.kv.get<number>(failCountKey(policyId), 0) + 1;
  deps.repos.kv.set(failCountKey(policyId), count);
  if (count >= SKILL_FAILURE_BACKOFF_LIMIT) {
    // Clear before writing the policy flag: if setSkillEligible throws,
    // the counter is already gone so the next run starts fresh rather
    // than re-tripping instantly or leaving stale state.
    deps.repos.kv.del(failCountKey(policyId));
    deps.repos.policies.setSkillEligible(policyId, false);
    deps.log.warn("skill.backoff.exhausted", { policyId, count });
  }
}
```

</details>

smoryan:
Thanks — confirmed and fixed in 3d2fe48: the counter is now cleared when the backoff trips, so a manual re-enable starts a fresh window instead of instantly re-tripping off the stale count. Locked by a new test (*gives a manually re-enabled policy a fresh backoff window*) plus two assertion updates reflecting the cleared-on-trip state. skill domain: 74 passed (12 files), tsc clean.

Memtensor-AI:

**Branch:** `fix/skill-crystallization-backoff`

zorrobyte:
Thanks for this. It fixes a real waste problem, and I'm running it locally on top of v2.0.20 (as a Hermes Agent memory provider against a self-hosted vLLM endpoint).

I found one issue while reviewing it: **transport-level failures count toward the backoff.** `runCrystallize` returns `skippedReason: "llm-failed: …"` for timeouts, endpoint-unavailable and rate-limit errors as well as for genuinely bad drafts. Only `llm-disabled` is excluded, so three transient errors (a model-server restart, a slow GPU, a 429 burst) flip `skill_eligible=false` on an otherwise good policy. Nothing re-enables it automatically. On self-hosted endpoints restarts are routine, so this can quietly retire useful policies.

Suggested fix: classify by the existing typed error codes and skip the bump for transient failures. Content failures (refusals, unusable drafts, verify failures) still count.

```ts
// crystallize.ts
const TRANSIENT_LLM_CODES: ReadonlySet<string> = new Set([
  ERROR_CODES.LLM_TIMEOUT,
  ERROR_CODES.LLM_UNAVAILABLE,
  ERROR_CODES.LLM_RATE_LIMITED,
]);
const TRANSIENT_NETWORK_RE =
  /\b(ECONNREFUSED|ECONNRESET|ETIMEDOUT|EAI_AGAIN|ENOTFOUND|EPIPE|UND_ERR_[A-Z_]+)\b|fetch failed|socket hang up|network error/i;

export function isTransientLlmError(err: unknown): boolean {
  if (err instanceof MemosError) return TRANSIENT_LLM_CODES.has(err.code);
  const message = err instanceof Error ? err.message : String(err);
  return TRANSIENT_NETWORK_RE.test(message);
}
// …both `llm-failed` returns gain: transient: isTransientLlmError(err)
// …and CrystallizeResult's failure arm gains: transient?: boolean

// skill.ts
if (crystResult.skippedReason !== "llm-disabled" && !crystResult.transient) {
  bumpFailureBackoff(deps, decision.policy.id);
}
```

Tests added to `tests/unit/skill/backoff.test.ts` (using the existing `throwingLlm` helper):
- 5× `MemosError(LLM_TIMEOUT)` gives `failCount` 0 and the policy stays eligible;
- `LLM_UNAVAILABLE`, `LLM_RATE_LIMITED` and a raw `fetch failed … ECONNREFUSED` are each asserted **per error**. Checking once at the end of the loop passes vacuously, because the backoff clears the counter when it trips;
- a plain non-transport error still counts (`failCount` 1).

Both transient tests fail without the `skill.ts` change and pass with it. The full `tests/unit` suite is otherwise unchanged. Happy to open this as a follow-up PR against your branch if that's easier.

smoryan:
Thanks @zorrobyte — great catch, and thanks for verifying it against a real deployment.

You're right: transport-level failures (timeout, connection refused, rate limiting) say nothing about the quality of the policy itself, so letting them trip the backoff can retire good policies with no recovery path. That was never the intent — the counter should only track content-level failures (refusals, unusable drafts, verification failures).

Please go ahead with a follow-up PR against this branch (`fix/skill-crystallization-backoff`). Your approach — classifying via the typed error codes plus a message-level network regex, marking `transient` on the failure arm, and skipping the bump in `skill.ts` — looks right to me. I'll review it as soon as it's up, and it can ride along when this PR lands upstream.

PR #2337, opened 2026-09-03
Preference memory fixes

bittergreen:
## Description

fix: preserve fast memories when async fine extraction fails
fix: filter polluted sources before preference extraction
fix: refine preference extraction prompts

## Type of change

## How Has This Been Tested?

- [x] Unit Test

## Checklist

## Reviewer Checklist
- [ ] Made sure Checks passed

### DIFF
--- a/src/memos/mem_reader/read_pref_memory/process_preference_memory.py
+++ b/src/memos/mem_reader/read_pref_memory/process_preference_memory.py
@@ -10,7 +10,12 @@ from typing import TYPE_CHECKING, Any
 from memos.context.context import ContextThreadPoolExecutor
 from memos.log import get_logger
 from memos.mem_reader.read_multi_modal import detect_lang
-from memos.memories.textual.item import TextualMemoryItem, TreeNodeTextualMemoryMetadata
+from memos.mem_reader.source_filter import MemorySourceFilter
+from memos.memories.textual.item import (
+    SourceMessage,
+    TextualMemoryItem,
+    TreeNodeTextualMemoryMetadata,
+)
 from memos.templates.prefer_complete_prompt import (
     NAIVE_EXPLICIT_PREFERENCE_EXTRACT_PROMPT,
     NAIVE_EXPLICIT_PREFERENCE_EXTRACT_PROMPT_ZH,
@@ -100,6 +105,7 @@ def _create_preference_memory_item(
     fast_item: TextualMemoryItem | None,
     info: dict[str, Any],
     embedder,
+    sources_override: list[SourceMessage] | None = None,
     **kwargs,
 ) -> TextualMemoryItem:
     """
@@ -133,7 +139,13 @@ def _create_preference_memory_item(
     embedding = embedder.embed([context_summary])[0] if embedder and context_summary else None
 
     # Extract sources from fast_item
-    sources = getattr(fast_item.metadata, "sources", []) if fast_item else []
+    sources = (
+        sources_override
+        if sources_override is not None
+        else getattr(fast_item.metadata, "sources", [])
+        if fast_item
+        else []
+    )
 
     # Create metadata
     metadata = TreeNodeTextualMemoryMetadata(
@@ -168,6 +180,7 @@ def _process_single_chunk_explicit(
     info: dict[str, Any],
     llm,
     embedder,
+    sources_override: list[SourceMessage] | None = None,
     **kwargs,
 ) -> list[TextualMemoryItem]:
     """Process a single chunk for explicit preferences."""
@@ -190,6 +203,7 @@ def _process_single_chunk_explicit(
             fast_item=fast_item,
             info=info,
             embedder=embedder,
+            sources_override=sources_override,
             **kwargs,
         )
         memories.append(memory)
@@ -203,6 +217,7 @@ def _process_single_chunk_implicit(
     info: dict[str, Any],
     llm,
     embedder,
+    sources_override: list[SourceMessage] | None = None,
     **kwargs,
 ) -> list[TextualMemoryItem]:
     """Process a single chunk for implicit preferences."""
@@ -225,6 +240,7 @@ def _process_single_chunk_implicit(
             fast_item=fast_item,
             info=info,
             embedder=embedder,
+            sources_override=sources_override,
             **kwargs,
         )
         memories.append(memory)
@@ -260,13 +276,20 @@ def process_preference_fine(
         return []
 
     try:
-        # Convert fast_memory_items to messages format
+        # Convert fast_memory_items to source-filtered messages format
+        source_filter = MemorySourceFilter()
         chunks = []
         for fast_item in fast_memory_items:
-            mem_str = fast_item.memory or ""
+            raw_sources = getattr(fast_item.metadata, "sources", None)
+            if raw_sources:
+                filtered_sources = source_filter.filter_for_preference(raw_sources)
+                mem_str = source_filter.sources_to_prompt_text(filtered_sources)
+            else:
+                filtered_sources = None
+                mem_str = fast_item.memory or ""
             if not mem_str.strip():
                 continue
-            chunks.append((mem_str, fast_item))
+            chunks.append((mem_str, fast_item, filtered_sources))
 
         if not chunks:
             return []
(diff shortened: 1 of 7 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2337
**Task:** `94e855360692bd94`
**Base:** `dev-v2.0.33`
**Head:** `wq-dev-v2.0.33`

---

### 1. `src/memos/mem_reader/read_pref_memory/process_preference_memory.py` (L284-L289)

`filtered_sources=None` conflates two distinct states:
1. The fast item had no `sources` field → `None` (no override, inherit from `fast_item.metadata.sources` later).
2. The fast item had sources but all were filtered away → ideally an empty list `[]` to signal "override with nothing".

Currently branch (2) never produces `[]` because the silent-drop issue above discards the item before it can reach `chunks`. Once that bug is fixed and items with all-filtered sources are allowed through with `filtered_sources=None`, `_create_preference_memory_item` will fall back to the **unfiltered** `fast_item.metadata.sources` (via `getattr(fast_item.metadata, "sources", [])`), leaking the raw, unfiltered sources into the created memory item — the opposite of what the filter is supposed to achieve.

Using an explicit sentinel (e.g. a dedicated `_SOURCES_NOT_SET` object, or always using `[]` for "empty after filtering") would make the two states unambiguous and prevent the unintended fallback.

**Before:**
```
            if raw_sources:
                filtered_sources = source_filter.filter_for_preference(raw_sources)
                mem_str = source_filter.sources_to_prompt_text(filtered_sources)
            else:
                filtered_sources = None
                mem_str = fast_item.memory or ""
```

**After:**
```suggestion
_SOURCES_NOT_SET: list[SourceMessage] | None = None  # module-level sentinel

# In process_preference_fine:
            if raw_sources:
                filtered_sources = source_filter.filter_for_preference(raw_sources)
                # filtered_sources is [] (empty list) when all sources are stripped,
                # which is a valid override that suppresses unfiltered inheritance.
                mem_str = source_filter.sources_to_prompt_text(filtered_sources)
            else:
                filtered_sources = _SOURCES_NOT_SET  # explicit: no sources field
                mem_str = fast_item.memory or ""

# In _create_preference_memory_item:
    sources = (
        getattr(fast_item.metadata, "sources", [])
        if sources_override is _SOURCES_NOT_SET
        else (sources_override or [])
    )
```

</details>

---

### 2. `src/memos/mem_scheduler/task_schedule_modules/handlers/mem_read_handler.py` (L182-L183)

Replacing `processed_memories = []` with `return` causes the handler to exit early on a transfer error, silently skipping critical cleanup logic that follows this block — specifically:
- `text_mem.delete(delete_ids, ...)` / `text_mem.soft_delete(...)` (lines 408/421): raw/working memory IDs are never cleaned up, causing them to linger indefinitely.
- `text_mem.memory_manager.remove_and_refresh_memory(user_name=user_name)` (line 440): the memory manager is never refreshed.

Previously, `processed_memories = []` let the function continue past the `if processed_memories` guard (which was already `False`), so the deletion and refresh steps still ran. The `return` breaks this.

Suggestion: keep `processed_memories = []` (or an equivalent sentinel) so the function naturally falls through to cleanup, or explicitly perform the necessary teardown in the `except` branch before returning.

---

### 3. `src/memos/mem_reader/source_filter.py` (L200)

**Bug: boundary marker at end of text silently drops the entire message.**

When a `user_query_boundary` pattern matches at the very end of the text (e.g., content is exactly `"### 用户消息为："`), `_extract_after_last_match` returns `""` (empty string). Because `""` is **not** `None`, the `extracted is None` guard does not trigger, and this line returns `"".strip()` = `""`. The caller `_strip_known_context_wrappers` then hits `if not text: return ""` and exits, so the entire source message is discarded instead of being left unchanged.

The fix is to also treat an empty extraction result as "no match found" — either by returning `None` from `_extract_after_last_match` when the extracted tail is empty, or by checking here:

**Before:**
```
            return text if extracted is None else extracted.strip()
```

**After:**
```suggestion
            if extracted is None or not extracted.strip():
                return text
            return extracted.strip()
```

</details>

---

### 4. `src/memos/mem_reader/source_filter.py` (L106)

**Potential false-negative: pattern requires whitespace between `System:` and `[`.**

`\s+` requires *one or more* whitespace characters, so `"System:[auto-task]"` (no space) will not be matched by `pattern.match()` and the source will pass through the filter unchanged, potentially leaking automation context into preference extraction prompts.

If both forms (`"System: ["` with a space and `"System:["` without) can occur in practice, use `\s*` instead:

**Before:**
```
                r"System:\s+\[",
```

**After:**
```suggestion
                r"System:\s*\[",
```

</details>

---

### 5. `tests/mem_scheduler/test_mem_read_handler_fast_cleanup.py` (L29-L30)

Bare dict access will raise `KeyError` if the production code calls `text_mem.get` with any ID not seeded in `_items`. In `_process_memories_with_reader`, the loop iterates over the caller-supplied `mem_ids` and passes each directly to `text_mem.get`; however, if a future change derives or transforms IDs before looking them up, the test crashes inside the fake with an opaque `KeyError` rather than reaching the delete-call assertions — making the no-delete guarantee unverifiable. Consider returning `None` for missing keys and letting the real logic handle the gap, or raising a more descriptive error.

**Before:**
```
    def get(self, mem_id, user_name=None):
        return self._items[mem_id]
```

**After:**
```suggestion
    def get(self, mem_id, user_name=None):
        if mem_id not in self._items:
            raise KeyError(f"_FakeTextMem: unexpected mem_id {mem_id!r} (not seeded)")
        return self._items[mem_id]
```

</details>

---

### 6. `tests/mem_scheduler/test_mem_read_handler_fast_cleanup.py` (L66-L68)

`__new__` bypasses `__init__`, so only `scheduler_context` is set on the instance. The diff of `mem_read_handler.py` shows `_process_memories_with_reader` uses only the module-level `logger` (not `self.logger`) and `self.scheduler_context`, so the current method body is safe. However, if any other method called transitively (e.g. `_try_submit_organize_task`) is patched or if the handler grows new instance-level attributes, this bare `__new__` setup will silently break. A `MagicMock(spec=MemReadMessageHandler)` with only `scheduler_context` injected — or a minimal `__init__` call — would be more resilient.

**Before:**
```
    handler = MemReadMessageHandler.__new__(MemReadMessageHandler)
    handler.scheduler_context = scheduler_context
    return handler, text_mem
```

**After:**
```suggestion
    handler = MagicMock(spec=MemReadMessageHandler)
    handler.scheduler_context = scheduler_context
    handler._process_memories_with_reader = lambda **kw: MemReadMessageHandler._process_memories_with_reader(handler, **kw)
    return handler, text_mem
```

</details>

---

### 7. `tests/mem_scheduler/test_mem_read_handler_fast_cleanup.py` (L71-L100)

Both test functions are identical except for `memory_version_switch`. The duplicated call site means any future change to the `_process_memories_with_reader` signature must be updated in two places independently — a maintenance burden. Use `@pytest.mark.parametrize` to eliminate the duplication.

**Before:**
```
def test_fine_transfer_error_does_not_delete_fast_memory():
    handler, text_mem = _build_handler(memory_version_switch="off")

    handler._process_memories_with_reader(
        mem_ids=["raw_1"],
        user_id="user_1",
        mem_cube_id="cube_1",
        text_mem=text_mem,
        user_name="cube_1",
        info={"trigger_source": "Messages"},
    )

    assert text_mem.delete_calls == []
    assert text_mem.soft_delete_calls == []


def test_fine_transfer_error_does_not_soft_delete_fast_memory_with_versions():
    handler, text_mem = _build_handler(memory_version_switch="on")

    handler._process_memories_with_reader(
        mem_ids=["raw_1"],
        user_id="user_1",
        mem_cube_id="cube_1",
        text_mem=text_mem,
        user_name="cube_1",
        info={"trigger_source": "Messages"},
    )

    assert text_mem.delete_calls == []
    assert text_mem.soft_delete_calls == []
```

**After:**
```suggestion
@pytest.mark.parametrize("version_switch", ["off", "on"])
def test_fine_transfer_error_does_not_delete_or_soft_delete(version_switch):
    handler, text_mem = _build_handler(memory_version_switch=version_switch)

    handler._process_memories_with_reader(
        mem_ids=["raw_1"],
        user_id="user_1",
        mem_cube_id="cube_1",
        text_mem=text_mem,
        user_name="cube_1",
        info={"trigger_source": "Messages"},
    )

    assert text_mem.delete_calls == []
    assert text_mem.soft_delete_calls == []
```

</details>

---

🧹 Filtered **1** low-confidence OCR finding(s) before posting/fix-loop (duplicate: 1).

Memtensor-AI:

All tests passed (13/13 executed). memos_python_core/changed-repo-python: 13/13. Duration: 8s

**Branch:** `wq-dev-v2.0.33`

whipser030 (commented):
good

whipser030 (approved):

Issue #1873, opened 2026-06-03
fix: openclaw doctor reports DuplicateOpenClawRuntimeError (false positive)


yujunos:
### Pre-submission checklist | 提交前检查

- [x] I have searched existing issues and this hasn't been mentioned before | 我已搜索现有问题，确认此问题尚未被提及
- [x] I have read the project documentation and confirmed this issue doesn't already exist | 我已阅读项目文档并确认此问题尚未存在
- [x] This issue is specific to MemOS and not a general software issue | 该问题是针对 MemOS 的，而不是一般软件问题

## Bug Description

When the OpenClaw gateway is running with `memos-local-plugin` v2.0.6, `openclaw doctor` always reports:

[plugins] memos-local: duplicate OpenClaw runtime blocked
[plugins] memos-local-plugin failed during register from .../dist/adapters/openclaw/index.js: DuplicateOpenClawRuntimeError: memos-local OpenClaw runtime is already active (pid=<gateway_pid>)

### How to Reproduce | 如何重现

The plugin works perfectly in the gateway — `onTurnStart`, `onTurnEnd`, and all memory tools function normally. Only `openclaw doctor` (a separate process) fails.

## Root Cause

`openclaw doctor` is an independent process that also loads plugins. The `acquireOpenClawRuntimeLock()` function in `runtime-lock.js` does not distinguish between:

1. A legitimate duplicate runtime instance (should be blocked)
2. The `openclaw doctor` diagnostic process (should be allowed or skipped)

## Steps to Reproduce

1. Install `@memtensor/memos-local-plugin` via `install.sh`
2. Start OpenClaw gateway (`openclaw gateway start`)
3. Run `openclaw doctor` → plugin registration fails with `DuplicateOpenClawRuntimeError`

## Expected Behavior

`openclaw doctor` should either:
- Skip memory plugin registration (read-only health check)
- Use a different lock namespace for diagnostic processes
- Report "plugin active (OK)" instead of "plugin failed"

## Environment

- Plugin: `@memtensor/memos-local-plugin@2.0.6`
- OpenClaw: `2026.5.28`
- Node.js: `v24.15.0`
- OS: Ubuntu Linux `6.8.0-117-generic`
- Install method: `curl -fsSL .../install.sh | bash`

## Workaround

The plugin actually works fine. The doctor error is a false positive and can be ignored.

### Environment | 环境信息

- Plugin: `@memtensor/memos-local-plugin@2.0.6`
[...]

PR #2350, opened 2026-09-08
fix: restore OpenClaw installation and automatic memory hooks

Hun-ger:
## Description

OpenClaw 2026.9.1/2026.9.2 reject the shell installer’s legacy `plugins.installs` records. Rebuilding a plugin registry in the gateway process can also trigger MemOS’s duplicate-runtime guard, leaving conversation hooks unavailable even though the Viewer and tools work.

This change removes only MemOS-owned legacy install records, detects supported stop/validation/capability-consent commands in both installers, and requires actual gateway health for the shell restart fallback. Plugin CLI help is probed with plugins disabled.

The adapter shares one runtime across full registries and module reloads while registering hooks in each host registry. Tool discovery borrows the existing core without owning a server or service. Cross-process locking remains in place. No dependency or package-version changes.

Related issue: #1873 (runtime-ownership context; this PR addresses same-process registry rebuilding, not that already-closed doctor issue).

## Type of change

- [x] Bug fix
- [x] Documentation update

## How Has This Been Tested?

- [x] `npm test`: **184 files passed; 1569 passed, 3 skipped**. Added configuration, gateway recovery, module reload and tool-discovery regressions.
- [x] `npm run lint` passed.
- [x] Real Windows PowerShell 5.1 harness: **7 scenarios passed** (`tests/powershell/openclaw-install-compat.ps1`).
- [x] `make format` with pinned Ruff 0.11.8: `All checks passed!`, `624 files left unchanged`.
- [x] Pre-commit checks on all changed files and `git diff --check` passed.
- [x] Actual macOS host binaries: configuration/startup/authenticated tool invocation passed on 2026.4.24, 2026.7.1-2, 2026.9.1 and 2026.9.2. Full installers passed on macOS 2026.9.1 and Windows 2026.9.2.
- [x] Automatic capture and subsequent recall passed on macOS 2026.9.1 and Windows 2026.9.2. Both platforms also passed 2026.7.1-2 tests with two separate sessions: two SQLite traces, the first turn’s unique code present in the second model request, and exactly one runtime bootstrap.

See `apps/memos-local-plugin/docs/OPENCLAW-COMPATIBILITY.md` for reproduction steps and platform-specific results. No Python library code or API contracts changed.

### Known limitations

- **2026.4.24 macOS:** standard `openclaw agent` preloads the plugin in a separate CLI process and still hits the runtime lock. Gateway capture/recall passed only with a client-only plugin-preloading workaround.
- **2026.4.24 Windows:** two isolated attempts failed gateway readiness; automatic capture/recall was not reached. Sampling showed synchronous file work in the old host’s plugin loader.

## Checklist

- [x] Self-reviewed the final changes.
- [x] Commented runtime ownership and host-version compatibility logic.
- [x] Added regression tests proving the relevant fixes.
- [x] Updated documentation in this PR; a separate MemOS-Docs change is not applicable.
- [x] Linked related issue context without claiming to close an unrelated issue.
- [x] Mentioned @MLittleprince, a recent local-plugin contributor, for review.

Maintainer review and GitHub CI results remain pending; the checks above describe completed local and machine tests.

### DIFF
--- a/apps/memos-local-plugin/README.md
+++ b/apps/memos-local-plugin/README.md
@@ -120,6 +120,28 @@ npm pack
 bash install.sh --version ./memtensor-memos-local-plugin-1.0.0-beta.1.tgz
 ```
 
+For OpenClaw, use the installer for local archives too:
+
+```bash
+bash install.sh --agent openclaw --version ./memtensor-memos-local-plugin-2.0.16-beta.1.tgz
+```
+
+Do not substitute `openclaw plugins install ./package.tgz` for this command:
+that raw-archive path can resolve development-only DeepSeek peer dependencies
+and fail with `ERESOLVE`, including on OpenClaw 2026.9.1 and 2026.9.2.
+The installer stages production dependencies and rebuilds `better-sqlite3`.
+OpenClaw's newer `npm-pack:` path avoids the peer-resolution conflict, but a
+successful managed install alone does not verify native bindings or initialize
+MemOS runtime configuration; the installer above remains the supported setup.
+
+When upgrading OpenClaw itself, migrate retired host configuration with
+`openclaw doctor --fix` before installing MemOS. The MemOS installer removes its
+own legacy `plugins.installs` records, preserves other plugins' old-host records,
+and uses the host CLI for capability consent when available. It does not rewrite
+the host's internal installation database. See
+[the compatibility test results](docs/OPENCLAW-COMPATIBILITY.md) for tested versions
+and limits.
+
 On Windows, run `install.ps1` from PowerShell instead of `install.sh` for
 OpenClaw or Hermes. The DSH one-command target currently supports macOS/Linux;
 Windows users can use DSH's lower-level `dsh plugin` flow.
--- a/apps/memos-local-plugin/adapters/openclaw/index.ts
+++ b/apps/memos-local-plugin/adapters/openclaw/index.ts
@@ -310,36 +310,12 @@ function isDiagnosticMode(): boolean {
   return false;
 }
 
-function register(api: OpenClawPluginApi): void {
-  const diagnosticMode = isDiagnosticMode();
-
-  let runtimeLock: OpenClawRuntimeLockHandle;
-  try {
-    runtimeLock = acquireOpenClawRuntimeLock({
-      home: resolveHome("openclaw"),
-      pluginId: PLUGIN_ID,
-      version: PLUGIN_VERSION,
-      viewerPort: OPENCLAW_VIEWER_PORT,
-      skipLock: diagnosticMode,
-    });
-
-    if (diagnosticMode) {
-      api.logger.info("memos-local: running in diagnostic mode (lock acquisition skipped)");
-    }
-  } catch (err) {
-    const duplicate = err instanceof DuplicateOpenClawRuntimeError;
-    api.logger.error("memos-local: duplicate OpenClaw runtime blocked", {
-      err: err instanceof Error ? err.message : String(err),
-      code: duplicate ? err.code : (err as { code?: unknown }).code,
-    });
-    throw err;
-  }
-
-  // OpenClaw publishes its clean, command-facing inbound body before
-  // prompt construction. Keep this store independent of core bootstrap
-  // so early messages are not lost while SQLite/providers initialize.
-  const inboundUserText = createOpenClawInboundTextStore();
-
+function registerRuntimeBindings(
+  api: OpenClawPluginApi,
+  ensureRuntime: () => Promise<PluginRuntime | null>,
+  inboundUserText: ReturnType<typeof createOpenClawInboundTextStore>,
+  currentRuntime: () => PluginRuntime | null,
+): void {
   // 1. Memory capability (prompt prelude) — register synchronously so the
   //    host immediately knows who owns the memory slot, even if bootstrap
   //    fails later.
@@ -391,31 +367,6 @@ function register(api: OpenClawPluginApi): void {
       return lines;
     },
   });
-
-  // 2. Kick off core bootstrap. OpenClaw only accepts tool / hook
-  //    registration during the synchronous `register(api)` window, so
-  //    tools register a shell now and wait for runtime inside execute().
-  let runtime: PluginRuntime | null = null;
-  let bootstrapError: Error | null = null;
-  const bootstrapPromise = createRuntime(api, runtimeLock, inboundUserText)
-    .then((r) => {
-      runtime = r;
-      api.logger.info("memos-local: plugin ready");
-    })
-    .catch((err) => {
-      bootstrapError = err instanceof Error ? err : new Error(String(err));
-      const duplicate = err instanceof DuplicateOpenClawRuntimeError;
-      api.logger.error("memos-local: bootstrap failed", {
-        err: bootstrapError.message,
-        code: duplicate ? err.code : (err as { code?: unknown }).code,
-      });
-    });
-
-  const ensureRuntime = async (): Promise<PluginRuntime | null> => {
-    if (runtime) return runtime;
-    await bootstrapPromise;
-    return runtime;
-  };
 
   /**
    * Helper for **void / fire-and-forget** hooks: dispatch `fn` against the
@@ -520,6 +471,7 @@ function register(api: OpenClawPluginApi): void {
   // already sync, so we can invoke it directly when the runtime is
   // ready and return undefined otherwise.
   api.on("tool_result_persist", (event, ctx) => {
+    const runtime = currentRuntime();
     if (!runtime) return; // bootstrap not finished — nothing to inject
     return runtime.bridge.handleToolResultPersist(event, ctx);
   });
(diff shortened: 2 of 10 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2350
**Task:** `ff702be2fb3531a2`
**Base:** `main`
**Head:** `fix/openclaw-install-compat`

🔍 **OpenCodeReview** found **17** issue(s) in this PR.

---

### 1. `apps/memos-local-plugin/install.ps1` (L67)

Passing $GatewayCommand as a single unquoted string to cmd.exe /c causes the shell to re-parse it, which works for simple cases but breaks if $Action ever contains spaces or special characters (e.g., an action name with a space, or shell metacharacters like &, |, >). The original code used a quoted literal; the new form removes that safety. Either keep the string interpolated inline (as the original did), or ensure $Action is validated/sanitized before being composed into $GatewayCommand.

**Before:**
```
        $GatewayOutput = @(& cmd.exe /d /c $GatewayCommand 2>&1)
```

**After:**
```suggestion
        $GatewayOutput = @(& cmd.exe /d /c "openclaw gateway $Action" 2>&1)
```

</details>

---

### 2. `apps/memos-local-plugin/install.ps1` (L64-L65)

Probing --help output to decide whether to append --force is fragile: it couples behavior to help-text wording, which can change without being a breaking API change. If a future release rewrites the help text but retains the flag, or renames the flag to something that still matches the regex, the logic silently mis-fires. A more reliable approach is to attempt the stop normally, check the exit code, and only retry with --force on the specific exit code that the CLI documents as "requires --force" (or gate on a version check if the CLI exposes one).

---

### 3. `apps/memos-local-plugin/install.ps1` (L89-L91)

Same help-text-scraping pattern: the presence of the word 'validate' anywhere in the help output (including in descriptions of other subcommands or example text) is used as the capability gate. This is an overly broad match that could produce false positives. Consider matching a more specific pattern such as a subcommand boundary, e.g., -match '(?m)^\s+validate\b', or matching the exact command synopsis line.

**Before:**
```
        $ConfigHelp = @(& cmd.exe /d /c "openclaw config --help" 2>&1) | Out-String
        if ($ConfigHelp -match 'validate') {
            $Output = @(& cmd.exe /d /c "openclaw config validate" 2>&1)
```

**After:**
```suggestion
        if ($ConfigHelp -match '(?m)^\s+validate\b') {
```

</details>

---

### 4. `apps/memos-local-plugin/install.ps1` (L105-L110)

When --accept-capabilities is NOT present in the help output, the function returns silently without enabling the plugin and without any warning to the user. A caller inspecting the result has no way to distinguish "plugin was enabled" from "plugin enable was silently skipped because the CLI is too old". Add an else branch that either throws or at minimum emits a warning so the installer doesn't proceed in a half-configured state.

**Before:**
```
        if ($EnableHelp -match '--accept-capabilities') {
            $Output = @(& cmd.exe /d /c "openclaw plugins enable memos-local-plugin --accept-capabilities" 2>&1)
            $ExitCode = $LASTEXITCODE
            $Output | ForEach-Object { Write-Host "$_" }
            if ($ExitCode -ne 0) { throw "OpenClaw could not enable the MemOS plugin (exit code $ExitCode)." }
        }
```

**After:**
```suggestion
        if ($EnableHelp -match '--accept-capabilities') {
            $Output = @(& cmd.exe /d /c "openclaw plugins enable memos-local-plugin --accept-capabilities" 2>&1)
            $ExitCode = $LASTEXITCODE
            $Output | ForEach-Object { Write-Host "$_" }
            if ($ExitCode -ne 0) { throw "OpenClaw could not enable the MemOS plugin (exit code $ExitCode)." }
        } else {
            Write-Warning "openclaw plugins enable does not support --accept-capabilities; plugin may not be enabled. Please upgrade OpenClaw and re-run the installer."
        }
```

</details>

---

### 5. `apps/memos-local-plugin/install.ps1` (L100-L104)

The env-var restore on lines 103-104 is inside the try block, not the finally block. If the cmd.exe call on line 102 throws a terminating error (e.g., openclaw is not on PATH), the restore is skipped and the process-level environment variables are left pointing at the temporary probe directory for the remainder of the process. The finally block at line 111 does restore them, but only after the throw propagates — any code that runs between the throw and the finally (e.g., nested finally blocks elsewhere) would see the wrong values. Since the finally block already handles restoration, remove the redundant mid-try restore and rely solely on finally.

**Before:**
```
        $env:OPENCLAW_STATE_DIR = $ProbeDir
        $env:OPENCLAW_CONFIG_PATH = Join-Path $ProbeDir "openclaw.json"
        $EnableHelp = @(& cmd.exe /d /c "openclaw plugins enable --help" 2>&1) | Out-String
        $env:OPENCLAW_STATE_DIR = $PreviousStateDir
        $env:OPENCLAW_CONFIG_PATH = $PreviousConfigPath
```

**After:**
```suggestion
        $EnableHelp = @(& cmd.exe /d /c "openclaw plugins enable --help" 2>&1) | Out-String
        # Env-var restoration is handled unconditionally in the finally block below.
```

</details>

---

### 6. `apps/memos-local-plugin/install.sh` (L703)

The replacement of `curl`/`lsof` with `openclaw health` conflates two distinct failure modes: (a) the subcommand does not exist on older CLIs, and (b) the gateway is genuinely unhealthy. Both cases produce a non-zero exit code, so an older CLI that has no `health` subcommand will always fall through to the `die` branch even when the gateway is already running. The old check was OS-level and always available; this one isn't.

Probe for the subcommand first and fall back to the original check when it's absent:
```bash
if "${oc_bin}" --help 2>/dev/null | grep -qw 'health'; then
  "${oc_bin}" health >/dev/null 2>&1
else
  curl -fsS --max-time 2 'http://127.0.0.1:18789' >/dev/null 2>&1 \
    || (command -v lsof >/dev/null 2>&1 && lsof -i ':18789' -t >/dev/null 2>&1)
fi
```

---

### 7. `apps/memos-local-plugin/install.sh` (L686-L690)

The temp directory allocated by `mktemp -d` is only cleaned up on the happy path. If the script exits between allocation and the `rm -rf` line — due to `set -e` on a preceding line, a signal (SIGINT, SIGTERM), or any other early exit — the directory leaks. On repeated or interrupted installs this accumulates.

Register a trap immediately after the `mktemp` call, then cancel it after cleanup:
```bash
probe_dir="$(mktemp -d)"
trap 'rm -rf -- "${probe_dir}"' EXIT
# ... use probe_dir ...
rm -rf -- "${probe_dir}"
trap - EXIT
```
If the script already has a top-level EXIT trap, append to it rather than replace it.

---

### 8. `apps/memos-local-plugin/install.sh` (L678-L679)

Grepping the parent command's help text for the string `'validate'` is fragile in both directions. It fires if the word appears anywhere in the output — examples, descriptions of unrelated flags, deprecation notices — and silently skips validation if the subcommand exists but the word is absent (abbreviated or localised help output).

Probe the subcommand directly instead:
```bash
if "${oc_bin}" config validate --help >/dev/null 2>&1; then
  "${oc_bin}" config validate || die "OpenClaw config validation failed; run openclaw doctor --fix and retry."
fi
```

---

### 9. `apps/memos-local-plugin/install.sh` (L685-L689)

The probe relies on `OPENCLAW_STATE_DIR` and `OPENCLAW_CONFIG_PATH` overrides to prevent the CLI from loading real plugin code during a `--help` call. If a CLI version ignores these variables, it loads real plugins, which may trigger network calls or credential loading as an unintended side effect. There is no verification that the override actually isolated the probe. At minimum, add a comment naming the minimum CLI version where these overrides are reliable.

---

### 10. `apps/memos-local-plugin/install.sh` (L504-L510)

Each flag is probed and appended independently, which is correct in isolation. However, the combined invocation `gateway stop --force --disable` may be rejected by intermediate CLI versions that introduced one flag before the other. Given the `|| true` guard the impact is low, but the minimum CLI version where both flags co-exist should be documented in a comment so future maintainers know whether the combination is safe.

---

### 11. `apps/memos-local-plugin/install.sh` (L685-L691)

The `--help | grep` capability-detection pattern is now used in two places (stop-args probe and accept-capabilities probe) with slightly different variable names and error-suppression styles. A third probe would likely drift further. Extract a small helper to keep detection consistent and testable:
```bash
# Returns 0 if oc_bin reports FLAG in the help output for SUBCOMMAND.
oc_supports() {
  local flag="${@: -1}"
  local -a subcmd=("${@:1:$#-1}")
  "${oc_bin}" "${subcmd[@]}" --help 2>/dev/null | grep -q -- "${flag}"
}
```

---

### 12. `apps/memos-local-plugin/adapters/openclaw/index.ts` (L568-L569)

Temporal Dead Zone (TDZ) hazard: `sharedRuntime` is referenced inside this `.catch()` closure before its `const` declaration several lines below. At runtime this is safe because the callback is always asynchronous and `sharedRuntime` will be initialized by then, but it is a latent trap — any future refactor that makes the callback synchronous (e.g., a test stub that settles the promise immediately) will throw a ReferenceError at that line.

Suggestion: declare `sharedRuntime` before `bootstrapPromise`, or introduce a `let sharedRuntimeRef: SharedRuntime | null = null` placeholder above the promise and assign it after construction, then reference that variable inside the catch handler.

---

### 13. `apps/memos-local-plugin/adapters/openclaw/index.ts` (L584-L592)

The guard `if (activeRuntimes.get(home.root) === sharedRuntime) activeRuntimes.delete(home.root)` is duplicated verbatim in both the bootstrap `.catch()` handler and the `stop()` service method. Duplicate eviction logic tends to drift — future edits (adding a log line, changing the equality semantics) typically land in one place but not the other.

Suggestion: extract a named helper before the promise chain, e.g.:
```ts
const evictSelf = () => {
  if (activeRuntimes.get(home.root) === sharedRuntime) activeRuntimes.delete(home.root);
};
```
then call `evictSelf()` in both `.catch()` and `stop()`.

---

### 14. `apps/memos-local-plugin/adapters/openclaw/index.ts` (L608-L611)

Because the `.catch()` on `bootstrapPromise` never re-throws, `await bootstrapPromise` in `stop()` always resolves regardless of whether bootstrap succeeded. This is intentional for teardown ordering, but it means a caller that invokes `stop()` after a failed bootstrap gets a silent success with no indication that shutdown was a no-op. The existing `start()` surfaces `bootstrapError` symmetrically — `stop()` should do the same.

Suggestion: after `await bootstrapPromise`, check and log the error:
```ts
if (bootstrapError) {
  api.logger.warn("memos-local: stop() called after failed bootstrap", { err: bootstrapError.message });
}
```

---

### 15. `apps/memos-local-plugin/adapters/openclaw/index.ts` (L523-L528)

Secondary registrations via this early-return path call `registerBindings` but skip the `api.registerService(...)` call at the bottom of `register`. If the host expects every plugin registration to produce a stoppable service entry, these secondary registrations will have no lifecycle hook and no `stop()` will be called on them. This may be intentional (the primary registry owns teardown), but it is not documented at the call site.

Suggestion: add an inline comment on the early return:
```ts
// Secondary registries share the runtime but do not own its lifecycle;
// the primary registry's stop() handles teardown.
return;
```

---

### 16. `apps/memos-local-plugin/tests/powershell/openclaw-install-compat.ps1` (L12)

`$Definition.Extent.Text` is dereferenced with no null guard, whereas the `$Policy` lookup right below is guarded. If `Invoke-OpenClawGatewayChecked` is renamed or removed from install.ps1, `$Ast.Find` returns `$null` and this line raises the opaque `You cannot call a method on a null-valued expression` instead of a clear 'function not found' diagnostic, making the fixture failure hard to triage. Add a guard mirroring the `$Policy` check.

**Before:**
```
Invoke-Expression $Definition.Extent.Text
```

**After:**
```suggestion
if (-not $Definition) { throw 'Missing Invoke-OpenClawGatewayChecked in installer' }
Invoke-Expression $Definition.Extent.Text
```

</details>

---

### 17. `apps/memos-local-plugin/tests/powershell/openclaw-install-compat.ps1` (L78)

`$script:ProbeDir` is never initialized at the top of the script; it is only assigned inside the `openclaw plugins enable --help` mock branch. The leak assertions (`Test-Path $script:ProbeDir`) therefore rely on a prior iteration having exercised that branch — in the `config validate` failure path, `$script:ProbeDir` still holds the stale path from the previous loop's (already-removed) directory, so the check is effectively a no-op there rather than validating cleanup. Initialize `$script:ProbeDir = $null` at script start and guard the leak assertions so a genuine leak on an early-failing path cannot be masked by stale state.

Memtensor-AI:

All tests passed (25/25 executed). memos_local_plugin/unit: 25/25. Duration: 6s

**Branch:** `fix/openclaw-install-compat`

PR #2376, opened 2026-09-16
Rebase dev-v2.0.34

bittergreen:
## Description

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Unit Test

## Checklist

## Reviewer Checklist
- [ ] Made sure Checks passed

### DIFF
--- a/apps/memos-local-plugin/README.md
+++ b/apps/memos-local-plugin/README.md
@@ -120,6 +120,28 @@ npm pack
 bash install.sh --version ./memtensor-memos-local-plugin-1.0.0-beta.1.tgz
 ```
 
+For OpenClaw, use the installer for local archives too:
+
+```bash
+bash install.sh --agent openclaw --version ./memtensor-memos-local-plugin-2.0.16-beta.1.tgz
+```
+
+Do not substitute `openclaw plugins install ./package.tgz` for this command:
+that raw-archive path can resolve development-only DeepSeek peer dependencies
+and fail with `ERESOLVE`, including on OpenClaw 2026.9.1 and 2026.9.2.
+The installer stages production dependencies and rebuilds `better-sqlite3`.
+OpenClaw's newer `npm-pack:` path avoids the peer-resolution conflict, but a
+successful managed install alone does not verify native bindings or initialize
+MemOS runtime configuration; the installer above remains the supported setup.
+
+When upgrading OpenClaw itself, migrate retired host configuration with
+`openclaw doctor --fix` before installing MemOS. The MemOS installer removes its
+own legacy `plugins.installs` records, preserves other plugins' old-host records,
+and uses the host CLI for capability consent when available. It does not rewrite
+the host's internal installation database. See
+[the compatibility test results](docs/OPENCLAW-COMPATIBILITY.md) for tested versions
+and limits.
+
 On Windows, run `install.ps1` from PowerShell instead of `install.sh` for
 OpenClaw or Hermes. The DSH one-command target currently supports macOS/Linux;
 Windows users can use DSH's lower-level `dsh plugin` flow.
--- a/apps/memos-local-plugin/adapters/openclaw/index.ts
+++ b/apps/memos-local-plugin/adapters/openclaw/index.ts
@@ -310,36 +310,12 @@ function isDiagnosticMode(): boolean {
   return false;
 }
 
-function register(api: OpenClawPluginApi): void {
-  const diagnosticMode = isDiagnosticMode();
-
-  let runtimeLock: OpenClawRuntimeLockHandle;
-  try {
-    runtimeLock = acquireOpenClawRuntimeLock({
-      home: resolveHome("openclaw"),
-      pluginId: PLUGIN_ID,
-      version: PLUGIN_VERSION,
-      viewerPort: OPENCLAW_VIEWER_PORT,
-      skipLock: diagnosticMode,
-    });
-
-    if (diagnosticMode) {
-      api.logger.info("memos-local: running in diagnostic mode (lock acquisition skipped)");
-    }
-  } catch (err) {
-    const duplicate = err instanceof DuplicateOpenClawRuntimeError;
-    api.logger.error("memos-local: duplicate OpenClaw runtime blocked", {
-      err: err instanceof Error ? err.message : String(err),
-      code: duplicate ? err.code : (err as { code?: unknown }).code,
-    });
-    throw err;
-  }
-
-  // OpenClaw publishes its clean, command-facing inbound body before
-  // prompt construction. Keep this store independent of core bootstrap
-  // so early messages are not lost while SQLite/providers initialize.
-  const inboundUserText = createOpenClawInboundTextStore();
-
+function registerRuntimeBindings(
+  api: OpenClawPluginApi,
+  ensureRuntime: () => Promise<PluginRuntime | null>,
+  inboundUserText: ReturnType<typeof createOpenClawInboundTextStore>,
+  currentRuntime: () => PluginRuntime | null,
+): void {
   // 1. Memory capability (prompt prelude) — register synchronously so the
   //    host immediately knows who owns the memory slot, even if bootstrap
   //    fails later.
@@ -391,31 +367,6 @@ function register(api: OpenClawPluginApi): void {
       return lines;
     },
   });
-
-  // 2. Kick off core bootstrap. OpenClaw only accepts tool / hook
-  //    registration during the synchronous `register(api)` window, so
-  //    tools register a shell now and wait for runtime inside execute().
-  let runtime: PluginRuntime | null = null;
-  let bootstrapError: Error | null = null;
-  const bootstrapPromise = createRuntime(api, runtimeLock, inboundUserText)
-    .then((r) => {
-      runtime = r;
-      api.logger.info("memos-local: plugin ready");
-    })
-    .catch((err) => {
-      bootstrapError = err instanceof Error ? err : new Error(String(err));
-      const duplicate = err instanceof DuplicateOpenClawRuntimeError;
-      api.logger.error("memos-local: bootstrap failed", {
-        err: bootstrapError.message,
-        code: duplicate ? err.code : (err as { code?: unknown }).code,
-      });
-    });
-
-  const ensureRuntime = async (): Promise<PluginRuntime | null> => {
-    if (runtime) return runtime;
-    await bootstrapPromise;
-    return runtime;
-  };
 
   /**
    * Helper for **void / fire-and-forget** hooks: dispatch `fn` against the
@@ -520,6 +471,7 @@ function register(api: OpenClawPluginApi): void {
   // already sync, so we can invoke it directly when the runtime is
   // ready and return undefined otherwise.
   api.on("tool_result_persist", (event, ctx) => {
+    const runtime = currentRuntime();
     if (!runtime) return; // bootstrap not finished — nothing to inject
     return runtime.bridge.handleToolResultPersist(event, ctx);
   });
@@ -542,6 +494,101 @@ function register(api: OpenClawPluginApi): void {
     void runWhenReady((r) => r.bridge.handleSubagentEnded(event, ctx), "subagent_ended");
   });
 
+}
+
+// OpenClaw may build a separate tool registry in the same process. That
+// registry borrows the full registration's core and never owns its lifecycle.
+interface SharedRuntime {
+  ensureRuntime: () => Promise<PluginRuntime | null>;
+  registerBindings: (api: OpenClawPluginApi) => void;
+}
+// Host registries can reload the module. Keep ownership process-wide while
+// retaining the filesystem lock against genuinely separate gateway processes.
+const runtimeKey = Symbol.for("memos.openclaw.activeRuntimes.v1");
+const processState = globalThis as typeof globalThis & {
+  [runtimeKey]?: Map<string, SharedRuntime>;
+};
+const activeRuntimes = processState[runtimeKey] ??= new Map<string, SharedRuntime>();
+
+function register(api: OpenClawPluginApi): void {
+  const home = resolveHome("openclaw");
+  if (api.registrationMode === "tool-discovery") {
+    registerOpenClawTools(api, {
+      agent: "openclaw",
+      getCore: async () => (await activeRuntimes.get(home.root)?.ensureRuntime())?.core ?? null,
+      log: api.logger,
+    });
+    return;
+  }
+  const existing = activeRuntimes.get(home.root);
+  if (existing) {
+    existing.registerBindings(api);
+    api.logger.info("memos-local: reused active runtime for host registry");
+    return;
+  }
+  const diagnosticMode = isDiagnosticMode();
+
+  let runtimeLock: OpenClawRuntimeLockHandle;
+  try {
+    runtimeLock = acquireOpenClawRuntimeLock({
+      home,
+      pluginId: PLUGIN_ID,
+      version: PLUGIN_VERSION,
+      viewerPort: OPENCLAW_VIEWER_PORT,
+      skipLock: diagnosticMode,
+    });
+
+    if (diagnosticMode) {
+      api.logger.info("memos-local: running in diagnostic mode (lock acquisition skipped)");
+    }
+  } catch (err) {
+    const duplicate = err instanceof DuplicateOpenClawRuntimeError;
+    api.logger.error("memos-local: duplicate OpenClaw runtime blocked", {
+      err: err instanceof Error ? err.message : String(err),
+      code: duplicate ? err.code : (err as { code?: unknown }).code,
+    });
+    throw err;
+  }
+
+  // OpenClaw publishes its clean, command-facing inbound body before
+  // prompt construction. Keep this store independent of core bootstrap
+  // so early messages are not lost while SQLite/providers initialize.
+  const inboundUserText = createOpenClawInboundTextStore();
+
+  // 2. Kick off core bootstrap. OpenClaw only accepts tool / hook
+  //    registration during the synchronous `register(api)` window, so
+  //    tools register a shell now and wait for runtime inside execute().
+  let runtime: PluginRuntime | null = null;
+  let bootstrapError: Error | null = null;
+  const bootstrapPromise = createRuntime(api, runtimeLock, inboundUserText)
+    .then((r) => {
+      runtime = r;
+      api.logger.info("memos-local: plugin ready");
+    })
+    .catch((err) => {
+      if (activeRuntimes.get(home.root) === sharedRuntime) activeRuntimes.delete(home.root);
+      bootstrapError = err instanceof Error ? err : new Error(String(err));
+      const duplicate = err instanceof DuplicateOpenClawRuntimeError;
+      api.logger.error("memos-local: bootstrap failed", {
+        err: bootstrapError.message,
+        code: duplicate ? err.code : (err as { code?: unknown }).code,
+      });
+    });
+
+  const ensureRuntime = async (): Promise<PluginRuntime | null> => {
+    if (runtime) return runtime;
+    await bootstrapPromise;
+    return runtime;
+  };
+
+  const sharedRuntime: SharedRuntime = {
+    ensureRuntime,
+    registerBindings: (target) => registerRuntimeBindings(target, ensureRuntime, inboundUserText, () => runtime),
+  };
+  activeRuntimes.set(home.root, sharedRuntime);
+
+  sharedRuntime.registerBindings(api);
+
   // 4. Service — lets the host flush + wait for ready and shut us down.
   //
   // OpenClaw's current loader (≥ 2026.4) keys the service registry by
@@ -559,6 +606,8 @@ function register(api: OpenClawPluginApi): void {
       if (bootstrapError) throw bootstrapError;
     },
     async stop() {
+      await bootstrapPromise;
+      if (activeRuntimes.get(home.root) === sharedRuntime) activeRuntimes.delete(home.root);
       if (runtime) await runtime.shutdown();
     },
   });
--- a/apps/memos-local-plugin/adapters/openclaw/openclaw-api.ts
+++ b/apps/memos-local-plugin/adapters/openclaw/openclaw-api.ts
@@ -330,6 +330,8 @@ export interface ServiceDescriptor {
 // ─── The façade we register against ───────────────────────────────────────
 
 export interface OpenClawPluginApi {
+  /** Older hosts omit this; tool discovery must not start a second runtime. */
+  registrationMode?: "full" | "discovery" | "tool-discovery" | "setup-only" | "setup-runtime" | "cli-metadata";
   /** Plugin id + metadata the host injected. */
   id: string;
   name: string;
--- a/apps/memos-local-plugin/docs/OPENCLAW-COMPATIBILITY.md
+++ b/apps/memos-local-plugin/docs/OPENCLAW-COMPATIBILITY.md
@@ -1,0 +1,133 @@
+# OpenClaw installation and automatic-memory compatibility
+
+Verified on 2026-09-08 against `main` commit `78a372a4`, using MemOS Local
+2.0.16-beta.1. The npm `latest` tag resolved to OpenClaw **2026.9.2**.
+
+## Changes
+
+- The shell installer no longer writes MemOS records into `plugins.installs`,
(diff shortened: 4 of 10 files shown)


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2376
**Task:** `17cb1fd3e51e09f2`
**Base:** `dev-v2.0.34`
**Head:** `wq-dev-v2.0.34.rebase`

🔍 **OpenCodeReview** found **25** issue(s) in this PR.

---

### 1. `apps/memos-local-plugin/install.ps1` (L62-L67)

The $GatewayCommand string is passed unquoted to cmd.exe, so if $Action ever contains spaces or shell metacharacters the command will be split incorrectly by cmd. Wrap the expanded command in quotes, or better yet pass the arguments as an array to the & operator directly rather than going through cmd.exe with a single string:

  $GatewayOutput = @(& openclaw gateway $Action --force 2>&1)

This avoids the shell-injection surface entirely.

**Before:**
```
        $GatewayCommand = "openclaw gateway $Action"
        if ($Action -eq "stop") {
            $StopHelp = @(& cmd.exe /d /c "openclaw gateway stop --help" 2>&1) | Out-String
            if ($StopHelp -match '--force') { $GatewayCommand += " --force" }
        }
        $GatewayOutput = @(& cmd.exe /d /c $GatewayCommand 2>&1)
```

**After:**
```suggestion
        $GatewayArgs = @("gateway", $Action)
        if ($Action -eq "stop") {
            $StopHelp = @(& openclaw gateway stop --help 2>&1) | Out-String
            if ($StopHelp -match '--force') { $GatewayArgs += "--force" }
        }
        $GatewayOutput = @(& openclaw @GatewayArgs 2>&1)
```

</details>

---

### 2. `apps/memos-local-plugin/install.ps1` (L102-L106)

The env-var restore happens between probing help and running the actual enable command, but only on the happy path. If the `--help` call itself throws, the finally block restores them — that is fine. However the real enable command runs after the restore, so cmd.exe will inherit the already-restored values. That is the intended behaviour, but the separation is easy to misread and could break silently if lines are reordered. Consider making the restore explicit with a dedicated inner try/finally or a clearly named helper so the ordering contract is visible.

---

### 3. `apps/memos-local-plugin/install.ps1` (L105-L110)

When $EnableHelp does not match '--accept-capabilities' the function silently succeeds without enabling the plugin. The caller at line 709 has no way to know the plugin was not actually enabled. This silent no-op could leave the installation in a broken state with no user-visible warning. At minimum emit a Write-Warn; ideally throw so the outer install fails loudly.

**Before:**
```
        if ($EnableHelp -match '--accept-capabilities') {
            $Output = @(& cmd.exe /d /c "openclaw plugins enable memos-local-plugin --accept-capabilities" 2>&1)
            $ExitCode = $LASTEXITCODE
            $Output | ForEach-Object { Write-Host "$_" }
            if ($ExitCode -ne 0) { throw "OpenClaw could not enable the MemOS plugin (exit code $ExitCode)." }
        }
```

**After:**
```suggestion
        if ($EnableHelp -match '--accept-capabilities') {
            $Output = @(& cmd.exe /d /c "openclaw plugins enable memos-local-plugin --accept-capabilities" 2>&1)
            $ExitCode = $LASTEXITCODE
            $Output | ForEach-Object { Write-Host "$_" }
            if ($ExitCode -ne 0) { throw "OpenClaw could not enable the MemOS plugin (exit code $ExitCode)." }
        } else {
            Write-Warn "openclaw plugins enable does not support --accept-capabilities; plugin may not be enabled."
        }
```

</details>

---

### 4. `apps/memos-local-plugin/install.ps1` (L98-L104)

The env vars are mutated in-process. In PowerShell, child processes inherit a copy of the environment, so cmd.exe subprocesses see the probe values as intended. However, any parallel PowerShell runspaces sharing this process (e.g. workflow jobs or -Parallel ForEach-Object) would also see the mutated values during the window between set and restore. This is unlikely to matter for an install script today, but worth noting: if concurrency is ever introduced, this pattern is not safe.

---

### 5. `apps/memos-local-plugin/adapters/openclaw/openclaw-api.ts` (L334)

The union type declares 6 distinct modes, but the implementation in `register()` (index.ts) only branches on `"tool-discovery"`. The remaining 5 values — `"full"`, `"discovery"`, `"setup-only"`, `"setup-runtime"`, and `"cli-metadata"` — all silently fall through to the full runtime-acquisition path (lock, bootstrap, service registration). If a host passes `"setup-only"` or `"cli-metadata"` expecting a lightweight, no-runtime invocation, it will instead get a full runtime started, which directly contradicts the guard the field was introduced to provide.

Either:
1. Narrow the type to only the values the plugin actually treats distinctly (`"tool-discovery"` | `string & {}` or just a broader `string`), so the contract matches the implementation; or
2. Add explicit branches for each mode that needs different handling (e.g., early-return for `"setup-only"` / `"cli-metadata"`).

**Before:**
```
  registrationMode?: "full" | "discovery" | "tool-discovery" | "setup-only" | "setup-runtime" | "cli-metadata";
```

**After:**
```suggestion
  // Only "tool-discovery" is actively branched in register(); keep the union
  // honest by removing unhandled members, or add explicit branches for each.
  registrationMode?: "full" | "tool-discovery";
```

</details>

---

### 6. `apps/memos-local-plugin/adapters/openclaw/index.ts` (L568-L576)

On bootstrap failure, `activeRuntimes` is cleaned up but `runtimeLock` is never released. Because `runtime` stays `null`, the `stop()` path's `if (runtime) await runtime.shutdown()` also skips release. The lock (filesystem and/or in-process) remains held for the rest of the process lifetime, causing any subsequent registration attempt — host module reload, a second registry, or a new gateway process — to hit `DuplicateOpenClawRuntimeError` and fail permanently despite no live runtime.

Suggestion: release the lock explicitly in the catch handler, and also guard `stop()` so the lock is released even when `runtime` is null:
```ts
.catch((err) => {
  if (activeRuntimes.get(home.root) === sharedRuntime) activeRuntimes.delete(home.root);
  runtimeLock?.release();   // release on bootstrap failure
  bootstrapError = err instanceof Error ? err : new Error(String(err));
  ...
});

// in stop():
async stop() {
  await bootstrapPromise;
  if (activeRuntimes.get(home.root) === sharedRuntime) activeRuntimes.delete(home.root);
  if (runtime) {
    await runtime.shutdown();
  } else {
    runtimeLock?.release();  // shutdown skipped, release directly
  }
}
```

---

### 7. `apps/memos-local-plugin/adapters/openclaw/index.ts` (L584-L590)

The map entry is published before `registerRuntimeBindings` is called. If any host API call inside `registerRuntimeBindings` (e.g. `api.on(...)`, capability registration) throws, the process-wide map is left with a permanently broken, partially-initialized entry. Every subsequent `register()` call will then take the `existing` branch and silently reuse a runtime whose hooks were never fully wired.

Suggestion: set the map entry only after binding registration succeeds:
```ts
try {
  sharedRuntime.registerBindings(api);
} catch (err) {
  runtimeLock?.release();
  throw err;
}
activeRuntimes.set(home.root, sharedRuntime);
```

---

### 8. `apps/memos-local-plugin/adapters/openclaw/index.ts` (L523-L528)

The reuse check is keyed only on `home.root` with no guard against the same `api` instance being registered twice. If the host calls `register(api)` again for the same instance (e.g. on a hot-reload that doesn't replace the api object), `registerRuntimeBindings` runs a second time against the same host, causing duplicate `api.on(...)` subscriptions — events will fire twice, tool capabilities will be registered twice, and side effects (like memory writes) will be doubled.

Suggestion: track which `api` instances have already received bindings, or at minimum document that the caller is responsible for ensuring single-call-per-api-instance semantics.

---

### 9. `apps/memos-local-plugin/adapters/openclaw/index.ts` (L515-L522)

The `getCore` callback silently returns `null` whenever no active runtime exists for `home.root` — which happens if tool-discovery registration arrives before the main registration, after a failed bootstrap removed the entry, or after `stop()`. The `registerOpenClawTools` call still completes without error, so tools appear registered and available, but every invocation will receive a null core and either throw internally or return empty/wrong results. This is a silent capability failure with no log or error surfaced at registration time.

Suggestion: log a warning at invocation time when core resolves to null, or defer tool registration until a runtime entry is confirmed to exist.

---

### 10. `apps/memos-local-plugin/adapters/openclaw/index.ts` (L608-L612)

Only the owning registry's `stop()` calls `runtime.shutdown()` and removes the map entry; borrowed registries that received `registerRuntimeBindings` bindings via the `existing` branch have no corresponding `stop()` path. If the host tears down a borrowed registry after the owner has already called `stop()`, any in-flight or subsequently-fired hooks in the borrowed registry will call into an already-shut-down runtime. There is no reference counter, no liveness check, and no error boundary for this case.

Suggestion: track borrowed registrations (e.g. a ref count) so `shutdown()` is deferred until all registered `api` instances have been stopped, or add a `runtime.isShutdown()` guard inside the hook dispatchers.

---

### 11. `apps/memos-local-plugin/tests/powershell/openclaw-install-compat.ps1` (L8-L12)

No null guard before using $Definition. If install.ps1 ever renames or removes `Invoke-OpenClawGatewayChecked`, this line throws a cryptic parameter-binding error on `$null` instead of a meaningful message. The sibling lookup for `Enable-OpenClawMemoryPlugin` already has the correct guard pattern — apply the same one here.

**Before:**
```
$Definition = $Ast.Find({ param($Node)
    $Node -is [System.Management.Automation.Language.FunctionDefinitionAst] -and
    $Node.Name -eq 'Invoke-OpenClawGatewayChecked'
}, $true)
Invoke-Expression $Definition.Extent.Text
```

**After:**
```suggestion
$Definition = $Ast.Find({ param($Node)
    $Node -is [System.Management.Automation.Language.FunctionDefinitionAst] -and
    $Node.Name -eq 'Invoke-OpenClawGatewayChecked'
}, $true)
if (-not $Definition) { throw 'Missing gateway helper' }
Invoke-Expression $Definition.Extent.Text
```

</details>

---

### 12. `apps/memos-local-plugin/tests/powershell/openclaw-install-compat.ps1` (L68-L77)

The snapshot and restoration assertions have no try/finally guard. If any assertion or invocation inside the loop throws before the 'Host paths not restored' check, `$env:OPENCLAW_STATE_DIR` and `$env:OPENCLAW_CONFIG_PATH` remain mutated for the remainder of the session. Since `$ErrorActionPreference = 'Stop'` is active at the top of the script, any unexpected error leaves the environment dirty for subsequent tests. Wrap the loop body (or the entire section) in try/finally to unconditionally restore the original values.

**Before:**
```
$OriginalState = $env:OPENCLAW_STATE_DIR
$OriginalConfig = $env:OPENCLAW_CONFIG_PATH
foreach ($Modern in @($false, $true)) {
    $script:Modern = $Modern
    $script:FailPolicy = ''
    $script:Calls.Clear()
    Enable-OpenClawMemoryPlugin
    $Accepted = $script:Calls.Contains('openclaw plugins enable memos-local-plugin --accept-capabilities')
    if ($Accepted -ne $Modern) { throw 'Capability consent feature detection failed' }
    if ($env:OPENCLAW_STATE_DIR -ne $OriginalState -or $env:OPENCLAW_CONFIG_PATH -ne $OriginalConfig) { throw 'Host paths not restored' }
```

**After:**
```suggestion
$OriginalState = $env:OPENCLAW_STATE_DIR
$OriginalConfig = $env:OPENCLAW_CONFIG_PATH
try {
    foreach ($Modern in @($false, $true)) {
        $script:Modern = $Modern
        $script:FailPolicy = ''
        $script:Calls.Clear()
        Enable-OpenClawMemoryPlugin
        $Accepted = $script:Calls.Contains('openclaw plugins enable memos-local-plugin --accept-capabilities')
        if ($Accepted -ne $Modern) { throw 'Capability consent feature detection failed' }
        if ($env:OPENCLAW_STATE_DIR -ne $OriginalState -or $env:OPENCLAW_CONFIG_PATH -ne $OriginalConfig) { throw 'Host paths not restored' }
```

</details>

---

### 13. `apps/memos-local-plugin/tests/powershell/openclaw-install-compat.ps1` (L78-L79)

`$script:ProbeDir` is only assigned inside the shim's `openclaw plugins enable --help` branch and is never initialized before the loop. On any iteration where that branch is not reached, `$script:ProbeDir` is `$null`, and `Test-Path $null` raises a terminating parameter-binding error under `$ErrorActionPreference = 'Stop'`, surfacing an unrelated failure message. Additionally, the variable is never reset per iteration, so the second pass validates the directory path from the first pass rather than the one just created — a leak on the second iteration would go undetected. Initialize `$script:ProbeDir` to `$null` at the top of each loop iteration and guard `Test-Path` accordingly.

**Before:**
```
    if (Test-Path $script:ProbeDir) { throw 'Temporary help configuration leaked' }
    Write-Output "PASS host policy (modern=$Modern)"
```

**After:**
```suggestion
    $script:ProbeDir = $null
    Enable-OpenClawMemoryPlugin
    # ... existing assertions ...
    if ($script:ProbeDir -ne $null -and (Test-Path $script:ProbeDir)) { throw 'Temporary help configuration leaked' }
```

</details>

---

### 14. `apps/memos-local-plugin/tests/powershell/openclaw-install-compat.ps1` (L50-L56)

Two issues with the harness state here. First, `$script:FailPolicy` is never initialized before this first loop — it is only assigned inside the later policy-failure loop. If the shim evaluates `$Command -eq $script:FailPolicy` before `$script:FailPolicy` is set, it compares against `$null`, which is always false and harmless today, but this is an implicit dependency on evaluation order that could cause confusion if the shim is extended. Second, `$script:Calls[-1]` on an empty list throws an index-out-of-range exception rather than a clear assertion failure. Add an explicit count check before accessing the last element.

**Before:**
```
foreach ($Modern in @($false, $true)) {
    $script:Modern = $Modern
    $script:Fail = $false
    $script:Calls = New-Object 'System.Collections.Generic.List[string]'
    Invoke-OpenClawGatewayChecked -Action stop
    $Expected = if ($Modern) { 'openclaw gateway stop --force' } else { 'openclaw gateway stop' }
    if ($script:Calls[-1] -ne $Expected) { throw "Wrong stop command: $($script:Calls[-1])" }
```

**After:**
```suggestion
$script:FailPolicy = ''
$script:ProbeDir = $null
foreach ($Modern in @($false, $true)) {
    $script:Modern = $Modern
    $script:Fail = $false
    $script:Calls = New-Object 'System.Collections.Generic.List[string]'
    Invoke-OpenClawGatewayChecked -Action stop
    $Expected = if ($Modern) { 'openclaw gateway stop --force' } else { 'openclaw gateway stop' }
    if ($script:Calls.Count -eq 0) { throw 'No commands were recorded' }
    if ($script:Calls[-1] -ne $Expected) { throw "Wrong stop command: $($script:Calls[-1])" }
```

</details>

---

### 15. `apps/memos-local-plugin/install.sh` (L510-L511)

Passing `--disable` on stop may write a persistent disabled marker to launchd (macOS) or systemd (Linux), preventing the gateway from auto-starting on the next boot. The recovery path in `cleanup_install_state` and the explicit `gateway start` later in this function do not pass a corresponding re-enable flag, so the service starts for the current session only and remains permanently disabled after reboot on hosts that support `--disable`.

If `--disable` is detected and applied, you need to detect and apply a corresponding `--enable` (or equivalent) on restart. One approach: track whether `--disable` was actually added to `stop_args`, then pass the symmetric flag to `gateway start` in the same way you probe for it.

**Before:**
```
    if grep -q -- '--disable' <<< "${stop_help}"; then stop_args+=(--disable); fi
    "${oc_bin}" "${stop_args[@]}" >/dev/null 2>&1 || true
```

**After:**
```suggestion
    if grep -q -- '--disable' <<< "${stop_help}"; then stop_args+=(--disable); fi
    "${oc_bin}" "${stop_args[@]}" >/dev/null 2>&1 || true
    # Track whether --disable was passed so the restart path can re-enable.
    local did_disable=false
    grep -q -- '--disable' <<< "${stop_help}" && did_disable=true
```

</details>

---

### 16. `apps/memos-local-plugin/install.sh` (L509-L510)

The unanchored `grep -q -- '--force'` pattern matches `--force` anywhere in the help text, including inside longer flag names like `--force-reinstall` or in example lines such as `e.g. --force`. A future OpenClaw CLI that adds `--force-stop` while keeping `--force` in a description sentence would cause false-positive flag injection, silently passing an unintended flag.

Anchor the match to the start of a flag token using a word-boundary or a more specific pattern, for example `grep -qE '(^|[[:space:]])--force([[:space:]]|,|$)'`. Apply the same fix to the `--accept-capabilities` probe below.

**Before:**
```
    if grep -q -- '--force' <<< "${stop_help}"; then stop_args+=(--force); fi
    if grep -q -- '--disable' <<< "${stop_help}"; then stop_args+=(--disable); fi
```

**After:**
```suggestion
    if grep -qE '(^|[[:space:]])--force([[:space:]]|,|$)' <<< "${stop_help}"; then stop_args+=(--force); fi
    if grep -qE '(^|[[:space:]])--disable([[:space:]]|,|$)' <<< "${stop_help}"; then stop_args+=(--disable); fi
```

</details>

---

### 17. `apps/memos-local-plugin/install.sh` (L691-L695)

Same unanchored grep issue as the stop-flag probes above. `--accept-capabilities` happening to appear in an example sentence or a related-flags section of help output would give a false positive. Use the same anchored pattern here for consistency.

**Before:**
```
  if grep -q -- '--accept-capabilities' <<< "${enable_help}"; then
    step "Enabling MemOS memory tools and conversation hooks"
    "${oc_bin}" plugins enable "${PLUGIN_ID}" --accept-capabilities \
      || die "OpenClaw could not enable the MemOS plugin."
  fi
```

**After:**
```suggestion
  if grep -qE '(^|[[:space:]])--accept-capabilities([[:space:]]|,|$)' <<< "${enable_help}"; then
    step "Enabling MemOS memory tools and conversation hooks"
    "${oc_bin}" plugins enable "${PLUGIN_ID}" --accept-capabilities \
      || die "OpenClaw could not enable the MemOS plugin."
  fi
```

</details>

---

### 18. `src/memos/configs/llm_rate_limit.py` (L100-L103)

Only the six explicitly listed suffixes are read from `MEMSCHEDULER_REDIS_*`. Any other variable in that namespace (e.g. a typo like `MEMSCHEDULER_REDIS_PASSWARD`) is silently ignored with no warning, leaving the Redis connection in an unexpected state (e.g. unauthenticated). Unlike the `MEMOS_LLM_RATE_LIMIT_*` namespace, there is no unsupported-key scan for `MEMSCHEDULER_REDIS_*`. Consider adding a parallel unsupported-key check, or at minimum logging a warning when unexpected `MEMSCHEDULER_REDIS_*` variables are found.

**Before:**
```
        for name in ("host", "port", "db", "username", "password", "ssl"):
            raw = os.getenv(f"MEMSCHEDULER_REDIS_{name.upper()}")
            if raw:
                env_values[f"redis_{name}"] = raw
```

**After:**
```suggestion
        _supported_redis = frozenset({f"MEMSCHEDULER_REDIS_{n.upper()}" for n in ("host", "port", "db", "username", "password", "ssl")})
        unrecognised_redis = sorted(
            name for name in os.environ
            if name.startswith("MEMSCHEDULER_REDIS_") and name not in _supported_redis
        )
        if unrecognised_redis:
            raise ConfigurationError(
                "Unrecognised MEMSCHEDULER_REDIS_* variable(s); check for typos: "
                + ", ".join(unrecognised_redis)
            )
        for name in ("host", "port", "db", "username", "password", "ssl"):
            raw = os.getenv(f"MEMSCHEDULER_REDIS_{name.upper()}")
            if raw:
                env_values[f"redis_{name}"] = raw
```

</details>

---

### 19. `src/memos/configs/llm_rate_limit.py` (L67)

The wildcard guard `model == "*"` is checked against the raw key before stripping, so a name like `"  *  "` (padded with spaces) passes both the blank check (`model.strip()` is `"*"`, which is truthy) and the wildcard check (raw value is not `"*"`). Apply the wildcard check on the stripped value to close this gap.

**Before:**
```
            if not model.strip() or model == "*":
```

**After:**
```suggestion
            if not model.strip() or model.strip() == "*":
```

</details>

---

### 20. `tests/configs/test_llm_rate_limit.py` (L57)

Unguarded dict access on `values["MOS_CHAT_MODEL"]` will produce a bare `KeyError` rather than a meaningful assertion failure if that key is ever renamed or removed from `docker/.env.example-full`. The same applies to `values[name]` in the preceding `monkeypatch.setenv` loop. Guard each lookup with an explicit `pytest.skip` or `assert` so failures are self-explanatory:
```python
assert "MOS_CHAT_MODEL" in values, "docker/.env.example-full is missing MOS_CHAT_MODEL"
```

---

### 21. `tests/configs/test_llm_rate_limit.py` (L121-L124)

`pytest.raises(ValueError)` is inconsistent with the rest of this test module, which asserts `ConfigurationError` for malformed configuration (e.g. `test_explicit_override_wins_over_invalid_environment`, `test_bad_rules_do_not_expose_contents`). `ConfigurationError` inherits from `MemOSError(Exception)`, not `ValueError`, so the two expectations are catching different exception types. The test passes today only because Pydantic v2's `ValidationError` is a `ValueError` subclass (from the model validator's `raise ValueError(...)` being wrapped), which is an implementation detail. If `validate_rules` is ever changed to raise `ConfigurationError` for consistency, this test silently fails to cover that path. Assert the module's own exception type instead:
```python
with pytest.raises((ValueError, ConfigurationError)):
    LLMRateLimitConfig.load()
```
or, if the intent is only to gate unknown rule fields:
```python
with pytest.raises(Exception):  # ValidationError wraps the ValueError from validate_rules
    LLMRateLimitConfig.load()
```

---

### 22. `tests/llms/test_qps_rate_limit.py` (L266-L267)

`SimpleNamespace` is already imported at module level (line 7) and is used by `test_permit_budget_is_shared_between_retries`. This local re-import is redundant and adds noise; remove it.

**Before:**
```
def test_closing_generator_closes_provider_stream(monkeypatch):
    from types import SimpleNamespace
```

**After:**
```suggestion
def test_closing_generator_closes_provider_stream(monkeypatch):
```

</details>

---

### 23. `tests/llms/test_qps_rate_limit.py` (L367-L375)

This test relies on real wall-clock timing: the mock script sleeps 20 ms while the acquire deadline is 5 ms. The 4× margin sounds comfortable but can still fail on a loaded CI runner — the OS scheduler might not yield control back within 5 ms of the timeout, and `time.monotonic()` is never patched here (the `isolated_settings` fixture only blocks socket connections). Consider either widening the margin (e.g., `time.sleep(0.1)` / `timeout=0.01`) or injecting a fake clock into the limiter so the test is deterministic.

---

### 24. `tests/llms/test_qps_rate_limit.py` (L129-L131)

This busy-wait loop spins on real wall-clock time to observe `pending_count == 2`. On a loaded CI runner the two `pool.submit` tasks may not have been scheduled by the time the main thread reaches the loop, causing it to spin for up to 1 second before either asserting incorrectly or racing past the assertion. A `threading.Barrier` or `threading.Event` signalled by the mock script when the second waiter has joined the queue would make this deterministic without the timing dependency.

---

### 25. `tests/llms/test_qps_rate_limit_redis.py` (L112-L121)

If `gate.wait(timeout=40)` times out, Python raises `threading.BrokenBarrierError` (a subclass of `RuntimeError`). Because nothing catches it in the child process, the process exits with a non-zero code. The parent then asserts `process.exitcode == 0`, producing a confusing `AssertionError` that hides the real cause.

Separately, if the Lua script call raises (e.g. `redis.ResponseError`, `redis.ConnectionError`), the exception propagates unhandled and `output.put(...)` is never called. The parent's `output.get(timeout=60)` will then block for the full 60 seconds before raising `queue.Empty`, making the test hang rather than fail fast.

Suggested fix: wrap the child body with explicit error handling and put a sentinel on failure so the parent can detect the failure immediately:

```python
def _process_attempts(socket_path, gate, output):
    client = redis.Redis(unix_socket_path=socket_path, decode_responses=True, socket_timeout=2)
    try:
        script = client.register_script(rate_limit.GCRA_LUA)
        client.ping()
        try:
            gate.wait(timeout=40)
        except threading.BrokenBarrierError:
            output.put(None)  # signal failure to parent
            return
        try:
            result = sum(script(keys=["test:processes"], args=[100_000_000, 2])[0] for _ in range(20))
            output.put(result)
        except Exception:
            output.put(None)
    finally:
        client.close()
```

And in the parent, assert `result is not None` before summing.

Memtensor-AI:

All tests passed (25/25 executed). memos_local_plugin/unit: 25/25. Duration: 6s

**Branch:** `wq-dev-v2.0.34.rebase`

PR #2384, opened 2026-09-18
fix(plugin): harden memory evolution and Hermes installation

Hun-ger:
## Summary

This branch improves the memory pipeline and local-plugin runtime compared with main:

- Strengthen L2 multilingual induction and L3 policy evolution with bounded clustering, retry handling, batched abstraction, complete member coverage, and policy metadata backfill.
- Prevent skill verification and reward retry loops with reverse trace-policy links, per-policy cooldowns, pending reward handling, and expanded skill integration coverage.
- Add configurable per-model Redis GCRA rate limiting and preserve LLM operation metadata through the provider facade.
- Improve preference extraction by filtering polluted sources, refining prompts, and cleaning up fast-memory fallback behavior.
- Harden OpenClaw compatibility and Hermes installation with environment detection, provider linking, staged deployment, recovery, and rollback handling.
- Add documentation, migration support, runtime compatibility checks, and regression tests across the plugin and core memory components.

## Validation

- pnpm lint passed.
- Targeted Vitest coverage passed: 32 files, 213 tests.
- Local package build and installation passed on macOS OpenClaw and Hermes.
- Installation and runtime health checks passed on Windows (192.168.0.75) for OpenClaw and Hermes.
- Installed artifacts were inspected to verify the updated runtime and pipeline code.

## Scope

- Generated package archives and local temporary directories remain untracked and are excluded from the commit.

### DIFF
--- a/apps/memos-local-plugin/core/llm/client.ts
+++ b/apps/memos-local-plugin/core/llm/client.ts
@@ -282,6 +282,7 @@ export function createLlmClientWithProvider(
       maxTokens: opts?.maxTokens ?? config.maxTokens ?? DEFAULT_MAX_TOKENS,
       jsonMode,
       stop: opts?.stop,
+      op: opts?.op,
     };
   }
 
--- a/apps/memos-local-plugin/core/llm/types.ts
+++ b/apps/memos-local-plugin/core/llm/types.ts
@@ -257,6 +257,14 @@ export interface ProviderCallInput {
   maxTokens: number;
   jsonMode: boolean;
   stop?: string[];
+  /**
+   * Logical call site (e.g. `capture.summarize`, `retrieval.filter`,
+   * `skill.evolve`). Forwarded from `LlmCallOptions.op` so providers
+   * can apply per-op behavior (request-body tweaks, routing overrides,
+   * reasoning kill-switches, per-op budget caps). Optional — providers
+   * must not assume it is set.
+   */
+  op?: string;
 }
 
 /** What providers return — pre-facade post-processing. */
--- a/apps/memos-local-plugin/tests/unit/llm/client.test.ts
+++ b/apps/memos-local-plugin/tests/unit/llm/client.test.ts
@@ -335,6 +335,51 @@ describe("llm/client", () => {
     );
   });
 
+  // ─── op propagation to providers (issue #2308) ────────────────────────
+  //
+  // The facade must forward `opts.op` onto the `ProviderCallInput` handed to
+  // `provider.complete()` / `provider.stream()` so per-op provider behavior
+  // (e.g. request-body tweaks, routing overrides, reasoning kill-switches
+  // keyed on `capture.summarize`) can fire. Dropping it silently makes
+  // those switches unreachable.
+  describe("op propagation (issue #2308)", () => {
+    it("complete forwards opts.op onto the provider input", async () => {
+      const fake = new FakeProvider("openai_compatible", () => ({ text: "ok", durationMs: 1 }));
+      const client = createLlmClientWithProvider(cfg(), fake);
+      await client.complete("hi", { op: "capture.summarize" });
+      expect(fake.lastInput?.op).toBe("capture.summarize");
+    });
+
+    it("completeJson forwards opts.op onto the provider input", async () => {
+      const fake = new FakeProvider("openai_compatible", () => ({
+        text: '{"a":1}',
+        durationMs: 1,
+      }));
+      const client = createLlmClientWithProvider(cfg(), fake);
+      await client.completeJson<{ a: number }>("score", { op: "retrieval.filter" });
+      expect(fake.lastInput?.op).toBe("retrieval.filter");
+    });
+
+    it("stream forwards opts.op onto the provider input (non-streaming provider)", async () => {
+      // FakeProvider has no stream(); the facade wraps complete() in a
+      // single-chunk iterable, which still exercises buildCallInput.
+      const fake = new FakeProvider("openai_compatible", () => ({ text: "one", durationMs: 1 }));
+      const client = createLlmClientWithProvider(cfg(), fake);
+      const parts: string[] = [];
+      for await (const c of client.stream("x", { op: "skill.evolve" })) {
+        if (!c.done) parts.push(c.delta);
+      }
+      expect(fake.lastInput?.op).toBe("skill.evolve");
+    });
+
+    it("omits op field when caller supplies no op (contract stays optional)", async () => {
+      const fake = new FakeProvider("openai_compatible", () => ({ text: "ok", durationMs: 1 }));
+      const client = createLlmClientWithProvider(cfg(), fake);
+      await client.complete("hi");
+      expect(fake.lastInput?.op).toBeUndefined();
+    });
+  });
+
   // ─── Circuit breaker (issue #1897) ──────────────────────────────────────
   describe("circuit breaker", () => {
     function statusSink(): { rows: LlmStatusDetail[]; push: (d: LlmStatusDetail) => void } {
--- a/apps/memos-local-plugin/tests/unit/llm/client.test.ts
+++ b/apps/memos-local-plugin/tests/unit/llm/client.test.ts
@@ -58,11 +58,17 @@ class FakeProvider implements LlmProvider {
 
 class StreamingProvider implements LlmProvider {
   readonly name: LlmProviderName = "openai_compatible";
+  public lastInput: ProviderCallInput | null = null;
+
   async complete(): Promise<ProviderCompletion> {
     return { text: "full", durationMs: 1 };
   }
   // eslint-disable-next-line require-yield
-  async *stream(): AsyncGenerator<LlmStreamChunk> {
+  async *stream(
+    _messages: LlmMessage[],
+    opts: ProviderCallInput,
+  ): AsyncGenerator<LlmStreamChunk> {
+    this.lastInput = opts;
     yield { delta: "he", done: false };
     yield { delta: "llo", done: false };
     yield {
@@ -372,7 +378,16 @@ describe("llm/client", () => {
       expect(fake.lastInput?.op).toBe("skill.evolve");
     });
 
-    it("omits op field when caller supplies no op (contract stays optional)", async () => {
+    it("stream forwards opts.op onto a native streaming provider", async () => {
+      const provider = new StreamingProvider();
+      const client = createLlmClientWithProvider(cfg(), provider);
+      for await (const _chunk of client.stream("x", { op: "capture.summarize" })) {
+        // Consume the stream so the provider receives the cooked input.
+      }
+      expect(provider.lastInput?.op).toBe("capture.summarize");
+    });
+
+    it("leaves op undefined when the caller supplies no op", async () => {
       const fake = new FakeProvider("openai_compatible", () => ({ text: "ok", durationMs: 1 }));
       const client = createLlmClientWithProvider(cfg(), fake);
       await client.complete("hi");
--- a/apps/memos-local-plugin/tests/unit/llm/client.test.ts
+++ b/apps/memos-local-plugin/tests/unit/llm/client.test.ts
@@ -322,6 +322,51 @@ describe("llm/client", () => {
     );
   });
 
+  // ─── op propagation to providers (issue #2308) ────────────────────────
+  //
+  // The facade must forward `opts.op` onto the `ProviderCallInput` handed to
+  // `provider.complete()` / `provider.stream()` so per-op provider behavior
+  // (e.g. request-body tweaks, routing overrides, reasoning kill-switches
+  // keyed on `capture.summarize`) can fire. Dropping it silently makes
+  // those switches unreachable.
+  describe("op propagation (issue #2308)", () => {
+    it("complete forwards opts.op onto the provider input", async () => {
+      const fake = new FakeProvider("openai_compatible", () => ({ text: "ok", durationMs: 1 }));
+      const client = createLlmClientWithProvider(cfg(), fake);
+      await client.complete("hi", { op: "capture.summarize" });
+      expect(fake.lastInput?.op).toBe("capture.summarize");
+    });
+
+    it("completeJson forwards opts.op onto the provider input", async () => {
+      const fake = new FakeProvider("openai_compatible", () => ({
+        text: '{"a":1}',
+        durationMs: 1,
+      }));
+      const client = createLlmClientWithProvider(cfg(), fake);
+      await client.completeJson<{ a: number }>("score", { op: "retrieval.filter" });
+      expect(fake.lastInput?.op).toBe("retrieval.filter");
+    });
+
+    it("stream forwards opts.op onto the provider input (non-streaming provider)", async () => {
+      // FakeProvider has no stream(); the facade wraps complete() in a
+      // single-chunk iterable, which still exercises buildCallInput.
+      const fake = new FakeProvider("openai_compatible", () => ({ text: "one", durationMs: 1 }));
+      const client = createLlmClientWithProvider(cfg(), fake);
+      const parts: string[] = [];
+      for await (const c of client.stream("x", { op: "skill.evolve" })) {
+        if (!c.done) parts.push(c.delta);
+      }
+      expect(fake.lastInput?.op).toBe("skill.evolve");
+    });
+
+    it("omits op field when caller supplies no op (contract stays optional)", async () => {
+      const fake = new FakeProvider("openai_compatible", () => ({ text: "ok", durationMs: 1 }));
+      const client = createLlmClientWithProvider(cfg(), fake);
+      await client.complete("hi");
+      expect(fake.lastInput?.op).toBeUndefined();
+    });
+  });
+
   // ─── Circuit breaker (issue #1897) ──────────────────────────────────────
   describe("circuit breaker", () => {
     function statusSink(): { rows: LlmStatusDetail[]; push: (d: LlmStatusDetail) => void } {


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2384
**Task:** `2d03d4cbe1ba4fef`
**Base:** `main`
**Head:** `fix-20260916-local-plugin`

🔍 **OpenCodeReview** found **25** issue(s) in this PR.

---

### 1. `apps/memos-local-plugin/core/llm/client.ts` (L850-L852)

The `record.code` string is accepted and cast directly to the `ERROR_CODES` union without checking whether it is actually a valid member of that union. An arbitrary bridge-error payload could carry any string here, silently bypassing the type constraint. Consider validating with an explicit allowlist check before using the cast, and falling back to `fallbackCode` for unrecognised values.

**Before:**
```
    const code = typeof record.code === "string" ? record.code : fallbackCode;
    const message = typeof record.message === "string" ? record.message : String(record.data ?? err);
    return new MemosError(code as (typeof ERROR_CODES)[keyof typeof ERROR_CODES], `${prefix}: ${message}`, {
```

**After:**
```suggestion
    const knownCodes = Object.values(ERROR_CODES) as string[];
    const rawCode = typeof record.code === "string" && knownCodes.includes(record.code) ? record.code : fallbackCode;
    const code = rawCode as (typeof ERROR_CODES)[keyof typeof ERROR_CODES];
    const message = typeof record.message === "string" ? record.message : String(record.data ?? err);
    return new MemosError(code, `${prefix}: ${message}`, {
```

</details>

---

### 2. `apps/memos-local-plugin/core/memory/l3/cluster.ts` (L30)

`maxPoliciesPerCluster` is added to `ClusterDeps.config` but is never read anywhere in this file — neither `clusterPolicies` nor `clusterUntagged` reference it. The cap is enforced in `l3.ts` via a batching loop, not here. Requiring the field in `ClusterDeps` without using it misleads callers into believing the cluster size is bounded by this function, and widens the required config surface for no benefit. Either remove it from this `Pick` or actually apply the cap inside `clusterUntagged` (e.g. break a group once it reaches `config.maxPoliciesPerCluster` members).

---

### 3. `apps/memos-local-plugin/core/memory/l3/cluster.ts` (L229)

The minimum-size threshold is inconsistent between code paths: the main loop uses `config.minPolicies` directly, but `clusterUntagged` silently floors it at `2`. If `minPolicies` is configured as `1`, tagged buckets admit single-policy clusters while untagged buckets silently require two members, causing untagged policies to be dropped from clustering with no visible indication in configuration or logs. Both paths should derive the threshold from one expression — either enforce `minPolicies >= 2` in the schema or apply `Math.max(2, config.minPolicies)` in both places.

---

### 4. `apps/memos-local-plugin/core/memory/l3/cluster.ts` (L195)

`ordered` is a sorted copy of `cohort`, so its length equals `cohort.length`. `cohort` is only assigned when `strict.length >= requiredPolicies` or `members.length >= requiredPolicies`, so `ordered.length < requiredPolicies` is always false. This check is dead code and should be removed to avoid implying a safety net that cannot trigger.

---

### 5. `apps/memos-local-plugin/core/memory/l3/cluster.ts` (L185)

The `key !== "_|_"` guard is unreachable. The early branch at line 126 already dispatches `_|_` keys to `clusterUntagged` and calls `continue`, so control never reaches this `else if` when `key === "_|_"`. The guard is misleading dead code; removing it clarifies that the two-stage admission logic only applies to tagged buckets.

---

### 6. `apps/memos-local-plugin/core/memory/l3/cluster.ts` (L237-L238)

The centroid is recomputed from scratch for every group on every member insertion, making the greedy loop O(n² · d) in embedding work and allocating a temporary array on each inner iteration. Consider caching the running sum and count per group to compute the centroid incrementally, or at minimum caching the last-known centroid per group outside the inner loop and invalidating it only on push.

---

### 7. `apps/memos-local-plugin/core/memory/l3/cluster.ts` (L256)

The sentinel string `"_|_"` is hardcoded in three separate locations in this file (line 126, line 185, and this key template). It is also the output of `domainKeyOf` when a policy matches no tag or tool regex. If that join separator ever changes, all three guards and the generated key prefix silently diverge. Extract a shared constant (e.g. `const UNTAGGED_KEY = "_|_" as const`) and build the vector-cluster key from it.

---

### 8. `apps/memos-local-plugin/core/memory/l3/l3.ts` (L200)

If `cluster.policies` is empty (length 0), the `for` loop never executes, `drafts` stays empty, and `combineBatchDrafts([])` is called. Inside it, `const first = drafts[0]!` evaluates to `undefined` (the non-null assertion suppresses the type error), so `first.draft.title` throws a `TypeError`. Additionally, `weightedPolicyCount` would be 0, making the confidence calculation produce `NaN`/`Infinity`. Add a guard before calling `combineBatchDrafts` and handle the zero-drafts case explicitly.

**Before:**
```
    draftRes ??= { ok: true, draft: combineBatchDrafts(drafts) };
```

**After:**
```suggestion
    if (drafts.length === 0 && draftRes === null) {
      abstractions.push(skipped(cluster, "no_centroid", { episodeIds, policyIds: [] }));
      continue;
    }
    draftRes ??= { ok: true, draft: combineBatchDrafts(drafts) };
```

</details>

---

### 9. `apps/memos-local-plugin/core/memory/l3/l3.ts` (L297-L301)

In the `update` branch, `persisted = true` is placed at the very end of the `try` block — after `updateBody`, `embeddingRetryQueue.enqueue`, `updateConfidence`, and `emit`. If any operation after `updateBody` throws (e.g. `emit` throws on a broken bus handler), the `catch` block runs and `persisted` remains `false`, causing `recordFailure` to be called even though the world model was already mutated. Move `persisted = true` immediately after the critical write (`updateBody`) to correctly reflect that the durable write succeeded.

**Before:**
```
        persisted = true;
      } catch (err) {
        warnings.push(stageWarn("merge", err, { clusterKey: cluster.key }));
      }
    } else {
```

**After:**
```suggestion
        repos.worldModel.updateBody(decision.target.id, {
          title: patch.title,
          body: patch.body,
          structure: patch.structure,
          domainTags: patch.domainTags,
          policyIds: patch.policyIds,
          sourceEpisodeIds: patch.sourceEpisodeIds,
          vec: patch.vec,
          updatedAt: now,
        });
        persisted = true; // mark immediately after the durable write
        ...
```

</details>

---

### 10. `apps/memos-local-plugin/core/memory/l3/l3.ts` (L347-L353)

Same issue as in the `update` branch: `persisted = true` is at the end of the `try` block in the `insert` path. If `emit` or `embeddingRetryQueue.enqueue` throws after `repos.worldModel.insert` succeeds, `persisted` stays `false` and `recordFailure` is called for a cluster whose world model was already inserted. Move `persisted = true` immediately after `repos.worldModel.insert(wm)` to accurately capture that the durable write succeeded.

**Before:**
```
        persisted = true;
      } catch (err) {
        warnings.push(stageWarn("insert", err, { clusterKey: cluster.key }));
      }
    }

    if (persisted) {
```

**After:**
```suggestion
        repos.worldModel.insert(wm);
        persisted = true; // mark immediately after the durable write
        ...
```

</details>

---

### 11. `apps/memos-local-plugin/core/memory/l3/l3.ts` (L194-L198)

When any batch fails, the loop discards all previously completed batches and breaks. For a large cluster that fails on, say, its last batch, this wastes N−1 successful LLM calls on every retry attempt, compounding the cost on exactly the failure path the new backoff mechanism targets. Consider saving the partial `drafts` alongside the failure state (e.g. in the KV retry record) so a subsequent run can resume from the failed offset instead of restarting from 0.

---

### 12. `apps/memos-local-plugin/core/memory/l3/l3.ts` (L532-L535)

The retry key encodes the full sorted policy member list. Any membership change (policy added/removed from the cluster) generates a brand-new key, silently resetting the failure counter and defeating the backoff for churning clusters. Conversely, `kv.del(retryKey(cluster))` on success only removes the entry for the *current* membership; keys written under previous memberships are never deleted, causing unbounded growth of `l3.retry.*` entries. `cluster.key` already uniquely identifies the cluster — use it alone as the retry key for consistency with `cooldownKey`.

**Before:**
```
function retryKey(cluster: PolicyCluster): string {
  const members = cluster.policies.map((p) => String(p.id)).sort().join(",");
  return `${KV_RETRY_PREFIX}${cluster.key}:${members}`;
}
```

**After:**
```suggestion
function retryKey(cluster: PolicyCluster): string {
  return `${KV_RETRY_PREFIX}${cluster.key}`;
}
```

</details>

---

### 13. `apps/memos-local-plugin/core/memory/l3/l3.ts` (L423-L424)

`title` is taken exclusively from the first batch. If `maxPoliciesPerCluster` splits semantically distinct policy groups across batches, the resulting world model is titled by whichever policies happen to be at offset 0 — an order-dependent, silent semantic loss. At minimum, document this limitation. If correctness matters here, consider a short follow-up LLM call to synthesize a title from all batch titles instead of picking one arbitrarily.

---

### 14. `apps/memos-local-plugin/core/memory/l3/l3.ts` (L107-L111)

`defaults.ts` already declares `maxPoliciesPerCluster: 20` in `DEFAULT_CONFIG.algorithm.l3Abstraction`, so the `?? 20` fallback here (and the identical one in `Math.max(1, config.maxPoliciesPerCluster ?? 20)` on the batching line) duplicates the canonical default. This creates a three-way drift risk: a future change to `defaults.ts` won't automatically update either inline literal. Remove the inline fallbacks and trust the config resolution layer to supply the default.

**Before:**
```
        config: {
          clusterMinSimilarity: config.clusterMinSimilarity,
          minPolicies: config.minPolicies,
          maxPoliciesPerCluster: config.maxPoliciesPerCluster ?? 20,
        },
```

**After:**
```suggestion
        config: {
          clusterMinSimilarity: config.clusterMinSimilarity,
          minPolicies: config.minPolicies,
          maxPoliciesPerCluster: config.maxPoliciesPerCluster,
        },
```

</details>

---

### 15. `apps/memos-local-plugin/core/skill/subscriber.ts` (L178-L181)

Hard-coded `limit: 200` silently truncates results when there are more than 200 active policies. Policies beyond the cutoff will never receive a reward signal for this episode, causing silent data loss in the learning loop. The config diffs for this PR introduce no corresponding config key (e.g. `maxActivePolicyLimit`) to drive this value, making it purely arbitrary. Either use a config-driven limit (and add pagination if the count can legitimately exceed it), or replace this lookup with a dedicated repo method that queries by `sourceEpisodeIds` directly (similar to the new `getLinkedPolicyIds` SQL query added in this PR) to avoid the full-table scan and the truncation risk altogether.

---

### 16. `apps/memos-local-plugin/core/skill/subscriber.ts` (L84)

`lastRewardRunAt` is never pruned. Every `PolicyId` that ever receives a reward run is inserted and retained for the lifetime of the subscriber. In a long-running process with ongoing policy creation, archival, and deletion, this map grows without bound and holds references to IDs of policies that no longer exist. Consider clearing an entry when its policy is archived or deleted (via the event bus), or evicting entries whose age exceeds a multiple of `cooldownMs`.

---

### 17. `apps/memos-local-plugin/core/skill/subscriber.ts` (L93)

The `!` non-null assertion is unnecessary here. The `while (queued.length > 0)` guard structurally guarantees `shift()` cannot return `undefined`. Keeping `!` implies to readers that the empty case was deliberately suppressed rather than being structurally impossible, which is misleading. Remove it.

**Before:**
```
      const next = queued.shift()!;
```

**After:**
```suggestion
      const next = queued.shift();
```

</details>

---

### 18. `apps/memos-local-plugin/tests/python/test_hermes_install_environment.py` (L30)

`str.index()` raises an unhandled `ValueError` if the exact Unicode em-dash comment heading `# ─── Hermes install` is ever renamed, reflowed, or the em-dash is replaced with a plain hyphen. This would crash the entire test module at collection time with no diagnostic pointing back to the root cause. Use `str.find()` followed by an explicit `assert` with a descriptive message, mirroring the pattern already used for `start` just above.

**Before:**
```
    end = source.index("# ─── Hermes install", start)
```

**After:**
```suggestion
    end = source.find("# ─── Hermes install", start)
    assert end >= 0, "installer section marker '# ─── Hermes install' not found; update this landmark if the installer heading changed"
```

</details>

---

### 19. `apps/memos-local-plugin/tests/python/test_hermes_install_environment.py` (L141-L142)

Same fragility as in `probe()`: `str.index()` raises `ValueError` if `# ─── Main` is renamed or the Unicode em-dash is changed, crashing the whole test file. Prefer `str.find()` + `assert` with a clear message so the failure points directly at the stale landmark string.

**Before:**
```
    start = source.index("resolve_hermes_environment() {")
    end = source.index("# ─── Main", start)
```

**After:**
```suggestion
    start = source.find("resolve_hermes_environment() {")
    assert start >= 0, "installer function 'resolve_hermes_environment()' not found"
    end = source.find("# ─── Main", start)
    assert end >= 0, "installer section marker '# ─── Main' not found; update this landmark if the installer heading changed"
```

</details>

---

### 20. `apps/memos-local-plugin/tests/python/test_hermes_install_environment.py` (L154-L160)

This test uses a fully stripped environment dict (no `os.environ` inheritance) while the `probe()` helper preserves the current process's environment minus `HERMES_*` keys. More importantly, this hand-crafted env omits variables like `TMPDIR`, `LANG`, and `LC_ALL` that some bash built-ins or shell utilities depend on. If the shell script internally calls `mktemp` or locale-sensitive tools and the defaults differ by platform, test behaviour will be non-deterministic. Use the same env-filtering approach as `probe()` — strip `HERMES_*` keys from `os.environ` and then layer the overrides on top — so the two approaches are consistent.

**Before:**
```
    result = subprocess.run(
        ["/bin/bash", "-c", script],
        env={
            "HOME": str(tmp_path),
            "PATH": "/usr/bin:/bin",
            "HERMES_PYTHON": str(tmp_path / "missing"),
        },
```

**After:**
```suggestion
    env = {k: v for k, v in os.environ.items() if not k.startswith("HERMES_")}
    env.update(HOME=str(tmp_path), PATH="/usr/bin:/bin", HERMES_PYTHON=str(tmp_path / "missing"))
    result = subprocess.run(
        ["/bin/bash", "-c", script],
        env=env,
```

</details>

---

### 21. `apps/memos-local-plugin/install.sh` (L911-L920)

The `rm -rf "${HERMES_LIVE_PREFIX}"` call has no error check. If it fails (e.g., permission denied, a daemon re-opened a file), the code falls through to the `mv` of the backup, which will fail because the destination still exists. That `mv` failure is only a `warn` (non-fatal), so the function returns success while the live prefix is left in a corrupted state and the backup is consumed with no actionable message to the user.

Suggestion: check the `rm` result and emit an error that names the manual recovery path if it fails:
```bash
if [[ -d "${HERMES_LIVE_PREFIX}" ]]; then
  rm -rf "${HERMES_LIVE_PREFIX}" \
    || { warn "Cannot remove live prefix ${HERMES_LIVE_PREFIX}; backup preserved at ${HERMES_PREFIX_BACKUP_ROOT}/live — manual recovery required."; return 1; }
fi
```

---

### 22. `apps/memos-local-plugin/install.sh` (L989-L991)

Declaring a variable `local` and then `export`-ing it in bash does make the variable local to the function's scope, but `export` also marks it for inheritance by every child process spawned within that scope. Every subsequent Python call in `install_hermes` — the profile defaults hook (`<<'PYEOF'`), the config patcher, the provider verification, and any `node` subprocesses that internally invoke Python — will inherit `PYTHONPATH="${hermes_source}"` and an empty `PYTHONHOME`. This can load wrong package versions for callers that have their own environment expectations, and it silently overwrites any `PYTHONPATH` the user had set.

Prefer passing the variable inline only to the specific calls that need it, or use a subshell to contain the export:
```bash
# Option A — inline for each Python call that needs it
PYTHONPATH="${hermes_source}" PYTHONHOME="" "${python_bin}" ...

# Option B — subshell scope
(
  export PYTHONPATH="${hermes_source}" PYTHONHOME=""
  # Hermes-specific steps here
)
```

---

### 23. `apps/memos-local-plugin/install.sh` (L1392-L1404)

The success-path cleanup at the end of `install_hermes` clears rollback state with individual `rm -rf` calls followed by variable resets. The `cleanup_install_state` EXIT trap unconditionally invokes `rollback_hermes_install`. If SIGINT/SIGTERM or an ERR trap fires mid-sequence — after, say, `HERMES_PROVIDER_BACKUP_DIR` is deleted but before the arrays are cleared — `rollback_hermes_install` will attempt to restore from non-existent backup paths. All restore failures in that path are `warn`-only, so the function returns 0 and the user sees no actionable error.

The safest fix is to consolidate this clearing into a single helper that is also called by `rollback_hermes_install` after a successful restore (which `restore_hermes_provider_targets` already does for the arrays). That way a mid-sequence interruption always leaves a consistent state. At minimum, clear the flag variables *before* deleting the directories so the rollback guard can detect that cleanup is already done:
```bash
# Disarm the rollback before removing backup material
HERMES_PREFIX_SWAPPED="false"
HERMES_LIVE_PREFIX=""
HERMES_PREFIX_BACKUP_ROOT_TMP="${HERMES_PREFIX_BACKUP_ROOT}"
HERMES_PREFIX_BACKUP_ROOT=""
BACKUP_DIR_TMP="${HERMES_PROVIDER_BACKUP_DIR}"
HERMES_PROVIDER_BACKUP_DIR=""
HERMES_PROVIDER_BACKUP_TARGETS=()
HERMES_PROVIDER_BACKUP_PATHS=()
[[ -n "${BACKUP_DIR_TMP}" ]]         && rm -rf "${BACKUP_DIR_TMP}"
[[ -n "${HERMES_PREFIX_BACKUP_ROOT_TMP}" ]] && rm -rf "${HERMES_PREFIX_BACKUP_ROOT_TMP}"
```

---

### 24. `apps/memos-local-plugin/install.sh` (L959-L975)

The global arrays `HERMES_PROVIDER_BACKUP_TARGETS` and `HERMES_PROVIDER_BACKUP_PATHS` are appended to (`+=`) inside `prepare_hermes_provider_targets` but never reset at entry. If this function were ever called a second time in the same session (e.g., a future retry loop or a dual-agent code path that also touches Hermes), stale entries from the first call would remain and be re-processed during rollback, causing double-restore attempts on already-moved or already-deleted paths.

Add a reset at the top of the function:
```bash
prepare_hermes_provider_targets() {
  HERMES_PROVIDER_BACKUP_TARGETS=()
  HERMES_PROVIDER_BACKUP_PATHS=()
  local backup_root="$1"
  shift
  ...
}
```

---

### 25. `apps/memos-local-plugin/install.sh` (L879-L882)

The probe timeout is hard-coded at 15 seconds with no user-configurable override. On network-mounted filesystems (NFS, SSHFS) or machines with slow Python cold-start, importing `hermes_cli` can legitimately exceed this, producing a spurious preflight failure that blocks the entire install. There is no `HERMES_PROBE_TIMEOUT` (or similar) environment variable documented or honoured.

Read the value from the environment with a safe default:
```python
try:
    _timeout = int(os.environ.get("HERMES_PROBE_TIMEOUT", "15"))
except ValueError:
    _timeout = 15
...
result = subprocess.run(
    [python, "-c", probe, root], env=child_env, cwd="/",
    capture_output=True, text=True, timeout=_timeout,
)
```
Document `HERMES_PROBE_TIMEOUT` alongside the other environment variables in the usage block at the top of the script.

Memtensor-AI:

All tests passed (11/11 executed). memos_local_plugin/changed-repo-python: 11/11. Duration: 3s

**Branch:** `fix-20260916-local-plugin`

PR #2400, opened 2026-09-21
ci: extend local plugin npm visibility timeout

MLittleprince:
## Summary

- Extend the local-plugin npm registry visibility window from 150 seconds to 360 seconds.
- Keep the normal-release duplicate-version guard strict; existing npm versions still require explicit recovery after release-owner verification.
- Make the six-minute default shared by the workflow, shell publisher, and Node verifier, with regression coverage.

## Why

npm can acknowledge a publish before version, integrity, and dist-tag metadata are visible through the public registry. The previous 150-second window could fail a release after npm had already accepted it. This change gives propagation up to six minutes while preserving the no-second-publish safety behavior.

## Validation

- `node --test .github/scripts/wait-for-local-plugin-npm-release.test.mjs .github/scripts/publish-local-plugin.test.mjs .github/scripts/prepare-memos-release.test.mjs` (94/94)
- `bash -n .github/scripts/publish-local-plugin.sh`
- `git diff --check`

No npm publish, tag creation, GitHub Release, Docs sync, or deployment was performed by this PR.


Memtensor-AI:
## 🤖 Open Code Review

**Target:** PR #2400
**Task:** `98255dd7cf513698`
**Base:** `main`
**Head:** `fix/local-plugin-npm-visibility-timeout-360`

---

### 1. `.github/scripts/wait-for-local-plugin-npm-release.test.mjs` (L256-L258)

The assertion `assert.equal(DEFAULT_NPM_VISIBILITY_TIMEOUT_SECONDS, 600)` is placed **after** the assertions that depend on it (`clock` and `attempts`), yet it is conceptually the precondition that explains why those values are expected to be `600_000` and `60`. If this assertion ever fails (e.g., the constant is changed), the reader first sees that `clock` or `attempts` is wrong, which is confusing.

Consider moving the constant assertion to the top — before `assert.rejects` — so it acts as a documented precondition and fails with a clearer message first.

**Before:**
```
  assert.equal(DEFAULT_NPM_VISIBILITY_TIMEOUT_SECONDS, 600);
  assert.equal(clock, 600_000);
  assert.equal(attempts, 60);
```

**After:**
```suggestion
  assert.equal(DEFAULT_NPM_VISIBILITY_TIMEOUT_SECONDS, 600); // precondition: verify the constant value before running the timed test
  let clock = 0;
  let attempts = 0;
  await assert.rejects(
    // ... (existing rejects block unchanged)
  );
  assert.equal(clock, 600_000);
  assert.equal(attempts, 60);
```

</details>

Memtensor-AI:

All tests passed (1/1 executed). memos_github_open_source/smoke: 1/1. Duration: 2s

**Branch:** `fix/local-plugin-npm-visibility-timeout-360`

